PR #8, opened 2018-04-30
just use fetchInitiative (w/ and w/o data)

### LINKED ISSUE #7: getInitiative > fetchInitiative
and make it grab the `data` by default.

jgravois:
resolves #7

[regina george](https://youtu.be/Pubd-spHN-0?t=2s)

### DIFF
--- a/packages/common-types/src/index.ts
+++ b/packages/common-types/src/index.ts
@@ -7,37 +7,12 @@ import { IItem } from "@esri/arcgis-rest-common-types";
  * One small step for Hub, one large leap for hubkind
  */
 export interface IInitiative {
-  item?: IInitiativeItem;
-  data?: IInitiativeData;
+  item: IInitiativeItem;
+  data?: {
+    [propName: string]: any;
+  };
 }
 
 export interface IInitiativeItem extends IItem {
   type: "Hub Initiative";
-}
-
-export interface IInitiativeData {
-  assets: [any];
-  source: string;
-  values: {
-    collaborationGroupId: string;
-    openDataGroupId: string;
-    followerGroups: any;
-    steps: [string];
-    bannerImage: any;
-    website: IHaveNoIdeaWhatToCallThis;
-    informTools: IHaveNoIdeaWhatToCallThis;
-    listenTools: IHaveNoIdeaWhatToCallThis;
-    monitorTools: IHaveNoIdeaWhatToCallThis;
-    initiativeKey: string;
-  };
-}
-
-export interface IHaveNoIdeaWhatToCallThis {
-  title: string;
-  description: string;
-  id: string;
-  templates: any;
-  templateItemIds: any;
-  configuredItemIds: [string];
-  items: any;
 }
--- a/packages/initiatives/src/initiatives.ts
+++ b/packages/initiatives/src/initiatives.ts
@@ -4,11 +4,7 @@
 import { request, IRequestOptions } from "@esri/arcgis-rest-request";
 import { IItem } from "@esri/arcgis-rest-common-types";
 import { getItem, getItemData } from "@esri/arcgis-rest-items";
-import {
-  IInitiative,
-  IInitiativeItem,
-  IInitiativeData
-} from "@esri/hub-common-types";
+import { IInitiative, IInitiativeItem } from "@esri/hub-common-types";
 
 export interface IInitiativeRequestOptions extends IRequestOptions {
   /**
@@ -41,7 +37,7 @@ export function fetchInitiative(
       // shape this into a model
       return {
         item: result[0] as IInitiativeItem,
-        data: result[1] as IInitiativeData
+        data: result[1]
       };
     });
   }
--- a/packages/common-types/package-lock.json
+++ b/packages/common-types/package-lock.json
@@ -1,0 +1,11 @@
+{
+	"requires": true,
+	"lockfileVersion": 1,
+	"dependencies": {
+		"@esri/arcgis-rest-common-types": {
+			"version": "1.1.1",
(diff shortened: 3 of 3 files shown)


packages/initiatives/src/initiatives.ts
```diff
@@ -5,39 +5,37 @@ import { request, IRequestOptions } from "@esri/arcgis-rest-request";
...
-    getItemData(id, requestOptions)
-  ]).then(result => {
-    // shape this into a model
-    return {
-      item: result[0],
-      data: result[1]
-    };
-  });
+  if (requestOptions && !requestOptions.data) {
+    return getItem(id, requestOptions);
```
tomwayson: in this case should this return `{ item }` so that consumers can expect the return value to have a consistent shape?

packages/initiatives/src/initiatives.ts
```diff
@@ -5,39 +5,37 @@ import { request, IRequestOptions } from "@esri/arcgis-rest-request";
...
- * @param requestOptions - Request options that may have authentication manager
+ * @param requestOptions - Initiative request options that may have authentication manager
  * @returns A Promise that will resolve with the Initiative item and data
  */
-export function getInitiativeWithData(
+export function fetchInitiative(
   id: string,
-  requestOptions?: IRequestOptions
+  requestOptions?: IInitiativeRequestOptions
 ): Promise<any> {
```
tomwayson: Ideally this would be `Promise<IHubInitiative>` - but what would that interface look like?

```ts
interface IHubInitiative {
  item: IItem,
  data?: {
    [propName: string]: any;
  }
}
```

That's not really useful though. It would be better to have `data?: IHubInitativeData` and define what properties can be in the `data`. Maybe also have `IHubInitiative.item` to declare what special `properties` it can have as well.

jgravois: makes sense to me. i'll tackle tomorrow.

jgravois:
alright, i made a somewhat half assed attempt to stub out an interface to define both an Initiative's item and data properties and made sure the remaining `fetchInitiative` method exposes a consistent return signature.

full disclosure: i scoured the internet for roughly 15 seconds and copy/pasted JSON from [the first](https://www.arcgis.com/sharing/rest/search?q=typekeywords:hubInitiative&f=json&num=100) public initiative i found.

http://eapsprod.surge.sh/items/9aa4670ba0834e9ca7d98583581ad507/edit

this probably goes without saying, but, as with `arcgis-rest-js`, there's pretty much no end to the amount of REST documentation we _could_ surface through TypeDoc. Given that Hub presently lacks any other public effort, i'm a-okay with tackling it here first (and hardest) if that's _really_ how we want to go about it.

tomwayson:
Looping in @dbouwman into the debate about whether this is the place that we want to document the properties of hub entities (like initiatives), since it sounds like he may be cooking up some [JSON schema dealio](http://json-schema.org/) for Hub JSON. I agree that that TS interfaces in this repo are less than ideal, but it is something we can do immediately. To that end, I would say that we don't need to be exhaustive about it.

jgravois:
screenshot 2018-05-01 15 42 49

packages/common-types/src/index.ts
```diff
@@ -1,10 +1,43 @@
...
+    bannerImage: any;
+    website: IHaveNoIdeaWhatToCallThis;
+    informTools: IHaveNoIdeaWhatToCallThis;
+    listenTools: IHaveNoIdeaWhatToCallThis;
+    monitorTools: IHaveNoIdeaWhatToCallThis;
+    initiativeKey: string;
+  };
+}
+
+export interface IHaveNoIdeaWhatToCallThis {
```
tomwayson: I'mma guess `IInitiativeTools`, but I'm not even sure that these are used anymore. Let's make sure we have a recent, valid initiative first and that we need to get into this level of detail.

packages/common-types/src/index.ts
```diff
@@ -1,10 +1,43 @@
...
 /**
  * One small step for Hub, one large leap for hubkind
  */
-export interface IHasZM {
-  hasZ?: boolean;
-  hasM?: boolean;
-}
\ No newline at end of file
+export interface IInitiative {
+  item?: IInitiativeItem;
```
tomwayson: I would expect `item` to be required

packages/common-types/src/index.ts
```diff
@@ -1,10 +1,43 @@
...
+  data?: IInitiativeData;
+}
+
+export interface IInitiativeItem extends IItem {
+  type: "Hub Initiative";
```
tomwayson: probably should include `properties: IInitiativeProperties` here

dbouwman:
well... my *current* (as of 15 min ago) thinking is that we doc the current Hub v2 Initiative Json in a md doc over in the Hub repo a la card-json.

Then, we setup an intern to json-schema-ify all that stuff over the summer, along with tooling to generate docs directly from the schemas.

So - that's the plan...
The current state of Inititive Json does not lend itself well to creating ISomeSuchStuff interfaces, nor are things structured to support that...
for example... data.values is literally a hash of *anything*s some entries are indicators, some are steps, some are flat arrays of id's others are arrays of objects.

While this structure works for the initiative editor app, it's inscrutable for other consumers. So we will be changing things so it's much more structured.

With all that said... for now I'd suggest making the typescript types as wide open as possible.

jgravois:
> I'd suggest making the typescript types as wide open as possible.

makes sense to me. just pushed another commit with the generic interface suggested in https://github.com/Esri/hub.js/pull/8#discussion_r185136944a

PR #27, opened 2018-05-11
add @esri/hub-auth

jgravois:
```js
const session = checkSignInStatus()

if (session) // skip login!
```

to do:
- [x] trap for stale tokens/malformed sessions internally
- [ ] ?

### DIFF
--- a/packages/auth/src/auth.ts
+++ b/packages/auth/src/auth.ts
@@ -9,43 +9,51 @@ import { UserSession } from "@esri/arcgis-rest-auth";
  * ```js
  * import { checkSignInStatus } from '@esri/hub-auth';
  *
- * checkSignInStatus()
- *   .then(session => session); // ready to go!
- *   .catch() // time to initiate login
+ * const session = checkSignInStatus()
+ * if (session) // user is already logged in!
  * ```
  *
  * @returns UserSession
  */
-export function checkSignInStatus(): Promise<UserSession> {
-  let auth: string;
-  let session: UserSession;
+export function checkSignInStatus(): UserSession {
+  /* istanbul ignore else */
+  if (typeof window !== "undefined" && typeof document !== "undefined") {
+    let auth: string;
+    let session: UserSession;
 
-  // first check for an esri_auth cookie
-  // https://stackoverflow.com/questions/10730362/get-cookie-by-name
-  const value = "; " + document.cookie;
-  const parts = value.split("; " + "esri_auth=");
-  if (parts.length === 2) {
-    auth = parts
-      .pop()
-      .split(";")
-      .shift();
-  }
+    // first check for an esri_auth cookie
+    // https://stackoverflow.com/questions/10730362/get-cookie-by-name
+    const value = "; " + document.cookie;
+    const parts = value.split("; " + "esri_auth=");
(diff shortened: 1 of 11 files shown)


packages/auth/src/auth.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  if (auth) {
+    const authJson = JSON.parse(decodeURIComponent(auth));
+    session = new UserSession({
+      token: authJson.token,
+      tokenExpires: new Date(authJson.expires)
+    });
+  }
+
+  // since we're isolating synchronous logic, is it even helpful to return a promise?
```
tomwayson: Yeah, I don't think we need a promise here.

packages/auth/src/auth.ts
```diff
@@ -0,0 +1,51 @@
...
+  return new Promise((resolve, reject) => {
+    if (session) resolve(session);
+    else reject("No evidence of authentication found.");
```
tomwayson: If we don't return a promise, I'm not sure we should `throw` if no auth is found - maybe just return `null` and let the caller decide what to do from there.

packages/auth/test/auth.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  expires: TOMORROW,
+  allSSL: false,
+  accountId: "uCXeTVfooP4IIcx",
+  role: "account_admin",
+  urlKey: "subdomain",
+  customBaseUrl: "maps.arcgis.com"
+};
+
+/*
+  right now jasmine isn't running these tests in Node.js.
```
tomwayson: So, this comment made me think that `checkSignInStatus()` should if no `window` or `document`. 

Given that, it seems like what we want is to run the tests below only in a browser, and then add another test that runs only in node that verifies the above error. Is that possible w/ our test set up?

jgravois:
```console
$ node --version
v8.11.1
$ npm --version
5.6.0
$ nvm --version
0.33.11
$ npm install 
npm ERR! code 1
npm ERR! Command failed: /usr/bin/git checkout a51eca7f8bafdc9bafab76d2305fedc64503304c
npm ERR! error: Your local changes to the following files would be overwritten by checkout:
npm ERR! 	.github/ISSUE_TEMPLATE/Bug_report.md
npm ERR! 	.github/ISSUE_TEMPLATE/Custom.md
npm ERR! 	.github/ISSUE_TEMPLATE/Feature_request.md
npm ERR! Please commit your changes or stash them before you switch branches.
npm ERR! Aborting
```
i've seen travis throw a lot of errors, but _complaining about local changes to non-existent files_ is a new one.

tomwayson:
FYI - as per our discussion the other day, I think we should hold on this until we understand whether or not it is useful at all to check local storage for `torii-provider-arcgis`. A few reasons:
- that may only apply to apps that run on the .arcgis.com domain, so maybe the check should be done at the Ember addon level
- I want to understand https://github.com/Esri/torii-provider-arcgis/pull/43 and the subsequent PRs b/c @dbouwman preferred the cookie over the session auth, whereas this PR does the opposite

In short, we may not even need this package.

jgravois:
to be continued...

Issue #29, opened 2018-05-17
Add hub-annotations package w/ fn to get annotation service URL


tomwayson:
First fn should probably be getAnnotationServiceUrl(orgId). We don't have the API documented for this, but [here's the current implementation](https://github.com/ArcGIS/opendata-ui/blob/63c3e5037f96fbef244c8b580248a4c690c34366/packages/ember-arcgis-opendata-services/addon/services/hub-annotations.js#L20-L45).

PR #31, opened 2018-05-18
add @esri/hub-annotations

jgravois:
```js
getAnnotationServiceUrl("abc123")
  .then(response => // "https://services.arcgis.com/abc123/arcgis/rest/services/Hub Annotations/FeatureServer");

getAnnotationServiceUrl("xyz123")
  .catch(error => // "No annotation service found. Commenting is likely not enabled.");
```
* also bumped all package dependencies to `"arcgis-rest-*": "^1.2.0"`
* and chopped ~~hub-~~ from all .umd filenames (better now than after we publish to npm)

AFFECTS PACKAGES:
@esri/hub-annotations
@esri/hub-common-types
@esri/hub-initiatives
@esri/hub-sites

ISSUES CLOSED: #29


packages/annotations/src/annotations.ts
```diff
@@ -0,0 +1,73 @@
...
+    ...requestOptions
+  }).then(response => {
+    const annoResponse = response as IAnnoSearchResult;
+    if (annoResponse.results && annoResponse.results.length > 0) {
+      // this will need to be smarter if there is more than one service
+      let url = annoResponse.results[0].url;
+      // force https
+      url = url.replace(/^http:/gi, "https:");
+
+      return new Promise<string>((resolve, reject) => {
```
tomwayson: can't we just `return url;` here instead of returning a `new Promise()`?

packages/annotations/test/annotations.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      done();
+    });
+  });
+});
+
+describe("searchAnnotations", () => {
+  afterEach(fetchMock.restore);
+
+  it("should query for annotations when no parameters are passed", done => {
+    fetchMock.once("*", annoQueryResponse);
```
tomwayson: in the long run it would probably better to stub `queryFeatures()` and verify that expected params were passed, but this'll get'r done for now

packages/annotations/src/annotations.ts
jgravois: apparently we can!

packages/annotations/test/annotations.test.ts
jgravois: i was already verifying the request parameters inside the `then()` block. so i _think_ you're picking on the fact that i'm mocking requests so generically (ie: just listening for `"*"`)? 

if so, we do it all over the place in `arcgis-rest-*` packages and i've never felt quite right about it, so i pushed up another commit here that sets up the mock server to be a bit more selective in the requests it will respond to.

PR #34, opened 2018-07-03
JSON API response signature for query annotation response

jgravois:
@dmfenton @akharris:
can you two let me know what you think of the proposed signature above?
would it be more convenient if i opened an issue in hub-indexer?

```js
searchAnnotations({ url: `https://services.arcgis.com/bkrWlSKcjUDFDtgw/arcgis/rest/services/Hub%20Annotations/FeatureServer/0` })

{
  data: [ // features from the hub annotations query response
    {
      attributes: {
        OBJECTID: 1,
        author: "casey",
        description: "this is where the comments go",
        created_at: 1349395200000,
        dataset_id: "sgj432"
        // Shape__Area: 1, 
        // Shape__Length: 1
      }
    },
    {
      attributes: {
        OBJECTID: 2,
        author: "casey",
        description: "this is where the comments go",
        created_at: 1349395200001,
        dataset_id: "xds466"
        // Shape__Area: 1, // removed
        // Shape__Length: 1 // removed
      }
    }
  ],
  included: [ // filtered list of metadata associated with users that posted a comment
    {
      id: "casey",
      type: "user",
      attributes: {
        fullName: "Casey Jones",
        firstName: "Casey",
        lastName: "Jones",
        description: "Neogeographer and purveyor of personal storytelling.",
        thumbnail: "Andrew_Turner_-_DC_Social_Meetup_-_Square.jpg",
        // etc.
      }
    }
  ],
  /*meta: { // query response without the features
    objectIdFieldName: "OBJECTID",
    globalIdFieldName: "",
    geometryType: "esriGeometryPolygon",
    spatialReference: {
      wkid: 4326,
      latestWkid: 4326
    }
  }*/
}
```


tomwayson:
Thanks @jgravois!

@dmfenton / @akharris for context, the idea is that we want to this client-side fn to return something that is going to look like what the response will be when we get it into the REST API. @jgravois has taken a stab at JSONAPIifying a query to the annotation feature service and including the related AGO users. The first iteration doesn't have to have _all_ the things that will be included in full on a REST response right now - just the ones needed by the client app du jour (ExB today, Urban tomorrow). The hope is that we either can iterate on this and fill in the gaps as needed, or just build the actual REST API, ideally w/o any breaking changes.

W/ that in mind, a few questions:
- is it a big deal that the features in `data` don't include `relationships`? I don't _need_ that to find the related user in the client app
- I think the minimally useful `meta` would be just `stats` w/ `count` and `totalCount` - see https://hub.arcgis.com/api/v2/groups - that way we can implement paging. However, that's not planned for next week anyway, so do we omit `meta` for the first iteration?
- should we switch over to using the JSONAPI query params?

dmfenton:
> is it a big deal that the features in data don't include relationships? I don't need that to find the related user in the client app

Nope

> I think the minimally useful meta would be just stats w/ count and totalCount - see https://hub.arcgis.com/api/v2/groups - that way we can implement paging. However, that's not planned for next week anyway, so do we omit meta for the first iteration?

Skip it

> should we switch over to using the JSONAPI query params?

JSONAPI query parameters are about the wire format not the internal function calls to this library. No need to switch anything

dmfenton:
> can you two let me know what you think of the proposed signature above?

The existing signature `searchAnnoations(options)` is fine. I probably would have gone with `searchAnnotations(url, options)`.

dmfenton:
In terms of the data signature, I don't think you have to do this as relationship. We may just want to merge that information into the annotation itself if you wanted.

dmfenton:
What's with the shape_area shape length stuff?

We can leave out meta for now, but not sure I see the use of:   objectIdFieldName: "OBJECTID",
    globalIdFieldName: "",
    geometryType: "esriGeometryPolygon",
    spatialReference: {
      wkid: 4326,
      latestWkid: 4326
    }
  }

jgravois:
thanks for the review @dmfenton!

> What's with the shape_area shape length stuff?

right now we're setting `outFields=*` in the actual query, but it wouldn't be a bad idea to ask for explicit fields instead.

> not sure I see the use of: objectIdFieldName: "OBJECTID"...

agreed. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

jgravois:
ready for another round of review.

packages/annotations/src/annotations.ts
```diff
@@ -62,28 +62,52 @@ export function getAnnotationServiceUrl(
 export function searchAnnotations(
   requestOptions: IQueryFeaturesRequestOptions
 ): Promise<any /* make it a type */> {
```
tomwayson: I think we should export an `IResourceObject` interface as per: http://jsonapi.org/format/#document-resource-objects, and a this should be `Promise<{data: IResourceObject[], included?: IResourceObject[]}`.

packages/annotations/src/annotations.ts
```diff
@@ -62,28 +62,52 @@ export function getAnnotationServiceUrl(
 export function searchAnnotations(
   requestOptions: IQueryFeaturesRequestOptions
```
tomwayson: I think we should make querying the users (which is very expensive) is optional. I'm not sure if the best thing is to extend `IQueryFeaturesRequestOptions` w/ something like `include?: 'users'` (and then strip that before passing along) or take in a second argument.

packages/annotations/src/annotations.ts
```diff
@@ -62,28 +62,52 @@ export function getAnnotationServiceUrl(
...
-    if (response.features && response.features.length > 0) {
-      response.features.forEach(function(comment) {
-        if (users.indexOf(comment.attributes.author) === -1) {
-          users.push(comment.attributes.author);
-        }
-      });
-    }
+    // use .reduce()?
+    response.features.forEach(function(comment) {
+      if (users.indexOf(comment.attributes.author) === -1) {
```
tomwayson: let's build up `data` in here and add the `id`s (and maybe later the relationships), so: 
```
const attributes = comment.attributes;
data.push({
  id: attributes.OBJECTID,
  type: 'annotation',
  attributes
});
if (include === 'users') {
  users.push(attributes.author);
}
```

packages/annotations/test/annotations.test.ts
```diff
@@ -81,7 +81,11 @@ describe("searchAnnotations", () => {
         "begin:https://services.arcgis.com/uCXeTVveQzP4IIcx/arcgis/rest/services/Hub Annotations/FeatureServer/0/query"
       );
       expect(options.method).toBe("GET");
```
tomwayson: we should now be able to use spies instead of `fetchMock` here too.

PR #35, opened 2018-07-05
pass through CRUD annotation methods

jgravois:


packages/annotations/src/add.ts
```diff
@@ -0,0 +1,19 @@
...
+} from "@esri/arcgis-rest-feature-service";
+
+/**
+ * Add an annotation to ArcGIS Hub.
+ * @param requestOptions - request options that may include authentication
+ * @returns A Promise that will resolve with decorated features from the annotation service for a Hub enabled ArcGIS Online organization.
+ */
+
+export function addAnnotations(
+  requestOptions: any
```
tomwayson: do we not have a `IAddFeaturesRequestOptions`?

jgravois:
> I think it would be better to just stub the underlying fns

i agree. haven't figured out how to accomplish that yet though. i did however push up another commit that introduces at least _some_ internal logic.

### `addAnnotations()`

* appends `created_at=` with a valid UTC timestamp
* appends `source=hub.js`
* appends an `author` attribute (if/when authentication is provided).
* each of these values defer to whatever is supplied by the developer.

### `updateAnnotations()`

* appends `updated_at=` with a valid UTC timestamp

### deleteAnnotations()

is still dumb as rocks.

packages/annotations/src/add.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Add an annotation to ArcGIS Hub. Uses authentication to derive authorship, appends a timestamp and sets a default status of "pending" to new comments by default.
+ * @param requestOptions - request options that may include authentication
+ * @returns A Promise that will resolve with response from the service after attempting to add one or more new annotations.
+ */
+export function addAnnotations(
+  requestOptions: IAddFeaturesRequestOptions
+): Promise<IAddFeaturesResult> {
+  const session = requestOptions.authentication as UserSession;
```
jgravois: TIL _a better way_ than the [`IUserRequestOptions`](https://github.com/Esri/arcgis-rest-js/blob/b144d0551c80023e13912cc92c3b7590be3da30c/packages/arcgis-rest-auth/src/authenticated-request-options.ts#L16-L21) business i rallied for earlier.

tomwayson: So this works if the `authentication` is an [ApplicationSession](https://esri.github.io/arcgis-rest-js/api/auth/ApplicationSession/), right? I think you're only casting it here so TS doesn't blow up on the next line, right?

jgravois:
> So this works if the `authentication` is an ApplicationSession, right?

yup. if an [ApplicationSession](https://esri.github.io/arcgis-rest-js/api/auth/ApplicationSession/) is passed, the method behaves the same way as it does when _no_ credentials are supplied.

i've added another test to demonstrate this and refactored _all_ the tests to stub upstream methods based on your ✨ example.

PR #41, opened 2018-07-12
fix(all): treat arcigs-rest-js packages as external and bump to latest

tomwayson:
AFFECTS PACKAGES:
@esri/hub-annotations
@esri/hub-common-types
@esri/hub-initiatives
@esri/hub-sites


tomwayson:
I copied and pasted the annotations.umd.js into ExB node_modules and this works, so let's :shipit:

PR #55, opened 2018-08-22
[DNM] Get Initiative and Apply Migrations

dbouwman:
**This still requires doc-comments** thus the DNM

Adds schema migration logic to the `fetchInitiative` function.

Applies 3 current migrations, which are stored in `packages/initiatives/src/migrations`

### DIFF
--- a/packages/initiatives/src/initiatives.ts
+++ b/packages/initiatives/src/initiatives.ts
@@ -1,11 +1,18 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { request, IRequestOptions, IParams } from "@esri/arcgis-rest-request";
+import {
+  request,
+  IRequestOptions,
+  IParams,
+  getPortalUrl
+} from "@esri/arcgis-rest-request";
 import { IItem } from "@esri/arcgis-rest-common-types";
 import { getItem, getItemData } from "@esri/arcgis-rest-items";
-import { IInitiative, IInitiativeItem } from "@esri/hub-common";
+import { IInitiativeModel, IInitiativeItem } from "@esri/hub-common";
 import { fetchDomain } from "@esri/hub-sites";
+import { getProp, cloneObject } from "@esri/hub-common";
+import { migrateSchema, CURRENT_SCHEMA_VERSION } from "./migrator";
 
 export interface IInitiativeRequestOptions extends IRequestOptions {
   /**
@@ -23,7 +30,7 @@ export interface IInitiativeRequestOptions extends IRequestOptions {
 export function fetchInitiative(
   id: string,
   requestOptions?: IInitiativeRequestOptions
-): Promise<IInitiative> {
+): Promise<IInitiativeModel> {
   if (requestOptions && !requestOptions.data) {
     return getItem(id, requestOptions).then(result => {
       return {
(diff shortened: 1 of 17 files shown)


packages/initiatives/src/migrations/apply-schema.ts
```diff
@@ -0,0 +1,85 @@
+import { IInitiativeModel } from "@esri/hub-common";
+import { getProp, cloneObject } from "@esri/hub-common";
+
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
```
jgravois: nitpick, but the copyright belongs way up top.

packages/initiatives/src/migrations/upgrade-one-dot-one.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return model;
+}
+
+export function getResourceUrl(
+  itemId: string,
+  resourceName: string,
+  portal?: string,
+  folder?: string
+): string {
```
jgravois: i doubt i'm going to be able to convince you to pass a single POJO argument to methods like this with props so you can use methods like `getPortalUrl()` easily, so i'll just skip it.

packages/initiatives/src/migrations/upgrade-one-dot-one.ts
```diff
@@ -0,0 +1,95 @@
...
+      };
+    }
+    return clone;
+  } else {
+    // console.debug(`Not upgrading CurVersion: ${curVersion}`);
+    return model;
+  }
+}
+
+export function _addDefaultResources(
```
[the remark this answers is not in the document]
dbouwman: I removed the `_` and marked it `@private` so it's no in the docs - seems a reasonable comprimise

packages/initiatives/src/migrations/apply-schema.ts
dbouwman: fixed

packages/initiatives/src/migrations/upgrade-one-dot-one.ts
dbouwman: I *could* be convinced... but lets land this geezer... this function is marked private anyhow

docs/acetate.config.js
```diff
@@ -125,15 +125,27 @@ module.exports = function(acetate) {
...
-        const declarationPages = typedoc.declarations.map(declaration => {
-          return createPage.fromTemplate(
-            declaration.src,
-            path.join(acetate.sourceDir, "api", "_declaration.html"),
-            Object.assign({}, declaration, {
-              layout: "api/_layout:content"
-            })
-          );
-        });
+        // const declarationPages = typedoc.declarations.map(declaration => {
```
tomwayson: let's remove this commented code and add this to rest-js too!

packages/initiatives/src/initiatives.ts
```diff
@@ -1,47 +1,53 @@
...
-  data: boolean;
-}
+import { getProp, cloneObject } from "@esri/hub-common";
+import { migrateSchema, CURRENT_SCHEMA_VERSION } from "./migrator";
 
 /**
- * Get the initiative item + data in one call
+ * ```js
+ * // fetch initiative by id, along with the data
+ * fetchInitiative('3ef...', {data:true})
```
tomwayson: I don't think that `{data:true}` is a valid options hash, right?

dbouwman: eagle-eyes++

docs/acetate.config.js
dbouwman: I'll remove the commented code, but I want to give this more thought before we push this into rest-js. I would like to have the option to create "full" docs... what I've done here is pretty "harsh" fix... there may be another acetate hook we can use, or we can pre-parse the typedoc output... I will look at this now...

PR #59, opened 2018-09-07
Added examples for simple Annotation CRUD

ajturner:
Added examples - but before merging I would like to discuss simple improvements to the method signatures that exposes less of the implementation.

- Directly use the `annotationsUrl` without requiring adding a layer index `+ '/0'`
- Remove redundant parameter names `addAnnotations({adds: [...]})`
- Make Delete consistent with updates and add 
  - `addAnnotations({annotations: [{attributes: {description: "Kinda Awesome!"}]})`
  - `updateAnnotations({annotations: [{attributes: {OBJECTID: 1, description: "More Awesomer!"}]})`
  - `deleteAnnotations({annotations: [{attributes: {OBJECTID: 1}]})`

### DIFF
--- a/packages/annotations/README.md
+++ b/packages/annotations/README.md
@@ -17,24 +17,29 @@
 npm install @esri/hub-annotations
 ```
 ```js
-import { searchAnnotations } from "@esri/hub-annotations";
+import { getAnnotationServiceUrl, searchAnnotations } from "@esri/hub-annotations";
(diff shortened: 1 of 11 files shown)


jgravois:
i pushed another commit (and rebased to resolve a merge conflict) to move the snippets into the API reference and link to it from the README.

Issue #58, opened 2018-09-07
Annotation CRUD should not need adds: updates: deletes


ajturner:
It's odd that addAnnotations requires `adds: []` - which seems redundant to the actual name. This is because the function is a thin wrapper to feature service, but unnecessary for these focused functions. 

```js
arcgisHub.addAnnotations({url: annotationsUrl + "/0", adds: [
              {attributes: {description: "Test Anonymous", target: "none", author: "Hamilton"}}
            ]}).then( response => {
              console.log("addAnnotations", response);
          });
```

should become

```js
arcgisHub.addAnnotations({url: annotationsUrl, annotations: [
              {attributes: {description: "Test Anonymous", target: "none", author: "Hamilton"}}
            ]}).then( response => {
              console.log("addAnnotations", response);
          });
```

jgravois:
an option with perhaps less cognitive overhead than introducing _yet another_ interface would be to make a breaking change to `arcgis-rest-js` to use the word `"features"` instead of `"adds/updates/deletes"`.

tomwayson:
@ajturner - you make a valid point, but I don't think the TS gymnastics in #63 are worth it. @jgravois brings up a good point re: `features` in rest-js. Perhaps the current signatures w/ `adds`, `updates`, etc is too leaky an abstraction (hardly an abstraction at all) and is not as clean an interface as using `annotations` and `features`, but it's at least consistent all the way down the stack (Hub.js -> rest-js -> REST API) and familiar to esri devs.

I don't think we should change the way we're doing it.

tomwayson:
@jgravois  it looks like we went w/ `features` here: https://github.com/Esri/ember-arcgis-server-services/#feature-service

jgravois:
now that rest-js `v1.14.1` has been released, in hub.js wenow support _both_ `features` and `adds/updates`.

PR #63, opened 2018-09-17
refactor(anno): improve updateAnno signature

jgravois:
addresses #58 

this is the least amount of TypeScript gymnastics/[type guarding](https://www.typescriptlang.org/docs/handbook/advanced-types.html) i could figure out to support the signature for crud operations that @ajturner suggested.

if everyone is happy with this approach, i'll do the same for add/delete in the same PR.

AFFECTS PACKAGES:
@esri/hub-annotations

### DIFF
--- a/packages/annotations/src/update.ts
+++ b/packages/annotations/src/update.ts
@@ -1,21 +1,58 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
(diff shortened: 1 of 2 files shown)


jgravois:
>  at the end of the day, I land on "don't change anything"

that works for me.

i've proposed upstream that we consider swapping to use `"feature"` instead the next time we introduce breaking changes. i'm in favor of tackling that sooner rather than later.

https://github.com/Esri/arcgis-rest-js/issues/137

PR #64, opened 2018-09-19
BUG: Copy Resources during Initiative Activation 

dbouwman:
Two underlying issues:

- Due to a type in a destructuring call, we were sending an undefined item id into the `copyImageResources` method
- Due to not setting the item.owner in the `model.save` method, we were sending undefined owner into the `copyImageResources` method

Also bumps ArcGIS Rest Js deps to 1.11.0


tomwayson:
@dbouwman I may be crazy, but I don't see the rest-js version bump in this PR

Issue #70, opened 2018-09-20
anno search docs page got mangled


tomwayson:
https://esri.github.io/hub.js/api/annotations/searchAnnotations/

<img width="1085" alt="image" src="https://user-images.githubusercontent.com/662944/45827255-e65b3780-bcaa-11e8-8106-38eb73d07bef.png">

The samples in add, update, and delete all seem fine.

PR #71, opened 2018-09-20
docs(anno search): fix formatting of code snippet on annotations sear…

tomwayson:
…ch docs

AFFECTS PACKAGES:
@esri/hub-annotations

ISSUES CLOSED: #70 

This fixes the code snippet, but the response is still a bit jacked below:

image

### DIFF
--- a/packages/annotations/src/search.ts
+++ b/packages/annotations/src/search.ts
@@ -18,10 +18,9 @@ export interface IResourceObject {
 }
 
 /**
- * ````js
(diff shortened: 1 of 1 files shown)


tomwayson:
BTW - I guess the deal is that we can't have empty lines like [this](https://github.com/Esri/hub.js/pull/71/files#diff-e2a22ab3025ff2e92a5b8e06304dcbc3L23)

PR #85, opened 2018-10-09
WIP: events setup

jPurush:
- [x] changelog

- [x] test coverage


jgravois:
~~- [ ] changelog~~

one of the benefits of [all our tooling](https://github.com/Esri/arcgis-rest-js/blob/master/CONTRIBUTING.md#formatting-commit-messsages) is that `npm run c` (instead of `git commit`) generates commit messages that can be parsed and CHANGELOGged automatically.

packages/events/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "browser": "dist/umd/events.umd.js",
+  "module": "dist/esm/index.js",
+  "js:next": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@esri/arcgis-rest-auth": "^1.11.0",
+    "@esri/arcgis-rest-common-types": "^1.11.0",
+    "@esri/arcgis-rest-feature-service": "^1.11.0",
+    "@esri/arcgis-rest-items": "^1.11.0",
```
tomwayson: i'd prune this list of deps to only what is currently needed (if it's not already) and re-add whatever else is needed later.

packages/events/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "dist/**"
+  ],
+  "scripts": {
+    "prepare": "npm run build",
+    "build": "npm run build:node && npm run build:umd && npm run build:esm",
+    "build:esm": "tsc --module es2015 --outDir ./dist/esm --declaration",
+    "build:umd": "rollup -c ../../umd-base-profile.js && rollup -c ../../umd-production-profile.js",
+    "build:node": "tsc --module commonjs --outDir ./dist/node",
+    "dev:esm": "tsc -w --module es2015 --outDir ./dist/esm --declaration",
+    "dev:umd": "onchange 'src/**/*.ts' -- npm run rollup:dev",
```
tomwayson: I forgot to PR https://github.com/Esri/hub.js/pull/86 before you started on this, so you'll want to make [these changes](https://github.com/Esri/hub.js/pull/86/files#diff-75f00b8a9192522e46174597f8d2ffdb) here too. Sorry

packages/events/src/util.ts
```diff
@@ -0,0 +1,66 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IItem } from "@esri/arcgis-rest-common-types";
+import { searchItems, ISearchResult } from "@esri/arcgis-rest-items";
+
+export interface IEventSearchResult extends ISearchResult {
```
tomwayson: Do we need these interfaces? Can't we just use `ISearchResult` and `IItem` in this case?

jgravois: i'm guessing this is so TypeScript would quit complaining when she introspected for the `url` prop.

this is similar to what i did with [`IUserRequestOptions`](https://esri.github.io/arcgis-rest-js/api/auth/IUserRequestOptions/), but it's definitely worth a quick snoop to see if we can avoid it.

tomwayson:
Looks like `@esri/arcgis-rest-auth` _is_ required based on the travis errors.

jPurush:
Had one remaining error with the assignment of a string value to [`spatialReference`](https://github.com/Esri/hub.js/pull/85/files#diff-ca54db34165fa849702097e5467f2da6R41) in the json for the mock here, got around it by using [`any`](https://github.com/Esri/hub.js/pull/85/files#diff-ca54db34165fa849702097e5467f2da6R3). 

This is the error itself: 

```
[ts] Type 'string' has no properties in common with type 'ISpatialReference'.
item.d.ts(15, 5): The expected type comes from property 'spatialReference' which is declared here on type 'IItem'
(property) IItemAdd.spatialReference?: ISpatialReference
```

packages/events/test/mocks/ago_search.ts
```diff
@@ -0,0 +1,384 @@
...
+      tags: [],
+      snippet: "Feature service for public Hub events",
+      thumbnail: "thumbnail/ago_downloaded.png",
+      documentation: null,
+      extent: [
+        [-8589300.590117617, 40.36926825227528],
+        [-73.96624645399964, 4722244.554455302]
+      ],
+      categories: [],
+      spatialReference: "null",
```
jgravois: TypeScript is likely complaining about this stringified `"null"`.

jgravois:
🎉🎉🎉

Issue #89, opened 2018-10-12
getProp can't be accessed from @esri/hub-common


staehlli:
When trying to access an initiative which lives in an Enterprise organization, the following error appears (might be unrelated to this):

![image](https://user-images.githubusercontent.com/19825939/46877390-0f519100-ce41-11e8-9833-17b52b9c3453.png)

This does not let me open the demo of the Getting Started CDN sample.

Initiative Item ID: d2d8791e5fb846049193b877935b3435
Initiative Website: http://test-initiative-arcgis-urban-dev.opendata.arcgis.com/

Issue #88, opened 2018-10-12
Getting Started: Add more/better information about linked initiative


staehlli:
Improving the Getting Started experience by:

- Removing "abc123" and add something like <yourInitiativeID>, like this the code does not run and the user knows they have to replace it with something
- Give an example of an initiative ID, e.g. a public initiative, or maybe even set up and link to a public one that is intended for the demo
- Give guidance on how to find initiative ID (e.g. open initiative, copy item id from URL)
- As the initiative linked for the demo needs to be public (otherwise not accessible), mention the sharing settings as a precondition so that the demo works.

PR #90, opened 2018-10-12
add new dependencies to guide and use search to return an id

### LINKED ISSUE #87: Update unpkg.com links to reflect new package names
In the docs there are refs to unpkg urls that use the `hub-` package names

i.e. https://unpkg.com/@esri/hub-annotations@1.3.0/dist/umd/hub-annotations.umd.js

The correct url would be
https://unpkg.com/@esri/hub-annotations@1.3.0/dist/umd/annotations.umd.js

jgravois:
resolves #87, #88, #89

@lisastaehli i've updated the getting started guide to use searchInitiatives() to get a reference to an `id` instead of going into a deeper explanation.

hopefully this makes things a little more approachable. in the medium-term we'll work on  writing additional demo apps and a 'What is an initiative' guide topic to cover what they are and how they are created.

### DIFF
--- a/docs/acetate.config.js
+++ b/docs/acetate.config.js
@@ -210,7 +210,7 @@ module.exports = function(acetate) {
   acetate.helper("cdnUrl", function(context, package) {
     return `https://unpkg.com/${
       package.name
-    }@${package.version}/dist/umd/${package.name.split("/")[1]}.umd.js`;
+    }@${package.version}/dist/umd/${package.name.replace("@esri/hub-", "")}.umd.js`;
   });
 
   acetate.helper("npmInstallCmd", function(context, package) {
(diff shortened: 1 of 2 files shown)


jgravois:
no. looks like this was the same issue as https://github.com/Esri/arcgis-rest-js/issues/179

jgravois:
>I think the CDN links are still wrong, they seem to be without hub prefix now.

@lisastaehli i'm not sure i follow you. doesn't the fact that the example works _prove_ that the CDN links are correct?

https://unpkg.com/@esri/hub-initiatives@1.3.0/dist/umd/initiatives.umd.js

you can see the entire directory here:

https://unpkg.com/@esri/hub-initiatives@1.3.0/dist/umd/

lisastaehli:
Sorry for the confusion. I manually adjusted the CDN links before and then it worked. In the code of the pull request I still spotted the ‘hub-‘ prefix. Anyway, I think this is fixed now as well. So all good :)

jgravois:
> In the code of the pull request I still spotted the ‘hub-‘ prefix.

ahh, gotcha. that's a [`string.replace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) to _strip_ the prefixes.

PR #93, opened 2018-10-15
Feature: helper method to create a new anno service

jgravois:
```js
const authentication = new UserSession({
  username: "jgravois",
  password: "shhh"
})

createAnnotationService({
  orgId: "abc",
  authentication
})
  .then(response => console.log(response))
  .catch(err => console.warn(err))
```

to do:
- [ ] settle on a response signature (for when the service already existed too)
- [ ] _real_ test coverage


packages/annotations/package.json
```diff
@@ -9,12 +9,15 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
-    "@esri/arcgis-rest-auth": "^1.11.0",
-    "@esri/arcgis-rest-common-types": "^1.11.0",
-    "@esri/arcgis-rest-feature-service": "^1.11.0",
-    "@esri/arcgis-rest-items": "^1.11.0",
-    "@esri/arcgis-rest-request": "^1.11.0",
-    "@esri/arcgis-rest-users": "^1.11.0",
+    "@esri/arcgis-rest-auth": "^1.13.1",
```
tomwayson: so, below other rest-js packages are updated to `1.12.0`, should we bump them to `1.13.1`?

jgravois:
> the code looks good.

thank you!

> are there any significant deviations in the logic from the Ember code?

nope. the flow is the same. the only minor differences are:

* we use `/hub_annotations/` instead of `/Hub%20Annotations/` in the service url now
* the SpatialReference is fixed and the default extent is the whole world now
* `editorTrackInfo` is no longer included when updating the layer definition (because it doesn't have any effect)
* we don't fetch the new item before we update it anymore (to save a round trip to the server)
* we add cross site scripting prevention settings when creating the service now (h/t to @MikeTschudi for that one)

> as far as what to do when a service already exists, what did you have in mind?

right now i'm just returning the existing item from search.

> for the test coverage, is it just a matter of needing to inspect the args sent to all the calls?

yup. this was completed in da9542d

ajturner:
> * we use `/hub_annotations/` instead of `/Hub%20Annotations/` in the service url now

This is good (boo multiple words). 

> * the SpatialReference is fixed and the default extent is the whole world now

Can we default to the Org extent? 

> * `editorTrackInfo` is no longer included when updating the layer definition (because it doesn't have any effect)

We want to track who edited what features.

tomwayson:
> We want to track who edited what features

[this line](https://github.com/Esri/hub.js/pull/93/files#diff-31e92e3790ef978fc5ea8a69d91212daR78) does that, right @jgravois?

> Can we default to the Org extent?

I was going to suggest the same.

> right now i'm just returning the existing item from search

I'm torn. On the one hand I think it's better to `reject()` b/c that is more information for the consumer. Then let them decide whether to show an error or just return use the existing service. On the other hand, we've already found the existing service for them, so I don't want them to have to re-execute the same query. If we did reject, we'd want to at least create `getAnnotationService()` to make it easy for them to re-query. We'd then have this fn call that, and we'd probably want to use that in `getAnnotationServiceUrl()` too.

In my opinion, all of that is the "righter" way, but I'm not sure it's worth the hassle.

jgravois:
> Can we default to the Org extent?

done (in b772e2e)

> We want to track who edited what features

i'm under the same impression as @tomwayson (that we already are), but i'll plan on doing some manual testing to confirm when i play around with spawning a child view layer.
 
> I'm not sure it's worth the hassle.

i agree. lets just rough it in now, i have a few other cleanup ideas we can explore in a follow-up PR.

jgravois:
the tweaks above ensure the following capabilities are set on new hosted annotation services:

- [X] Enable editing.
- [ ] ~~Keep track of created and updated features.~~
- [X] Keep track of who created and last updated features.
- [X] Editing: Add, update, ~~and delete~~ features
- [X] Visible: Editors can see all features
- [X] Editors can only edit their own features (requires tracking)
- [X] Anonymous access: Only add new features, if allowed above (requires tracking)
- [X] service cant be deleted

i dont know what this one even means, and whats more, i just wasted 4 hours of my life unintentionally confirming that setting it breaks _everything_:

- [ ] Keep track of created and updated features.

to put it another way, i have confirmed manually that the settings here ensure the following

1. anonymous users can **only** add new features
2. named users can add new features **and** update their own features
3. organization administrators (and i assume those with special priviledges) **can** update comments from named users
3. **no one** can delete features

PR #97, opened 2018-10-29
adding getEventQueryFromType util method

jPurush:
util to convert event type to where clause

### DIFF
--- a/packages/events/src/util.ts
+++ b/packages/events/src/util.ts
@@ -66,38 +66,38 @@ export function getEventServiceUrl(
  */
 export function getEventQueryFromType(
   type: string,
-  params: IQueryFeaturesRequestOptions,
-  requestOptions?: IRequestOptions
+  params: IQueryFeaturesRequestOptions
 ): IQueryFeaturesRequestOptions {
   // this allows us to ask for type === upcoming | past | cancelled | draft
   // and get an appropriate `where` and `orderByFields`
   let typeWhere;
+  const newParams = Object.assign({}, params);
   if (type === "cancelled") {
     if (!params.orderByFields) {
       // if orderByFields was passed in, use it, otherwise use appropriate one for cancelled
-      params.orderByFields = "EditDate DESC";
+      newParams.orderByFields = "EditDate DESC";
     }
     typeWhere = `isCancelled=1 AND status<>'draft'`;
   } else if (type === "draft") {
     if (!params.orderByFields) {
-      params.orderByFields = "EditDate DESC";
+      newParams.orderByFields = "EditDate DESC";
     }
-    const session = requestOptions.authentication as UserSession;
+    const session = params.authentication as UserSession;
     const user = session ? session.username : null;
     typeWhere = `Creator = '${user}' AND status = 'draft'`;
   } else {
     if (!params.orderByFields) {
       // if orderByFields was passed in, use it, otherwise use appropriate one for type
-      params.orderByFields =
+      newParams.orderByFields =
         type === "upcoming" ? "startDate ASC" : "startDate DESC";
     }
     const operator = type === "upcoming" ? ">" : "<=";
     typeWhere = `endDate${operator}CURRENT_TIMESTAMP AND (isCancelled<>1 OR isCancelled IS NULL) AND status<>'draft'`;
   }
   if (params.where) {
-    params.where = params.where + ` AND ${typeWhere}`;
+    newParams.where = params.where + ` AND ${typeWhere}`;
   } else {
-    params.where = typeWhere;
+    newParams.where = typeWhere;
   }
-  return params;
+  return newParams;
 }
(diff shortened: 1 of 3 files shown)


packages/events/src/util.ts
```diff
@@ -55,3 +56,48 @@ export function getEventServiceUrl(
...
+ * @param type - string to indicate event type with the options `upcoming`, `past`, `cancelled` and `draft`
+ * @param params - query features request options
+ * @param requestOptions - request options that may include authentication
+ * @returns an IQueryFeaturesRequestOptions with modified Where and OrderBy params
+ */
+export function getEventQueryFromType(
+  type: string,
+  params: IQueryFeaturesRequestOptions,
+  requestOptions?: IRequestOptions
+): IQueryFeaturesRequestOptions {
```
jgravois: when a method has more than one _required_ input parameter, the (undocumented) convention we've agreed on is to stuff them all into a custom `IRequestOptions` interface.
```ts
export interface IGetEventQueryRequestOptions extends IQueryFeaturesRequestOptions {
  eventType: string;
}

export function getEventQueryFromType(requestOptions: IGetEventQueryRequestOptions): any
```

tomwayson: This is a helper util though, so we don't need to follow that convention here.

The signature as defined is misleading. I'm not sure this should return a `IQueryFeaturesRequestOptions`. Seems to me it should return just `{ where: string; orderByFields: string; }`

jPurush: How about if there is nothing returned at all? The function actually manipulates the input `IQueryFeaturesRequestOptions`

packages/events/src/util.ts
```diff
@@ -55,3 +56,48 @@ export function getEventServiceUrl(
     }
   });
 }
+
+/**
+ * Get the events query based on type.
+ * @param type - string to indicate event type with the options `upcoming`, `past`, `cancelled` and `draft`
+ * @param params - query features request options
+ * @param requestOptions - request options that may include authentication
```
tomwayson: I think this should just take `type: string; requestOptions: IQueryFeaturesRequestOptions`. There's no need for `params`. `requestOptions` would have the `where` and `orderByFields` either as it's own properties, or under it's own `params`.

packages/events/src/util.ts
tomwayson: The function should not have side effects if avoidable. Whatever it needs to change should be returned. Then the caller can do

```js
newOptions = getEventQueryFromType(type, originalOptions);
queryFeatures({...originalOptions, ...newOptions});
```

packages/events/src/util.ts
```diff
@@ -55,3 +56,47 @@ export function getEventServiceUrl(
...
+    if (!params.orderByFields) {
+      // if orderByFields was passed in, use it, otherwise use appropriate one for type
+      newParams.orderByFields =
+        type === "upcoming" ? "startDate ASC" : "startDate DESC";
+    }
+    const operator = type === "upcoming" ? ">" : "<=";
+    typeWhere = `endDate${operator}CURRENT_TIMESTAMP AND (isCancelled<>1 OR isCancelled IS NULL) AND status<>'draft'`;
+  }
+  if (params.where) {
+    newParams.where = params.where + ` AND ${typeWhere}`;
```
rgwozdz: Since you are already using string templating you could skip the concatenation with `+` and just do ``newParams.where = `${params.where}  AND ${typeWhere}`;``

packages/events/src/util.ts
```diff
@@ -55,3 +56,47 @@ export function getEventServiceUrl(
...
+
+/**
+ * Get the events query based on type.
+ * @param type - string to indicate event type with the options `upcoming`, `past`, `cancelled` and `draft`
+ * @param params - query features request options
+ * @returns an IQueryFeaturesRequestOptions that has the same values as `params` but for the modified Where and OrderBy properties
+ */
+export function getEventQueryFromType(
+  type: string,
+  params: IQueryFeaturesRequestOptions
```
tomwayson: great work.

one last suggestion, change the name of `params` to `options`. I know it's only internal to this function, but it will appear in the docs, and in hub.js/rest-js land, "options" and "params" have specific meaning.

packages/events/src/util.ts
```diff
@@ -55,3 +56,47 @@ export function getEventServiceUrl(
...
 }
+
+/**
+ * Get the events query based on type.
+ * @param type - string to indicate event type with the options `upcoming`, `past`, `cancelled` and `draft`
+ * @param params - query features request options
+ * @returns an IQueryFeaturesRequestOptions that has the same values as `params` but for the modified Where and OrderBy properties
+ */
+export function getEventQueryFromType(
+  type: string,
```
jgravois: if you're already going back in here one more time...
```ts
// not just _any_ string
type: "upcoming" | "past" | "cancelled" | "draft"
```

jPurush: I did not know I could do that! that's awesome, thanks @jgravois

PR #98, opened 2018-10-29
add guide for working with annotation services

jgravois:
while i was in there i realized that the code i wrote previously for tacking on the 'author' attribute in annotation CRUD functions needed to be removed.

`created_at`, `updated_at`, `author` and `updater` are all managed by the server when editor tracking is enabled.

### DIFF
--- a/docs/src/guides/work-with-annotations.md
+++ b/docs/src/guides/work-with-annotations.md
@@ -34,7 +34,7 @@ getPortal("http://custom.maps.arcgis.com")
 import { UserSession } from "@esri/arcgis-rest-auth";
 import { createAnnotationService } from "@esri/hub-annotations";
 
-// this user will be the owner of the new annotation service
+// user with sufficient privileges to create a new hosted feature service
 const authentication = new UserSession({
   username: "joe",
   password: "shhhh"
--- a/docs/src/guides/work-with-annotations.md
+++ b/docs/src/guides/work-with-annotations.md
@@ -1,0 +1,68 @@
+---
+title: Work with annotations.
+navTitle: Work with annotations
+description: Learn how to work with an annotation service using @esri/hub.js.
+order: 40
+group: 1-get-started
+---
+
+Annotations are geographic comments created by an Author for discussion of an Initiative, about an Events, or feedback of a Dataset. This feedback can also be submitted anonymously.
+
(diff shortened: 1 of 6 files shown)


docs/src/guides/work-with-annotations.md
```diff
@@ -0,0 +1,68 @@
...
+
+@esri/hub.js can be used to _create_ an ArcGIS Hub compatible annotation service in an organization too.
+
+```js
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { createAnnotationService } from "@esri/hub-annotations";
+
+// this user will be the owner of the new annotation service
+const authentication = new UserSession({
+  username: "joe",
```
tomwayson: you might want to use "joeAdmin" to convey who is expected to be able to do this kind of thing.

packages/annotations/src/add.ts
```diff
@@ -45,14 +45,9 @@ export interface IAddAnnotationsRequestOptions
...
   requestOptions: IAddAnnotationsRequestOptions
 ): Promise<IAddFeaturesResult> {
-  const session = requestOptions.authentication as UserSession;
-  const author = session ? session.username : null;
-
   requestOptions.adds.forEach(function(anno) {
     const defaults = {
-      created_at: new Date().getTime(),
       status: "pending",
-      author,
```
tomwayson: I assume the idea behind removing `updated_at` and `created_at` is that they are added on the server. 

However, why remove `author`? That seems like a breaking change.

jgravois: the server adds `author` as well. my guess is that it introspects it from the token.

docs/src/guides/work-with-annotations.md
jgravois: not a bad idea, but admin privs aren't required. maybe `level2joe`?

packages/annotations/src/add.ts
tomwayson: https://www.youtube.com/watch?v=3TUNSzbEypA

docs/src/guides/work-with-annotations.md
tomwayson: `userWithPrivledges`?

jgravois: i went with a more explicit code comment.

PR #106, opened 2018-11-21
fix: skip fetching metadata for annotations submitted anonymously

jgravois:
and append `geometry` to results when an annotation includes it.


jPurush:
This LGTM!

PR #107, opened 2018-11-22
fix invalid semver ranges all over the place

jgravois:
:rage1:


jgravois:
i didn't think of this previously, but now that v1.14.1 of @esri/arcgis-rest has been released, for TypeScript users sake we actually _have_ to make sure our code accounts for the fact that `features` can now be passed as a constructor option to addAnnotations. (see https://github.com/Esri/arcgis-rest-js/issues/391 for more info)

neither change on its own would be enough to please travis so i decided to just kill two 🐦s with one 🗿and push a backwards compatible patch directly onto this PR.

PR #111, opened 2018-12-18
add support for annotation voting

### LINKED ISSUE #110: Support voting on annotations
As proposed by Andrew: https://gist.github.com/ajturner/cd7b927b387b360e4c1f20a0f6a34014

We propose to add the following functions to hub.js:
* add positive or negative vote to an annotation (e.g. `addAnnotationVote`)
* get positive votes for an annotation (e.g. `searchAnnotationUpVotes`)
* get negative votes for an annotation (e.g. `searchAnnotationDownVotes`)

jgravois:
resolves #110 
```js
const orgId = "uCX...";
const authentication = new arcgisRest.UserSession({ username, password })

const serviceUrl = await arcgisHub.getAnnotationServiceUrl(orgId)
const url = `${serviceUrl}/0`;

const searchResponse = await arcgisHub.searchAnnotations({
  url,
  where: "OBJECTID = 1"
});

arcgisHub.voteOnAnnotation({
  url,
  authentication,
  annotation: searchResponse.data[0],
  downVote: true
});
```
to do:
-  [x] tests e5fddb3

### DIFF
--- a/packages/annotations/src/layer-definition.ts
+++ b/packages/annotations/src/layer-definition.ts
@@ -202,6 +202,15 @@ export const annotationServiceDefinition: ILayerDefinition = {
       editable: true,
       domain: null,
       defaultValue: null
+    },
+    {
+      name: "value",
+      type: "esriFieldTypeInteger",
+      alias: "value",
+      nullable: true,
+      editable: true,
+      domain: null,
+      defaultValue: null
     }
   ],
   geometryType: "esriGeometryPolygon",
--- a/packages/annotations/src/vote.ts
+++ b/packages/annotations/src/vote.ts
@@ -1,0 +1,83 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import {
+  addFeatures,
+  IAddFeaturesRequestOptions,
+  IAddFeaturesResult
+} from "@esri/arcgis-rest-feature-service";
+
+import { IAnnoFeature } from "./add";
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { searchAnnotations } from "./search";
+
+export interface IVoteRequestOptions extends IRequestOptions {
+  url: string;
+  annotation: IAnnoFeature;
+  /**
+   * Set to 'true' in order to vote another annotation down.
+   */
+  downVote?: boolean;
+}
+
+/**
+ * ```js
+ * import { vote } from "@esri/hub-annotations";
+ * //
+ * voteOnAnnotation({
+ *   url,
+ *   annotation: { attributes: {} },
+ *   authentication
+ * })
+ *   .then(response);
+ * ```
+ * Upvote (or downvote) another annotation.
+ * @param requestOptions - request options that may include authentication
+ * @returns A Promise that will resolve with the response from the service after attempting to vote on an annotation.
+ */
+export function voteOnAnnotation(
+  requestOptions: IVoteRequestOptions
+): Promise<IAddFeaturesResult> {
+  const annotation = requestOptions.annotation;
+  const auth = requestOptions.authentication as UserSession;
+  const url = requestOptions.url;
+
+  if (!auth) {
+    throw Error(`Voting by anonymous users is not supported.`);
+  }
+
+  if (annotation.attributes.author === auth.username) {
+    throw Error(`Users may not vote on their own comment/idea.`);
+  }
+
+  return searchAnnotations({
+    url,
+    where: `parent_id = '${annotation.attributes.OBJECTID}' AND author = '${
+      auth.username
+    }'`
+  }).then(searchResponse => {
+    // dont let folks vote on an idea more than once
+    if (searchResponse.data.length > 0) {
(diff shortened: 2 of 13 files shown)


packages/annotations/src/vote.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+  }
+
+  return searchAnnotations({
+    url,
+    where: `parent_id = ${annotation.attributes.OBJECTID} AND author = '${
+      auth.username
+    }'`
+  }).then(searchResponse => {
+    // dont let folks vote on an idea more than once
```
tomwayson: What if they want to change their vote from an upvote to a downvote? We don't (yet) provide an API to remove the previous vote first, so it seems that unless you know the implementation details you can't really change your vote. Is that intentional?

packages/annotations/src/vote.ts
```diff
@@ -0,0 +1,125 @@
...
+        const deleteOptions: IDeleteFeaturesRequestOptions = {
+          url,
+          authentication: requestOptions.authentication,
+          objectIds: [queryResponse.features[0].attributes.OBJECTID]
+        };
+        return (
+          deleteFeatures(deleteOptions)
+            // i have NO idea why the union return type is being ignored
+            .then(() => {
+              return { addResults: [{ success: true }] } as IAddFeaturesResult;
```
jgravois: in cc576c880617b57cb64443a0fa30d48f4ff0809f i added support for vote switching. this is accomplished by deleting the previous vote (if the new vote is different than the first).

for reasons i don't understand, i had to shape the response of the deleteFeatures() call to impersonate `IAddFeaturesResult` in order to get TypeScript to shut up.

packages/annotations/src/layer-definition.ts
```diff
@@ -202,6 +202,15 @@ export const annotationServiceDefinition: ILayerDefinition = {
...
       defaultValue: null
+    },
+    {
+      name: "value",
+      type: "esriFieldTypeInteger",
+      alias: "value",
+      nullable: true,
+      editable: true,
+      domain: null,
+      defaultValue: null
```
jgravois: just to be explicit, this creates a new `value` field in hosted feature services that have yet to be created. i'm not sure what we can/should do about the ones that are already 'in the wild'.

packages/annotations/src/vote.ts
```diff
@@ -0,0 +1,125 @@
...
+  if (annotation.attributes.author === auth.username) {
+    return new Promise((resolve, reject) => {
+      reject(
+        new ArcGISRequestError(`Users may not vote on their own comment/idea.`)
+      );
+    });
+  }
+
+  // searchAnnotations() would make more xhrs for user metadata
+  return queryFeatures({
```
jgravois: i realized during the most recent refactor that `searchAnnotations()` creates even more network chatter (via appending user metadata), so i switched to using `queryFeatures()` instead.

packages/annotations/src/vote.ts
```diff
@@ -0,0 +1,125 @@
...
+    url,
+    where: `parent_id = ${annotation.attributes.OBJECTID} AND author = '${
+      auth.username
+    }'`
+  }).then((
+    queryResponse: any /* why can't i declare it as IQueryFeaturesResponse ? */
```
jgravois: here's another one headscratcher.

`queryFeatures()` returns a union type, yet i can't explicitly cast a response i know will be `IQueryFeaturesResponse` (and not `IQueryResponse`) to introspect properties that i **know** will be available, so i've resorted to yet another `any`.

my TS foo is too weak.

packages/annotations/src/layer-definition.ts
tomwayson: we should probably create a migrate function

packages/annotations/src/vote.ts
tomwayson: by cast, did you mean `(queryResponse as IQueryFeaturesResponse)`?

packages/annotations/src/layer-definition.ts
tomwayson: to be clear, I don't necessarily mean that has to happen in this PR.

packages/annotations/src/vote.ts
jgravois: yup.  	

> Argument of type 'IQueryFeaturesResponse' is not assignable to parameter of type '(value: IQueryFeaturesResponse | IQueryResponse) => IQueryFeaturesResponse | IQueryResponse | PromiseLike<IQueryFeaturesResponse | IQueryResponse>'.
  Type 'IQueryFeaturesResponse' provides no match for the signature '(value: IQueryFeaturesResponse | IQueryResponse): IQueryFeaturesResponse | IQueryResponse | PromiseLike<IQueryFeaturesResponse | IQueryResponse>

🤔

packages/annotations/src/layer-definition.ts
jgravois: that's what i assumed.

packages/annotations/src/vote.ts
tomwayson: it seems to me that after deleting we still need to add the new vote, but you `return` here, so I don't think that happens. Am I missing something?

PR #115, opened 2019-01-23
adding functions for votes per comment 

jPurush:


jgravois:
let me know if you need any help getting this landed.

packages/annotations/src/search.ts
```diff
@@ -105,3 +108,160 @@ export function searchAnnotations(
...
+  requestOptions.groupByFieldsForStatistics = "value";
+  const outStat: IStatisticDefinition = {
+    statisticType: "count",
+    onStatisticField: "value",
+    outStatisticFieldName: "value_count"
+  };
+  requestOptions.outStatistics = [outStat];
+  // filtering for the comment
+  const commentFilteringClause = "parent_id=" + annotation.attributes.OBJECTID;
+  requestOptions.where += requestOptions.where ? "+AND+" : "";
```
jPurush: is this `where` check unnecessary?

jgravois: my gut reaction is `yes'.

i don't think this method needs to support blending a filter provided in `IQueryFeaturesRequestOptions` when querying for summary stats on the annotation it already has a reference to.

packages/annotations/src/search.ts
```diff
@@ -105,3 +108,160 @@ export function searchAnnotations(
...
+ * ```
+ * Query for up and down votes on a comment from ArcGIS Hub.
+ * @param requestOptions - request options that may include authentication
+ * @param annotation - the annotation for which votes need to be counted
+ * @returns A Promise that will resolve with summary statistics for the specified annotation from the annotation service for a Hub enabled ArcGIS Online organization.
+ */
+export function searchAnnotationVotes(
+  requestOptions: IQueryFeaturesRequestOptions,
+  annotation: IAnnoFeature
+): Promise<{ data: IVoteResourceObject[] }> {
```
jgravois: lets rename this function and refactor it so that only one input object needs to be passed.
```ts
export function searchSingleAnnotationVotes(
  requestOptions: ISearchAnnoRequestOptions // IRequestOptions + url + IAnnoFeature
): Promise<{ data: IVoteResourceObject[] }> {
```
we can still cast an `IQueryFeaturesRequestOptions` internally, but i don't _think_ we want/need to allow folks to pass other query filters.

packages/annotations/src/search.ts
```diff
@@ -105,3 +108,160 @@ export function searchAnnotations(
...
+      }
+    );
+    data.push(resource);
+    return { data };
+  });
+}
+
+/**
+ * ```js
+ * import { searchAnnotationVotes } from "@esri/hub-annotations";
```
jgravois:
```js
import { searchAllAnnotationVotes } from "@esri/hub-annotations";
```

packages/annotations/src/search.ts
```diff
@@ -105,3 +108,167 @@ export function searchAnnotations(
...
+ *     //     id,
+ *     //     type: "votes",
+ *     //     attributes: {
+ *     //       upVotes: 3,
+ *     //       downVotes: 0
+ *     //     }
+ *     //   }]
+ *    });
+ * ```
+ * Query for up and down votes on a comment from ArcGIS Hub.
```
jgravois:
```
Query for up and down votes on multiple comments from ArcGIS Hub.
```

packages/annotations/src/search.ts
```diff
@@ -105,3 +108,167 @@ export function searchAnnotations(
...
+    downVotes: number;
+  };
+}
+
+/**
+ * ```js
+ * import { searchSingleAnnotationVotes } from "@esri/hub-annotations";
+ * //
+ * searchSingleAnnotationVotes({ url: annotationsUrl + "/0",
+ *                         annotation: {
```
jgravois: can you reindent this code snippet?

packages/annotations/src/search.ts
jgravois: this one still needs updated.

packages/annotations/src/search.ts
jgravois: it looks like you mixed up the descriptions for your two new methods, when you made this update.

PR #116, opened 2019-01-29
Add maybeAdd util fn

dbouwman:
Once you see how this works, you will ask yourself... "Self, how have I lived without this?"

Ok, maybe it's not that awesome, but in Hub-land we do a LOT of data manipulation, and that involves a lot of checking if some deep-nested-property exists, and if to, grabbing it's value and pushing it into another object, or into an array.

`maybeAdd` allows us to drop a lot of the `if(valueExists) { pushValueIntoArray }` code.

Instead we can just 
```js
let ids =[];
ids = maybeAdd(ids, getProp(model, 'item.id'));
ids = maybeAdd(ids, getProp(model, 'item.properties.serviceId'));
ids = maybeAdd(ids, getProp(model, 'item.data.values.webmap'));
```
or get super cool and use it in a reduce...

```js
const ids = ['item.id', 'item.properties.sourceServiceId', 'data.values.webmap'].reduce((acc, key) => {
  return maybeAdd(acc, getProp(model, key));
}, []);
```

And if you wanted to do this with an object?

```js
const idHash = ['item.id', 'item.properties.serviceId', 'data.values.webmap'].reduce((acc, key) => {
  let prop = key.split('.').reverse()[0];
  return maybeAdd(acc, getProp(model, key, prop));
}, {});
```

As good functional citizens, we clone the `arrayOrObject` because mutation makes us sad. The value itself is appended by-ref if an object or array, and by-val if it's a simple type.

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+/* Copyright (c) 2018-2019 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
 /**
@@ -117,7 +117,7 @@ export function createId(prefix: string = "i"): string {
 
 /**
  * If value is not null, push it into an array, or append as a property of an object.
- * This is a very useful companion to the getProp utility. Note: the array or object
+ * This is a very useful companion to [getProp()](../getProp/). Note: the array or object
  * that is passed in is cloned before being appended to.
  *
  * Allows for code like:
@@ -141,13 +141,13 @@ export function createId(prefix: string = "i"): string {
  *  }
  * };
  *
- * // lets pluck some id's into an array...
+ * // lets pluck some values and place them in an array...
  * let vals = maybeAdd([], getProp(obj, 'item.properties.sourceId'));
  * // vals => ['3ef]
  *
- * // now try to get a value from a property that is missing...
+ * // try to get a value from a property that is missing...
  * vals = maybeAdd(vals, getProp(obj, 'item.properties.childId'));
- * // vals => ['3ef]
+ * // vals => []
  *
  * // Let's pluck some details into an object. This time we'll use an array
  * // of properties to pluck out of the source
@@ -170,7 +170,7 @@ export function createId(prefix: string = "i"): string {
  * ```
  * @param objectOrArray - the object (or array) to update
  * @param val - the possibly null value
- * @param key - optional key (used when appending to an objeect)
+ * @param key - optional key (used when appending to an object)
  */
 export function maybeAdd(
   objectOrArray: any,
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -116,15 +116,15 @@ export function createId(prefix: string = "i"): string {
 }
 
 /**
- * If value is not null, push it into an array, or append as a property of an object.
- * This is a very useful companion to [getProp()](../getProp/). Note: the array or object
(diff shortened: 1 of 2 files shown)


packages/common/src/util.ts
```diff
@@ -115,6 +115,46 @@ export function createId(prefix: string = "i"): string {
     .substr(2, 8)}`;
 }
 
+/**
+ * If value is not null, push it into an array, or append as a property of an object
+ * Allows for code like:
+ * ```js
+ * const vals = maybeAdd([], getProp(obj, 'some.deep.path.thatMayBe.undefined'));
+ * ```
```
jgravois: i'm _pretty sure_ our static site generator can only deal with a single snippet though, so your TypeDoc might make it barf. (you can run `npm run doc:serve` to find out for sure.)

another option would be to just rip off the _flawless_ ✨ example here:
https://github.com/Esri/arcgis-rest-js/blob/8173aa5c2d92b2f3f47ff108a8fb575fe449b99b/packages/arcgis-rest-geocoder/src/geocode.ts#L68-L87

packages/common/src/util.ts
```diff
@@ -115,6 +115,46 @@ export function createId(prefix: string = "i"): string {
...
+ *
+ * or
+ *
+ * ```js
+ * let summary = {}
+ * ['item.title', 'item.description', 'item.snippet'].forEach((k) => {
+ *   summary = maybeAdd(summary, getProp(model, k));
+ * })
```
jgravois: both of these snippets would be even _more_ useful if they were explicit about what is returned.

packages/common/src/util.ts
```diff
@@ -115,6 +115,46 @@ export function createId(prefix: string = "i"): string {
...
+ * ```
+ * @param objectOrArray
```
jgravois: providing a description here would flesh out the table in the API reference.

jgravois:
@dbouwman awesome! this looks great.

i pushed up a commit of my own with a few typo fixes.

dbouwman:
After a chat with @mjuniper I decided to break this into two functions `maybePush` for arrays and `maybeAdd` for objects. Updated docs & tests.

image

jgravois:
> I'm not sure if this is a "just on Dave's machine" issue

alas, it is. i shutter to think what you could accomplish with a fully operational battlestation.

packages/common/src/util.ts
```diff
@@ -115,6 +115,124 @@ export function createId(prefix: string = "i"): string {
...
+ * vals = maybeAdd(getProp(obj, 'item.properties.childId'), vals);
+ * // vals => ['3ef]
+ *
+ * // easily pluck values via property paths
+ * const summary = [
+ *  'item.id',
+ *  'item.properties.sourceId',
+ *  'item.properties.childId',
+ *  'data.parcelLayer.itemId'].reduce((acc, prop) => {
+ *   return maybeAdd(getProp(model, key), acc);
```
jgravois: `maybePush`?

packages/common/src/util.ts
```diff
@@ -115,6 +115,124 @@ export function createId(prefix: string = "i"): string {
...
+ *  data: {
+ *    theme: 'orange',
+ *    parcelLayer: {
+ *      itemId: '7ca',
+ *      primaryField: 'PIN'
+ *    }
+ *  }
+ * };
+ * // lets pluck some id's into an array...
+ * let vals = maybeAdd(getProp(model, 'item.properties.sourceId'), []);
```
jgravois: `maybePush`?

PR #121, opened 2019-01-31
fixed voting function description error

jPurush:


packages/annotations/src/search.ts
```diff
@@ -144,7 +144,7 @@ export interface IVoteResourceObject {
  *     //   }]
  *    });
  * ```
- * Query for up and down votes on multiple comments from ArcGIS Hub.
+ * Query for up and down votes on a comment from ArcGIS Hub.
```
tomwayson: we should probably use "annotation" instead of "comment" here and throughout the docs

jgravois:
thx y'all!

this fix will hit the live page momentarily. :rocket:

tomwayson:
thx! I was just checking .travis.yml to confirm that ^^^

Issue #123, opened 2019-02-06
searchAnnotations returns votes as annotations


mvdbster:
https://github.com/Esri/hub.js/blob/67016dc988d3ab435b3bea75795d905c9c9fa3cb/packages/annotations/src/search.ts#L79

`searchAnnotations` returns votes as `type: 'annotations'`.

There is a workaround by adding `parent_id IS NULL` to the `where` clause. It would probably be better to add this filter automatically?

jgravois:
> It would probably be better to add this filter automatically?

definitely. thanks for the report!

jgravois:
resolved in [`v1.7.1`](https://github.com/Esri/hub.js/releases/tag/v1.7.1) 🚀. just reach out if you have any questions/problems updating.

Issue #125, opened 2019-02-07
Annotations: Down Voting deletes features which is not possible with user role 'User' 


lisastaehli:
When a user is logged in with user role 'User' (can only add items, but not delete), the down voting does not work.

deleteFeatures will throw an error message `Operation is not supported`

jgravois:
that's strange. both downvoting and upvoting should _add_ a record. (with a `value` of -1 and 1 respectively). 

I'll dig in and investigate further when I'm back in the office tomorrow.

lisastaehli:
This happens when the same user first up votes an annotation and then down votes it (e.g. I changed my mind about it).

jgravois:
resolved in [`v1.7.1`](https://github.com/Esri/hub.js/releases/tag/v1.7.1) 🚀. just reach out if you have any questions/problems updating.

Issue #124, opened 2019-02-06
annotations: field name 'value' cannot be added on ArcGIS Online


lisastaehli:
When attempting to add the field `value` to the annotations layer on ArcGIS Online, an error message is displayed saying that `The field name cannot be the same as an SQL keyword` and the field cannot be added with this name:

![image](https://user-images.githubusercontent.com/27951811/52357583-b2ed2f80-2a36-11e9-8a93-b1c7919ec94f.png)

In ArcGIS Pro, the field name `value` can be added without any problem.

This is just meant to be an information and could be a reason to change this field name in the future to avoid issues with ArcGIS Online.

jgravois:
_very_ interesting. it'd probably be useful to tackle this sooner rather than later.

tomwayson:
Thanks @lisastaehli.

@jgravois how did add that field to our services?

I wonder if this error will come up when using the API (i.e. via #113)

jgravois:
in [`v1.7.1`](https://github.com/Esri/hub.js/releases/tag/v1.7.1) we switched to using `vote` as the fieldname (instead of `value`). 

_please_ reach out if you have any questions/problems updating.

Issue #118, opened 2019-01-29
update placeholder value in <meta> tag


jgravois:
![bad SEO](https://user-images.githubusercontent.com/3011734/51930989-a4f13c00-23b0-11e9-923b-4ac9573de51c.png)

https://github.com/Esri/hub.js/blob/198d4a49a4b0fd14c9d46c20ddd85fc72fb4a0f5/docs/src/_layout.html#L9

> A modular, high quality toolkit for working with [ArcGIS Hub](https://hub.arcgis.com/).

the above would be an improvement. bonus points for injecting something more specific in individual API ref pages.

PR #127, opened 2019-02-11
fix: value > vote, loosen delete protection, omit votes from search

jgravois:
* ensure that upvotes and downvotes aren't returned by search
* replace fieldname `value` with `vote`
* ensure that users can delete their own vote
* more descriptive `<meta>` tag in docs

@lisastaehli let me know if the second item in the list above is problematic. if you don't want to blow away your existing service entirely you should be able to add `vote` as an additional field in the ArcGIS Online UI. i'm happy to help if you run into any trouble.

AFFECTS PACKAGES:
@esri/hub-annotations

RESOLVES:
#123, #124, #125, #118

### DIFF
--- a/docs/src/_layout.html
+++ b/docs/src/_layout.html
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, user-scalable=no">
     <title>{{title if title}}{{" | " + titleSegments | join(" |") if titleSegments}}{{" | " if title}}@esri/hub.js</title>
     {% if description %}
-    <meta name="description" content="A modular, high quality toolkit for working with <a href='https://hub.arcgis.com/'>ArcGIS Hub</a>">
+    <meta name="description" content="A modular, high quality toolkit for working with ArcGIS Hub">
     {% endif %}
     <link rel="stylesheet" href="https://s3-us-west-1.amazonaws.com/patterns.esri.com/files/calcite-web/1.0.0-rc.7/css/calcite-web.min.css">
     <link rel="stylesheet" href="/hub.js/css/style.css">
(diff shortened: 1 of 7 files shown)


docs/src/_layout.html
```diff
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, user-scalable=no">
     <title>{{title if title}}{{" | " + titleSegments | join(" |") if titleSegments}}{{" | " if title}}@esri/hub.js</title>
     {% if description %}
-    <meta name="description" content="This is an example of a meta description. This will often show up in search results.">
+    <meta name="description" content="A modular, high quality toolkit for working with <a href='https://hub.arcgis.com/'>ArcGIS Hub</a>">
```
tomwayson: can meta tags contain anchor tags?

jgravois: > The value must be a free-form string that describes the page ... There must not be more than one `meta` element where the `name` attribute value is an ASCII case-insensitive match for `description` per document.

https://html.spec.whatwg.org/multipage/semantics.html#meta-description

i'm pretty sure that means "no" 🤔

lisastaehli:
I tried this and voting works fine, however the `outStatistics` are still on the wrong field `value`, so you would to change that as well to `vote` in order to fetch and summarize the votes correctly.

image

jgravois:
> `outStatistics` are still on the wrong field `value`

sorry about that @lisastaehli! i've released [`v1.7.2`](https://github.com/Esri/hub.js/releases/tag/v1.7.2) with a fix.

lisastaehli:
Awesome :) thanks. Now everything is working fine and we will migrate to 1.7.2

PR #128, opened 2019-02-12
chore: switch from coveralls to codecov

jgravois:
ref: https://github.com/Esri/arcgis-rest-js/pull/439


jgravois:
just waiting on our esteemed admins to grant access to the GitHub App.

PR #146, opened 2019-03-26
feat/registerForEvent

jgravois:
the (not so) heavy lifting for https://github.com/esridc/hub-components/pull/12

to do:
- [x] figure out just what API we hit to register folks for an event 😂 https://github.com/Esri/arcgis-rest-js/pull/494


jgravois:
obviously Travis won't greenlight this until we actually release Esri/arcgis-rest-js#494

packages/domains/package.json
```diff
@@ -16,7 +16,7 @@
     "@esri/hub-common": "^1.8.0"
```
tomwayson: Looks like this should be updated to 1.9.0 too.

packages/events/test/register.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    it("should help a user register for an event", done => {
+      fetchMock.once("*", { success: true, groupId: "5bc" });
+
+      registerForEvent({ groupId: "5bc", ...MOCK_REQOPTS })
+        .then(() => {
+          expect(fetchMock.called()).toEqual(true);
+          const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
+          expect(url).toEqual(
+            "https://myorg.maps.arcgis.com/sharing/rest/community/groups/5bc/join"
```
tomwayson: You're testing implementation details of rest-js. It would be better to stub `joinGroup()` and verify that it was called w/ the right args.

packages/events/test/register.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+    });
+    it("should help a user unregister for an event.", done => {
+      fetchMock.once("*", { success: true, groupId: "5bc" });
+
+      unregisterForEvent({ groupId: "5bc", ...MOCK_REQOPTS })
+        .then(() => {
+          expect(fetchMock.called()).toEqual(true);
+          const [url, options]: [string, RequestInit] = fetchMock.lastCall("*");
+          expect(url).toEqual(
```
tomwayson: better to stub `leaveGroup()` and verify that it was called w/ the right args.

packages/events/test/register.test.ts
jgravois: good catch.

i initially skipped this out of laziness, but to be fair, we couldn't have spied until `joinGroup` and `leaveGroup` landed anyway.

tomwayson: I figgered it was a 🐔 -> 🥚  situation

PR #149, opened 2019-04-02
added method to get rest api host url

jPurush:


packages/common/src/api.ts
```diff
@@ -0,0 +1,15 @@
+import { IRequestOptions, getPortalUrl } from "@esri/arcgis-rest-request";
+
+/**
+ * @param requestOptions - request options that may include authentication
+ */
+export function getHubRESTAPIBaseUrl(requestOptions: IRequestOptions): string {
```
jgravois: since we already have `getPortalUrl()` maybe just `getHubBaseUrl()`?

jPurush: yes! that sounds much much better. `getHubHostUrl` was something I was wondering about as well.

tomwayson: I think something w/ "API" is better than "base", maybe `getHubApiUrl()`

jgravois: `getHubUrl()`

tomwayson: I think it has to have `Api` in there.

jgravois: > I think it has to have Api in there.

works for me. a code snippet for the API reference would be nice though.

```js
/**
 * ```js
 * import { getHubApiUrl() } from "@esri/hub-common";
 * //
 * getHubApiUrl({ portal: "https://custom.maps.arcgis.com/sharing/rest" })
 * >> "https://hub.arcgis.com"
 * ```
 * Retrieves the Hub API Url associated with a specific ArcGIS Online organization.
 * @param requestOptions - request options that may include authentication
 * @returns the associated Hub API Url as a string.
 */

```

jgravois:
safe to close this one out, right?

jPurush:
> safe to close this one out, right?

@tomwayson felt this should just be left open. I'm ok with either

packages/common/src/api.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function getHubApiUrl(requestOptions: IRequestOptions): string {
+  const portalUrl = getPortalUrl(requestOptions);
+  if (portalUrl.match(/(qaext|\.mapsqa\.)/)) {
+    return "https://hubqa.arcgis.com";
+  } else if (portalUrl.match(/(devext|\.mapsdevext\.)/)) {
+    return "https://hubdev.arcgis.com";
+  } else if (portalUrl.match(/(www|\.maps\.)/)) {
+    return "https://hub.arcgis.com";
+  } else return undefined;
```
jPurush: - I removed the prod check and the undefined return locally. But however, if I pass in `https://randomsite.com/` instead of  `https://www.arcgis.com/sharing/rest` as `ro.authentication.portal` , I still get `https://hub.arcgis.com` which I think is incorrect. So I'm voting for keeping the `undefined` logic as it's not truly dead code. 

- In the current version of the function, if I pass in `https://www.randomsite.com/` as `ro.authentication.portal` , I get `https://hub.arcgis.com`-  is that expected?

PR #151, opened 2019-04-06
redirecting events to use hub api

jPurush:


jgravois:
these are the two untested branches of code.

Screenshot 2019-04-08 09 49 21

> else path not taken

packages/events/src/util.ts
```diff
@@ -56,11 +57,21 @@ export function getEventServiceUrl(
...
-      return url;
+      if (result && result.url) {
+        const host = getHubUrl(requestOptions);
+
+        // Extract the Event service's view name; the view returned depends
+        // on permission level of request user
+        const view = result.url.match(/services\/(.*?)\/FeatureServer/);
+
+        // Generate a root url for the hub-indexer event routes
+        /* istanbul ignore next */
```
jgravois: just curious, why the `/* istanbul ignore next */`?

jPurush: > just curious, why the `/* istanbul ignore next */` here?

Great catch! I meant to use `else`. Fixed it now

packages/events/src/util.ts
```diff
@@ -56,11 +57,21 @@ export function getEventServiceUrl(
         // pick the highest access level that this user has access to
         result = obj.admin || obj.org || /* istanbul ignore next */ obj.public;
       }
-      // single-layer service
-      let url = `${result.url}/0`;
-      // force https
-      url = url.replace(/^http:/gi, "https:");
-      return url;
+      if (result && result.url) {
```
tomwayson: We didn't check for `result && result.url` before, so we can probably remove this check from this PR. By that I mean the old code `let url = `${result.url}/0`;` would have thrown if `result` was falsey and would have returned `/0` URL if `result.url` was falsey.

I think trying to handle this edge case now adds a lot of complexity to this PR, and I don't think we should try to handle that case until it comes up as problem and we've decided how it should be handled (i.e. is returning `null` the right thing to do, or should it throw an error?).

PR #152, opened 2019-04-09
search package

pranavkulkarni:
Module to search for ArcGIS items and format them for display in ArcGIS Hub

Structure --
```
src
  ago
    helpers
      aggs
      filters
      format
      page
      sort
    search.ts // searches for items and returns raw ago results
    params.ts // defines ISearchParams interface to hub.js/search
    serialize.ts // serialize raw UI params into query parameters that V3 understands
    format-item-collection.ts // format raw ago results into hub v3 style datasets
    compute-items-facets.ts // compute aggs/facets based on ago results and also calculate aggs for custom facets
```

Sample of how this will be used in UI code --

```
const authentication = new arcgisRest.UserSession({
      portal: 'https://qaext.arcgis.com/sharing/rest',
      username: 'dcadminqa',
      password: 'dcadminqa1'
    })
    const portal = 'https://qaext.arcgis.com/sharing/rest'
    const params = {
      q: 'crime',
      sort: 'name',
      agg: { fields: 'tags,collection,owner,source,hasApi,downloadable' },
      start: 1,
      num: 10
    }
    authentication.getToken(portal)
    .then(token => {
      const headers = { authorization: token, portal }
      const serializedParams = arcgisHub.serialize(params)
      console.log(serializedParams)
      return fetch(`hub.arcgis.com/api/v3/search?${serializedParams}`, { headers })
    })
    .then(unifiedResultsPromise => {
      return unifiedResultsPromise.json()
    })
    .then(unifiedResults => {
      console.log(unifiedResults)
      // UPDATE UI model with unified results
    })
```

Sample of how hub indexer api v3 new search route will use hub.js --

```
   const search = require('@esri/hub-search')

  // 1. perform ago search with facets and formatting
  const agoResults = await search.agoSearch(req.query, req.headers.authorization, req.headers.portal)

  // 2. compute items facets
  const facets = await search.computeItemsFacets(agoResults.aggregations, req.query, req.headers.authorization, req.headers.portal)

  // 3. format item collection based on actual search results and computed facets
  const model = search.agoFormatItemCollection(agoResults, facets, req.query)

  return model
```


jPurush:
Can we add a README similar to [one](https://github.com/Esri/hub.js/blob/master/packages/auth/README.md) from the other packages ?

pranavkulkarni:
Yes, I will once this PR is complete. The scope of this package is still in the works.

.gitignore
```diff
@@ -80,4 +80,4 @@ packages/*/debug/
 /packages/*/.rpt2_cache
 
 ## for john to poke at
-test.html
```
jgravois: we shouldnt stop ignoring this file.

packages/search/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@esri/hub-search",
+  "version": "1.10.0",
+  "description": "Module to search ArcGIS items.",
```
jgravois: maybe this?

> Module to search for ArcGIS items and format them for display in ArcGIS Hub.

packages/search/src/ago/format-item.ts
```diff
@@ -0,0 +1,64 @@
...
+  const hubTypes = hubTypeLookup(item.type);
+  return Object.assign({}, item, {
+    /* computed or null attributes so items & datasets look the same */
+    name: item.title,
+    searchDescription: item.description,
+    hubType: hubTypes[0] || "Other"
+  });
+}
+
+function highlights(item: IItem, query: string) {
```
jgravois: if you doc the return type, it'll show up in the API reference automatically.
```ts
function highlights(item: IItem, query: string):{ name: string, searchDescription: string } {
```

packages/search/src/ago/encode-ago-query.ts
```diff
@@ -0,0 +1,43 @@
+import { IQueryObject } from "./create-q";
+import { calcStart, handleFilter, getSortField } from "./helpers";
+
+export function encodeAgoQuery(queryObject: IQueryObject) {
```
jgravois: i dont think `encode` the right word for this. maybe `concatenate` would be clearer?

packages/search/src/ago/search.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Search for Items in ArcGIS
+ *
+ * @export
+ * @param {IParams} searchRequestOptions
+ * @returns {Promise<ISearchResult>}
+ */
+export function search(
+  params: ISearchParams,
+  authentication: UserSession
+): Promise<ISearchResult> {
```
jgravois: the underlying package `request` supports other requestOptions like passing a custom fetch implementation and custom headers so your signature that strips things down to `params` and `authentication` only is problematic.

@patrickarlt just so happens to have started work on a search param builder of his own in https://github.com/Esri/arcgis-rest-js/pull/509 so we should _really_ try to pool effort.

packages/search/src/common/create-filters.ts
```diff
@@ -0,0 +1,45 @@
+import { ISearchParams } from "./params";
+import { paramSchema } from "./param-schema";
+
+/**
+ * Create filters
+ *
+ * @export
```
jgravois: no need for a JSDoc `@export` declaration, TSDoc picks up on the keyword in the actual code.

packages/search/src/common/params.ts
```diff
@@ -0,0 +1,27 @@
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+* Apache-2.0 */
+
+export interface ISearchParams {
```
jgravois: we actually consider it an anti-pattern to extend [`IParams`](https://esri.github.io/arcgis-rest-js/api/request/IParams). 

see https://github.com/Esri/arcgis-rest-js/pull/78 for more info.

packages/search/src/index.ts
```diff
@@ -0,0 +1,7 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+export {
+  search as agoSearch,
```
jgravois: if you're going to export it as `agoSearch`, why not just give the function that name to begin with?

tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "module": "es2015",                     /* Specify module code generation: 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
     "lib": [
       "dom",
-      "es2015"
+      "es2015",
+      "es2017"
```
jgravois: i'll have to do a little digging to understand what the implications of this are.

just curious, what made you reach for it?

.gitignore
pranavkulkarni: Yes, I have been testing with that. I remember to add it back before merging

packages/search/src/ago/format-item.ts
pranavkulkarni: sweet. I'll add the niceties once the basics are in place

packages/search/src/index.ts
pranavkulkarni: yeah good point

tsconfig.json
pranavkulkarni: I was not able to `array.includes` or `string.includes`. Adding es2017 worked.

packages/search/src/ago/search.ts
pranavkulkarni: I just took a look at Patrick's work and I think it can work well in building search queries in this library

packages/search/src/ago/encode-ago-query.ts
pranavkulkarni: or may be `createQuery`. I think Daniel had said something like that.

packages/search/src/common/params.ts
pranavkulkarni: I am not sure I understand this correctly. I am not extending `IParams` here. So it is good, right?

jgravois: > I am not extending IParams

you've got me there.

the point is that you want your new method needs to accept an instance of `IRequestOptions` to pass through to `request`.

this means that you'd be better served to extend `IRequestOptions` with new custom props and pass them down internally yourself.

check out pat's example here: https://github.com/Esri/arcgis-rest-js/pull/78#issuecomment-352124790

.gitignore
jgravois: for sure.

the point of the file is to have something to use for local testing that doesn't ever get committed.

tsconfig.json
jgravois: > Adding `es2017` worked.

Unfortunately we need to ask TypeScript to target `es5`. Adding `es2017` as an additional lib enables support for the features you want in TypeScript-land, but it also means that consumers of our lib would be required to bring their own polyfill to the party because TypeScript doesn't transpile modern JS the way babel does.

i can't believe i'm linking to a reddit thread as supporting evidence, but it confirmed my hunch.

* https://stackoverflow.com/questions/41173215/does-typescript-transpilation-handle-transpiling-es6-to-es5
* https://www.reddit.com/r/typescript/comments/8gh2rt/any_way_to_use_es78_syntax_in_typescript_without/

i feel you though. i've reached for `.includes()` myself more than once here too.

packages/search/src/common/hub-type-map.ts
```diff
@@ -0,0 +1,117 @@
+interface IHubTypeMap {
```
haoliangyu: Let's reuse the existing code at https://github.com/Esri/hub.js/blob/master/packages/common/src/categories.ts or https://www.npmjs.com/package/@esri/arcgis-hub-content-types.

packages/search/src/ago/search.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @export
+ * @param {IParams} searchRequestOptions
+ * @returns {Promise<ISearchResult>}
+ */
+export function agoSearch(
+  params: ISearchParams,
+  authentication: UserSession
+): Promise<ISearchResult> {
+  // console.log("Inside hubJS params = ", params);
+  const queryObj: IQueryObject = createQueryObject(params);
```
haoliangyu: Is the `params -> query object -> ago params` pipeline necessary? Can we remove the intermediate step?

packages/search/src/ago/format-item.ts
```diff
@@ -0,0 +1,67 @@
...
+import { IItem } from "@esri/arcgis-rest-common-types";
+
+export function formatItem(item: IItem, query: string) {
+  const formattedItem: any = {
+    id: item.id,
+    type: "item",
+    attributes: formatItemAttributes(item)
+  };
+  if (query) {
+    if (!item.meta) {
```
haoliangyu: The `IItem` interface does not define the `meta` key. Where is it from?

packages/search/src/common/params.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/* Copyright (c) 2017 Environmental Systems Research Institute, Inc.
+* Apache-2.0 */
+
+export interface ISearchParams extends IRequestOptions {
+  access?: string;
+  agg?: { fields: string[] };
+  bbox?: string;
+  catalog?: { groupIds: string[]; pageIds: string[] };
+  catalogToggle?: boolean;
```
haoliangyu: This is a parameter to control the UI. It can be removed.

jgravois:
i accidentally dismissed @haoliangyu's review, but i _think_ all his feedback was addressed anyway.

@pranavkulkarni is this PR ready for another round of code review?

pranavkulkarni:
@jgravois this PR is ready

packages/search/src/ago/helpers/filters/has-api.ts
```diff
@@ -0,0 +1,21 @@
+import { getProp } from "@esri/hub-common";
+const apiTypes = ["Feature Service", "Map Service", "Image Service"];
+
+export function hasApi(queryFilters: any) {
```
jgravois: when this library is bundled as a UMD all the methods hang off of an `arcgisHub` namespace. 

this means that in the browser, `filters.hasApi` will collide with `aggs.hasApi`

jgravois:
can you give this a once over too @tomwayson?

pranavkulkarni:
Thanks @jgravois I will address the points you raised here. I will also beautify comments.

packages/search/src/ago/helpers/filters/has-api.ts
```diff
@@ -0,0 +1,21 @@
+import { getProp } from "@esri/hub-common";
+const apiTypes = ["Feature Service", "Map Service", "Image Service"];
```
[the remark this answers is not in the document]
pranavkulkarni: Good catch! I'll move it to hub-common

packages/search/src/ago/helpers/filters/has-api.ts
pranavkulkarni: Interesting... the UI is not supposed to calls these functions directly. They are being used by `agoSearch` and `computeItemsFacets` that are exposed by this library and the UI will talk to just those functions. Do I have still rename them?

jgravois: i didn't see before that you included `"hasApi"` in the blacklist. thats definitely good.

even if callers that load the UMD don't call these methods directly though, things will go pear shaped when they're called _internally_, because i don't think there is a mechanism in place in that build to differentiate them.

you have two options:
1. make the names unique, just to be safe.
2. test loading the UMD in the browser and calling both methods that lean on `hasApi` internally to confirm that i'm wrong.

package-lock.json
```diff
@@ -8116,6 +8116,7 @@
             "color-convert": "^1.9.0"
           }
         },
+<<<<<<< HEAD
```
jgravois: this is what Travis is choking on.

besides fixing it manually, you're also welcome to just delete the file and rerun `npm install` in the root of the repo to recreate it.

jgravois:
i'll release Hub.js `v2.1.0` tomorrow.

PR #163, opened 2019-05-21
search improvements

pranavkulkarni:
In this PR --

1. Developed an algorithm to encode urls cleanly in serialize function. The crux of the algorithm is to handle filters and non-filters differently. It can URI encode any nested object like `{ a: { b: 2 }, c: 3 }` as `a[b]=2&c=3`
2. Removed circular dependency in computeItemsFacets function. I have addressed it by creating a `get-items` module shared between `agoSearch` and `computeItemsFacets`.
3. Fixed a bug wherein a blank string `""` didn't construct AGO query properly. So I have added a check in place for that.

### DIFF
--- a/packages/search/src/ago/helpers/filters/encode-params.ts
+++ b/packages/search/src/ago/helpers/filters/encode-params.ts
@@ -1,0 +1,77 @@
+import { getProp } from "@esri/hub-common";
+
+/**
+ * Url-encoding of search params. This function is generic enough to encode a deeply nested object
+ * ```
+ * Example:
+ * Input: { a: { b: 2 }, c: 3 }
+ * Output: 'a[b]=2&c=3'
+ * ```
+ * @param {Any} params (query params from hub indexer)
+ * @returns {String}
+ */
+export function encodeParams(params: any = {}): string {
+  // get raw paths
+  const paths = getPaths(params);
+  const flatPaths = paths.filter(path => {
+    return typeof getProp(params, path.join(".")) !== "object";
+  });
+  const parts: string[] = [];
+  // for each nested path, we want to surround it with `[]`
+  // i.e. if a path is like ['a', 'b'], we want encoding as 'a[b]=2' given the input object { a: { b: 2 }, c: 3 }
+  flatPaths.forEach(path => {
+    let str = "";
+    for (let i = 0; i < path.length; i++) {
+      if (i === 0) {
+        str += path[i];
+      } else {
+        str += `[${path[i]}]`;
+      }
+    }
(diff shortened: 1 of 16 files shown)


packages/search/src/ago/helpers/filters/encode-params.ts
```diff
@@ -0,0 +1,77 @@
...
+    for (let i = 0; i < path.length; i++) {
+      if (i === 0) {
+        str += path[i];
+      } else {
+        str += `[${path[i]}]`;
+      }
+    }
+    const right = encodeURIComponent(getProp(params, path.join(".")));
+    const left = encodeURIComponent(str);
+    if (right) {
```
haoliangyu: Can the value be `null` or `undefined`? Because

```
encodeURIComponent(null)
> "null"

encodeURIComponent(undefined)
> "undefined"
```

Also, it is better to move this to the beginning of the function.

packages/search/src/ago/helpers/filters/encode-params.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function getPaths(root: any = {}): string[][] {
+  const paths: string[][] = [];
+  const nodes: any[] = [
+    {
+      obj: root,
+      path: []
+    }
+  ];
+  while (nodes.length > 0) {
```
haoliangyu: Good use of DFS :+1:

packages/search/src/ago/helpers/filters/encode-params.ts
```diff
@@ -0,0 +1,77 @@
...
+    const n = nodes.pop();
+    Object.keys(n.obj).forEach(k => {
+      if (typeof n.obj[k] === "object") {
+        const path = n.path.concat(k);
+        paths.push(path);
```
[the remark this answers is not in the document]
pranavkulkarni: I wanted to keep this a generic function that gives paths to all properties (nodes) of an object.

packages/search/src/ago/helpers/filters/encode-params.ts
pranavkulkarni: It will never be null or undefined because the path is calculated on params. So the path will be present in the params object. I will add an extra check like 
```
const right = encodeURIComponent(getProp(params, path.join(".")) || "");
```

pranavkulkarni: > Also, it is better to move this to the beginning of the function.

Sorry, I didn't understand that comment ^

haoliangyu: I mean having an earlier exit, but it doesn't really matter.

PR #165, opened 2019-06-03
f/initiative followers group

mjuniper:
Updates follow initiative to account for new initiative followers group.

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -278,23 +278,3 @@ export function camelize(value: string): string {
 export function unique(value: any, index: number, ary: any[]): boolean {
   return ary.indexOf(value) === index;
 }
-
-/**
- * Concatenates two arrays
- * Allows for code like:
- *
- * ```js
- * // flatten an array
- * const ary = [ [ 1, 2, 3 ], [ 3, 4, 5] , [ 1 ] ];
- *
- * const result = ary.reduce(concat);
- *
- * result => [ 1, 2, 3, 3, 4, 5, 1 ]
- * ```
- * @param ary1 - an array to concatenate
- * @param ary2 - an array to concatenate
- * @returns {any[]} - a concatenated array
- */
-export function concat(ary1: any[], ary2: any[]): any[] {
-  return [...ary1, ...ary2];
-}
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -258,3 +258,43 @@ export function camelize(value: string): string {
       return match.toLowerCase();
     });
 }
+
+/**
+ * Determines if a value is unique in an array
+ * Allows for code like:
+ *
+ * ```js
+ * const ary = [ 1, 2, 3, 3, 4, 5, 1 ];
+ *
+ * const result = ary.filter(unique);
+ *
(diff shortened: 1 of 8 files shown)


packages/common/src/util.ts
```diff
@@ -258,3 +258,43 @@ export function camelize(value: string): string {
...
+ *
+ * const result = ary.reduce(concat);
+ *
+ * result => [ 1, 2, 3, 3, 4, 5, 1 ]
+ * ```
+ * @param ary1 - an array to concatenate
+ * @param ary2 - an array to concatenate
+ * @returns {any[]} - a concatenated array
+ */
+export function concat(ary1: any[], ary2: any[]): any[] {
```
tomwayson: I think we should ditch this for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

packages/initiatives/src/follow.ts
```diff
@@ -1,18 +1,46 @@
...
 export interface IFollowInitiativeOptions extends IRequestOptions {
   initiativeId: string;
   authentication: UserSession;
 }
 
-const getTag = (initiativeId: string) => `hubInitiativeId|${initiativeId}`;
+const getUserTag = (initiativeId: string) => `hubInitiativeId|${initiativeId}`;
+
+const initiativeIdFromUserTag = (tag: string) =>
+  tag.replace(/^hubInitiativeId\|/, "");
```
tomwayson: should probably add a `tag &&` before `tag.replace()`

packages/initiatives/src/follow.ts
```diff
@@ -61,34 +126,67 @@ export function followInitiative(
...
-    // https://stackoverflow.com/questions/9792927/javascript-array-search-and-remove-string
-    const index = tags.indexOf(tag);
-    tags.splice(index, 1);
+  })
+    .then(user => {
+      // don't update if not already following
+      if (!isUserFollowing(user, requestOptions.initiativeId)) {
+        return Promise.reject(`user is not following this initiative.`);
+      }
+      // if not already following, pass the user on
```
jPurush: I think you mean `// if already following `

packages/common/src/util.ts
mjuniper: Well it's not quite the same. the reason i want a concat function with that signature (`concat(ary1, ary2)`)
is so i can have a 2 dimensional array and call `my2dAry.reduce(concat, [])` to flatten it.

packages/initiatives/src/follow.ts
```diff
@@ -61,34 +126,67 @@ export function followInitiative(
...
+      return user;
+    })
+    .then(user => {
+      const tag = getUserTag(requestOptions.initiativeId);
+      const tags = JSON.parse(JSON.stringify(user.tags));
```
dbouwman: I think you could use `cloneObject` here... but since the tags array is primitive types I don't think you'd really need to b/c those are by-val anyhow... I think :)

packages/common/src/util.ts
tomwayson: maybe we should have a `flattenArray()` fn instead?

PR #167, opened 2019-06-04
Pass authentication to getUser function

alexus37:


tomwayson:
Thanks @alexus37 your commits will land in #170!

PR #171, opened 2019-06-17
Bug/get users with portal

sandromartis:
PR #170 did not solve our issues since the portal has to be passed to `getUser` for anonymous fetching of annotations.

This should solve it.


tomwayson:
This was released in v2.2.4

PR #172, opened 2019-06-21
F/type categories

pranavkulkarni:
- Add collection aka typeCategories attribute to AGO items. Logic is picked from https://github.com/ArcGIS/hub-indexer/blob/master/packages/duke/utils/determine-type.js and https://github.com/ArcGIS/hub-indexer/blob/master/packages/duke/compose/dataset.js#L150
- Fix a mistake in docs

### DIFF
--- a/packages/common/src/content.ts
+++ b/packages/common/src/content.ts
@@ -53,16 +53,11 @@ export function getTypes(category: string = ""): string[] {
  * getType(item)
  * > [ 'Hub Site Application' ]
  * ```
- * To do.
- * @param layer Layer object.
  * @param item Item object.
- * @returns type of the input layer/item.
+ * @returns type of the input item.
  *
  */
-export function getType(layer: any = {}, item: any = {}): string {
-  if (layer.type) {
-    return layer.type;
-  }
+export function getType(item: any = {}): string {
   const typeKeywords = item.typeKeywords || [];
   if (item.type === "Web Mapping Application") {
     if (typeKeywords.includes("hubSite")) {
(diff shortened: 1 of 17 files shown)


packages/search/src/ago/helpers/format/format-item.ts
```diff
@@ -48,7 +48,8 @@ function formatItemAttributes(item: IItem) {
     // computed or null attributes so items & datasets look the same
     name: item.title,
     searchDescription: item.description,
-    hubType: hubType || "Other"
+    hubType: hubType || "Other",
+    collection: getTypeCategories({}, item)
```
jPurush: should we consider that `collection` could be added to item attributes in the future?

pranavkulkarni: Good question - I don't think collection will be part of AGO because it is a hub specific terminology. May be @dmfenton can weigh in on this.

packages/common/src/content.ts
```diff
@@ -45,3 +45,57 @@ export function getCategory(itemType: string = ""): string {
 export function getTypes(category: string = ""): string[] {
   return categories[category.toLowerCase()];
 }
+
+/**
+ * ```js
+ * import { getType } from "@esri/hub-common";
+ * //
+ * getType(item)
```
tomwayson: this example usage doesn't seem to match the signature below, seems like this should be `getType(layer)` which yields a string layer type, not `[ 'Hub Site Application' ]`.

Also a description of the intended use would help me understand this better - if `layer` is the first argument, I assume this is _primarily_ used to get the layer type.

packages/common/src/content.ts
```diff
@@ -45,3 +45,57 @@ export function getCategory(itemType: string = ""): string {
...
+ * > [ 'Hub Site Application' ]
+ * ```
+ * To do.
+ * @param layer Layer object.
+ * @param item Item object.
+ * @returns type of the input layer/item.
+ *
+ */
+export function getType(layer: any = {}, item: any = {}): string {
```
tomwayson: Conceptually, do we have our own types? Or is it more like we use a superset of layer/item types? If it's the former, maybe a better name of this fn would be `getHubType()`?

packages/common/src/content.ts
pranavkulkarni: I was trying to accommodate what hub indexer currently does into hub.js. Actually for formatting AGO item results, we won't have a layer object, but we use the same function in hub indexer for layer and item slightly differently. How about I do just `getItemType(item)` which does not deal with layer?

Issue #145, opened 2019-03-20
Create initiative from blank (don't require a template)


ajturner:
We need to support creating a _blank_ initiative that doesn't require activating from a template.

robertsteilberg:
This is now possible!

PR #182, opened 2019-08-19
Create blank initiative

robertsteilberg:
This refactor creates a new orchestration method for creating the two groups required for activating an initiative so that it can handle users who do not have all the privs necessary to create
groups. It also allows activateInitiative to activate an embedded initiative by adding support for copying embedded image resources to AGO.

AFFECTS PACKAGES:
@esri/hub-initiatives

ISSUES CLOSED: #145

### DIFF
--- a/packages/initiatives/src/activate.ts
+++ b/packages/initiatives/src/activate.ts
@@ -4,21 +4,20 @@
 import { IRequestOptions } from "@esri/arcgis-rest-request";
 import { IInitiativeModel, createId, camelize } from "@esri/hub-common";
 import { getInitiative } from "./get";
-import { getUniqueGroupName, createInitiativeGroup } from "./groups";
+import { createInitiativeGroups } from "./groups";
 import {
   createInitiativeModelFromTemplate,
   IInitiativeTemplateOptions
 } from "./templates";
 import { addInitiative } from "./add";
-import { copyImageResources } from "./util";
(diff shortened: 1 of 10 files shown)


packages/initiatives/src/activate.ts
```diff
@@ -4,21 +4,20 @@
...
   shareItemWithGroup,
-  IGroupSharingOptions,
-  getSelf,
-  IPortal
+  IGroupSharingOptions
 } from "@esri/arcgis-rest-portal";
+import { getProp } from "@esri/hub-common";
 
-const steps = [
+export const steps = [
```
robertsteilberg: Exported so we can properly test the callback

packages/initiatives/src/groups.ts
```diff
@@ -7,10 +7,91 @@ import {
...
+ * @param {string} collaborationGroupName name of the collaboration group to be created
+ * @param {string} dataGroupName name of the content group to be created
+ * @param {IRequestOptions} ro
+ * @returns {Promise<any>} object containing the groupIds
+ */
+export function createInitiativeGroups(
+  collaborationGroupName: string,
+  dataGroupName: string,
+  ro: IRequestOptions
+): Promise<any> {
```
robertsteilberg: Made some liberal use of `any` in this fn... is that ok?

PR #186, opened 2019-08-30
fix(search): multiple filters should be ANDed together

dmfenton:
when searhcing for thigns tagged "tres" and things of type "web map", the generated query should be "type: web map AND tags: trees"

currently they ared ORed together so you get thigns that are web maps or trees


pranavkulkarni:
Where’s the actual code change in handle-filter.ts file in that PR? the change from OR to AND

dmfenton:
good call, I accidentally left it out

PR #192, opened 2019-09-09
fix(search): format item extent properly

dmfenton:
item extents are now formatted as from the Hub API
```js
{
 type: 'envelope',
 coordinates: [ [1,2], [3,4] ]
}
```


dmfenton:
@tomwayson will do. it's going to come in my /search on enterprise sites PR sometime today

PR #204, opened 2019-10-31
feat(filter-between-dates): filter AGO items by modified date

pranavkulkarni:
filter AGO items by modified date

AFFECTS PACKAGES:
@esri/hub-common
@esri/hub-search

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -309,3 +309,19 @@ export function extend(
     return obj;
   }, extended);
 }
+
+/**
+ * Add number of days to a given date
+ *
+ * @export
+ * @param {string} date
+ * @param {number} numOfDays
+ * @returns {number} unix timestamp with numOfDays added to the given date
+ */
+export function addDays(date: string, numOfDays: number): string {
+  const given = new Date(date);
+  const dateString = new Date(
+    given.setDate(given.getDate() + numOfDays)
+  ).toISOString();
+  return dateString.split("T")[0];
+}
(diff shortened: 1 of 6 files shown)


pranavkulkarni:
> I am not very familiar with this codebase. But I have two questions:
> 
> 1. AGO range queries can be used on non-date fields. Does this preclude allowing for that in the future?
No. It is flexible for any range query. We are following what v3 api does like `/api/v3/search?filter[modified]=between(2019/10/30,2019/10/31)`

> 2. Wouldn't it be better to accept dates as a timestamp or a date object rather than a string?
Hub.js is used to serialize query parameters in the UI because of unified search. The same query params are used to search against v3 backend and against AGO using this hub-search package. The serialized query params appear in the URL so I think we decided to go with dates like `2019/10/31` rather than unix timestamp. The v3 backend converts that to unix timestamp automatically thanks to Elasticsearch. We need the same thing to happen in hub.js which is used by the backend to query AGO.

packages/common/src/util.ts
```diff
@@ -309,3 +309,19 @@ export function extend(
...
+/**
+ * Add number of days to a given date
+ *
+ * @export
+ * @param {string} date
+ * @param {number} numOfDays
+ * @returns {string} date string with numOfDays added to the given date
+ */
+export function addDays(date: string, numOfDays: number): string {
+  const given = new Date(date);
```
trescube: Is there error handling in place to cover the case where `new Date(date)` returns `Invalid Date`?

pranavkulkarni: Yes, It is handled by hub-indexer here - https://github.com/ArcGIS/hub-indexer/blob/master/packages/api/search/items.js#L27-L32

trescube: It's not going to get that far:  

```
> addDays('this is not a date', 1)
RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at addDays (repl:5:5)
```

PR #209, opened 2019-11-18
feat(support-more-than-3-countFields): support-more-than-3-countFields

pranavkulkarni:
AGO supports max 3 countFields in a request. Hub search should send multiple requests for countFields > 3, collect responses and return

AFFECTS PACKAGES:
@esri/hub-common
@esri/hub-search

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -338,8 +338,10 @@ export function addDays(date: string, numOfDays: number): string {
  */
 export function chunkArray(arr: any[], size: number) {
   const results = [];
-  while (arr.length) {
-    results.push(arr.splice(0, size));
+  let index = 0;
+  while (index < arr.length) {
+    results.push(arr.slice(index, index + size));
+    index += size;
   }
   return results;
 }
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -329,3 +329,17 @@ export function addDays(date: string, numOfDays: number): string {
     throw new Error("Invalid Date");
   }
 }
+
+/**
+ * Returns an array with arrays of the given size.
+ *
+ * @param arr Array to split
+ * @param size Size of every group
+ */
+export function chunkArray(arr: any[], size: number) {
+  const results = [];
+  while (arr.length) {
+    results.push(arr.splice(0, size));
+  }
+  return results;
+}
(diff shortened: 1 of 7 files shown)


packages/common/src/util.ts
```diff
@@ -329,3 +329,17 @@ export function addDays(date: string, numOfDays: number): string {
...
   }
 }
+
+/**
+ * Returns an array with arrays of the given size.
+ *
+ * @param arr Array to split
+ * @param size Size of every group
+ */
+export function chunkArray(arr: any[], size: number) {
```
mikeringrose: i'd definitely comment that arr is mutated. prefer immutability where possible.
```function chunkArray(array, size) {
  const results = [];
  let index = 0;
  while (index < array.length) {
    results.push(array.slice(index, size + index));
    index += size;
  }
  return results;
}
```

packages/search/src/ago/get-items.ts
```diff
@@ -11,14 +12,47 @@ export async function getItems(
...
-    params: {
-      token,
-      countFields: agoParams.countFields,
-      countSize: agoParams.countSize
-    },
-    portal,
-    authentication
-  });
+  if (agoParams.countFields) {
+    // if countFields are defined
```
mikeringrose: delete comment, the line is pretty obvious

packages/search/src/ago/get-items.ts
```diff
@@ -11,14 +12,47 @@ export async function getItems(
...
+    // AGO allows only 3 countFields at max
+    const chunkedCountFields = chunkArray(
+      agoParams.countFields.split(","),
+      3
```
mikeringrose: i'd define as a constant `const MAX_COUNTFIELDS = 3` and then delete the comment

packages/search/src/ago/get-items.ts
```diff
@@ -11,14 +12,47 @@ export async function getItems(
...
+    ).map(fieldArrayChunk => fieldArrayChunk.join(","));
+    let allCounts: Array<{
+      fieldName: string;
+      fieldValues: Array<{
+        value: any;
+        count: number;
+      }>;
+    }> = [];
+    let results;
+    for (const chunk of chunkedCountFields) {
```
mikeringrose: Minor: could just do a `Promise.all` with the searchItems requests and let the execution potentially happen in parallel

packages/search/src/ago/get-items.ts
```diff
@@ -11,14 +14,48 @@ export async function getItems(
...
+    });
+    const responses = await Promise.all(promises);
+    let allCounts: Array<{
+      fieldName: string;
+      fieldValues: Array<{
+        value: any;
+        count: number;
+      }>;
+    }> = [];
+    for (const response of responses) {
```
mikeringrose: could just do
```
const results = responses[0];
results.aggregations.counts = responses.map(response => getProp(response, "aggregations.counts") || []);
return results;
```

PR #211, opened 2019-12-21
Exactly matching group titles for existance check

drewctate:
[bug](https://esriarlington.tpondemand.com/entity/104035-bug-000126797-unique-initiative-name-detection)

The `checkGroupExists` function is used by Hub to check for initiatiative existance by searching for stuff like "My Initiative Name Core Team." But the way `checkGroupExists` is currently set up, it will match the group name fuzzily. This PR forces it to match group names exactly.

### DIFF
--- a/packages/initiatives/package-lock.json
+++ b/packages/initiatives/package-lock.json
@@ -1,10 +1,4 @@
 {
-<<<<<<< HEAD
-=======
-	"name": "@esri/hub-initiatives",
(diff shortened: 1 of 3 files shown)


dbouwman:
Thanks @benstoltz - I think this is fine as it's a single-purpose function, and exact matching is an improvement.

benstoltz:
Well then 👍 once CI is green @drewctate

PR #219, opened 2020-02-07
Changing tags filter to default to any

mikeringrose:
Filters should be consistent and all use the same default values. This PR updates the `tags` filter to `any` to be consistent with the others.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@esri/hub.js",
-  "version": "3.6.5",
+  "version": "3.6.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@esri/hub.js",
-  "version": "3.6.5",
+  "version": "3.6.6",
```
haoliangyu: I think the version number shouldn't be updated until release.

trescube: @haoliangyu  is correct, the deploy process updates this.

mikeringrose: good to know.

PR #223, opened 2020-02-27
chore(migrate initiative item to create recommendedTemplates array): …

robertsteilberg:
migrate initiative item to create recommendedTemplates array

This migration bumps the initiative item schema version from 2.1 to 2.2, though it doesn't yet set that prop because we need the migration to run every time an initiative is loaded. It will traverse the
initiative item's steps array, adding any present template ids to an array called
recommendedTemplates that sits at the model.data node. Importantly, this migration only runs once,
so subsequent templates added to the steps array post-migration will not be re added to
recommendedTemplates.

AFFECTS PACKAGES:
@esri/hub-initiatives


packages/initiatives/src/migrator.ts
```diff
@@ -6,12 +6,13 @@ import { applyInitialSchema } from "./migrations/apply-schema";
...
 import { upgradeToTwoDotZero } from "./migrations/upgrade-two-dot-zero";
 import { upgradeToTwoDotOne } from "./migrations/upgrade-two-dot-one";
+import { upgradeToTwoDotTwo } from "./migrations/upgrade-two-dot-two";
 
 /**
  * Current Schema Version
  * @protected
  */
-export const CURRENT_SCHEMA_VERSION = 2.1;
+export const CURRENT_SCHEMA_VERSION = 2.2;
```
dbouwman: Should this stay at 2.1?

robertsteilberg: fixed

PR #230, opened 2020-03-30
Fixed import statement

drspacemanphd:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@esri/hub.js",
-  "version": "3.6.5",
+  "version": "3.6.6",
```
trescube: back this out, there's a process that semantically versions this automatically.

packages/search/src/ago/encode-ago-query.ts
```diff
@@ -57,7 +57,15 @@ export function encodeAgoQuery(queryParams: any = {}) {
     }
   }
   if (queryParams.agg && queryParams.agg.fields) {
-    const { countFields, countSize } = createAggs(queryParams.agg.fields);
+    // fields may be passed as array of fields, rather than comma-separated string
+    // if so join fields to string, else leave as is
+    let fields;
+    if (_.isArray(queryParams.agg.fields)) {
```
trescube: No need for lodash here, just use `Array.isArray`.  It's not bullet-proof, but should work in this constrained environment.

trescube: Alternatively, you could use `_.castArray` that will convert to an array if it isn't already one.

Issue #248, opened 2020-05-08
Potential POA for excluding fetch-blob from UMD builds


drewctate:
Fetch-blob is a Node implementation of `Blob`. UMD builds are never run in a node environment. Therefore, excluding `fetch-blob` from UMD builds would both get rid of some tooling warnings (https://github.com/Esri/solution.js/issues/316) and slightly shrink our bundle size.

To do this, we could use the `Rollup` alias plugin to direct the fetch blob import to an empty module like so:
```
import alias from '@rollup/plugin-alias';

module.exports = {
  input: 'src/index.js',
  output: {
    dir: 'output',
    format: 'cjs'
  },
  plugins: [
    alias({
      entries: [
        { find: 'fetch-blob', replacement: './dummy-module' },
      ]
    })
  ]
};
```

This would allow the code to be none-the-wiser. The [`fetch-blob` import](https://github.com/Esri/hub.js/blob/246a0671dd78023731e1e79d5c11b4bfeac56be6/packages/common/src/resources/_which-blob/_which-blob.ts#L1-L2) would remain syntactically unchanged, only redirected to an empty module for UMD purposes.

tomwayson:
I'm not sure this issue is isolated to creating UMD builds.

While working on #250 I created a test Ember app w/ [the same Hub.js/rest-js dependencies as the Hub](https://github.com/ArcGIS/opendata-ui/blob/d2f44142e5a72b601352b7ba4302da6a90eda33c/packages/opendata-ui/package.json#L216-L227), and then added an application controller w/:

```js
import { _checkStatusAndParseJson } from '@esri/hub-sites';
import { getItems } from '@esri/hub-search';
```

and when I run a build via `ember s` I get:

```bash
ERROR in ./node_modules/fetch-blob/index.js
Module not found: Error: Can't resolve 'stream' in '/Users/tom/test/tslib-test/node_modules/fetch-blob'
```

I'm not sure how the Hub ember app's build is even working right now.

dbouwman:
Could we do something like what rest-request does here --> https://github.com/Esri/arcgis-rest-js/blob/master/packages/arcgis-rest-request/src/utils/encode-form-data.ts#L22

`if (typeof Blob !== "undefined") { ...work with Blob ...}`

and for now, just throw in Node?

tomwayson:
@drewctate  I thought maybe ember-auto-import was bringing in the UMD builds in my test app, but I see this in the debug output:

```bash
[...]

PR #249, opened 2020-05-08
Solution.js Related Issues

dbouwman:
- change imports to not use the `index` files - resolves some circular deps
- drop use of `util` and instead internalize some type check fns

~TODO: This still needs a resolution around `FetchBlob` as per #248~

Opted to drop "isomorphic" support for `Blob` as that was causing all manner of grief in UMD and (apparently) ESM builds


packages/common/src/i18n/fetch-hub-translation.ts
```diff
@@ -1,5 +1,5 @@
 import { IPortal } from "@esri/arcgis-rest-portal";
-import { getHubLocaleAssetUrl } from "..";
+import { getHubLocaleAssetUrl } from "../urls/get-hub-locale-asset-url";
```
drewctate, line 2: Great. As an aside I think that `../urls/` would also fix the problem.

PR #252, opened 2020-05-11
removing es2017 from the tsconfig lib array

drewctate:


tomwayson:
I suspect the above example works b/c `item` is type `any`, and therefore for all TS knows `typeKeywords` might have an `inlcudes()`???

PR #268, opened 2020-05-27
page utils

drewctate:


packages/sites/src/pages/create-page-model-from-template.ts
```diff
@@ -0,0 +1,57 @@
...
+    deepSet(
+      pageModel,
+      "item.properties.parentInitiativeId",
+      parentInitiativeId
+    );
+  }
+  // put the slug into the hash so we can use it in following templates
+  deepSet(pageModel, "data.values.slug", slugify(pageModel.item.title));
+  // strip out any non-templated linked sites since they will not be accessible
+  const linkedSites = getProp(template, "data.values.sites");
```
drewctate, line 50: Seems like this logic does the opposite of what is described in the comment above (e.g. it strips out _templated_ sites and leaves _non-templated_ ones. Is this a mistake @dbouwman ?

Issue #254, opened 2020-05-11
remove exisrting use of async/await


tomwayson:
in favor of #250

tomwayson:
closing this since #488 - at least in the Hub app we can opt into the ES2017 build and use the untranspiled `async`/`await` code.

PR #271, opened 2020-06-01
[DNM] Initial prototype of Hub content methods

ajturner:
This is a suggested package for managing "Hub Content". Content are a composition of Portal Items plus additional metadata, relationships, and capabilities (e.g. downloads). 

The `hub-content` package provides standard _CRUD_ operations using a proposed `Hub Content Model` that has common elements for all content types and specific `attributes` per unique content type (e.g. Event times, dataset fields, etc.) 

The primary purpose of this package is to provide a simple, consistent, and extensible interface to Hub content to improve the developer experience for Hub sub-systems (UI, API, widgets, etc) and possibly performance by dynamic loading Hub packages as necessary - and not up-front in the client UI.

The CRUD methods are also a facade for the dual Hub API and Portal API. All content information is durably stored in Portal, but when there is a Hub API (online, public items). Then the content is also added directly to that index. While Hub does create an index through backend notifications and harvesting, there can be significant delay or multiple edits that may result in out-of-date information being available in Hub index. So when using the Hub UI and Hub.js this information can be more consistent. 

Content that is not yet in the Hub API (non-public, or created outside of Hub) then Hub API calls fall-back to direct Portal calls. 

## Content-specific metadata and methods

While this prototype provides a single common interface to all Content types - there may be significant additional methods for specific content types that suggests new packages. For example, `hub-datasets` may be useful for download, field operations, analyses, etc. These content-type packages would use `hub-content` for the information interface and then add specifics methods or metadata modifications.

## Other Hub.js packages

This prototype focused on Hub content. However, we should also prototype a complete Hub API wrapper for `hub-search` to call the Hub and Portal Search API, content downloads, as well as interfaces for `hub-users`, `hub-teams`, etc.

### DIFF
--- a/packages/common/src/api.ts
+++ b/packages/common/src/api.ts
@@ -1,6 +1,8 @@
 import { IRequestOptions } from "@esri/arcgis-rest-request";
 import { getPortalUrl } from "@esri/arcgis-rest-portal";
 import { _getHubUrlFromPortalHostname } from './urls/_get-hub-url-from-portal-hostname';
+import { IHubRequestOptions } from "./types";
+
 /**
  * ```js
  * import { getHubApiUrl() } from "@esri/hub-common";
@@ -17,4 +19,29 @@ export function getHubApiUrl(requestOptions: IRequestOptions): string {
   
 }
 
+export function getFromHubAPI(id:string, hubRequestOptions?: IHubRequestOptions): Promise<any> {
+  let hubUrl = "https://hub.arcgis.com/api";
+  let headers:any = {};
 
+  if (hubRequestOptions !== undefined) {
+    hubUrl = hubRequestOptions.hubApiUrl;
+    headers['Authorization'] =  hubRequestOptions.authentication.token;
+  }
+  
+  const url = `${hubUrl}/v3/datasets/${id}`;
+  const opts = {
+    method: "GET",
+    mode: "cors",
+    headers
+  } as RequestInit;
+  return fetch(url, opts)
+    .then(raw => raw.json())
+    .then(contentData => {
+      return contentData;
+    })
+    .catch(err => {
+      throw Error(
+        `getFromHubAPI: Error requesting from Hub API: ${err}`
+      );
+    }); 
+}
--- a/packages/common/src/api.ts
+++ b/packages/common/src/api.ts
@@ -34,8 +34,16 @@ export function getFromHubAPI(id:string, hubRequestOptions?: IHubRequestOptions)
     mode: "cors",
     headers
   } as RequestInit;
+
+  // TODO: handle Hub missing content by ID
   return fetch(url, opts)
-    .then(raw => raw.json())
+    .then(response => {
+      if (!response.ok) {
+        // TODO: handle passing up enumerated Hub API error codes
+        throw Error(response);
+      }
+      return response.json()
+    )
     .then(contentData => {
       return contentData;
     })
--- a/packages/content/README.md
+++ b/packages/content/README.md
@@ -1,0 +1,148 @@
+# @esri/hub-content
+
+> Manage Hub content [`@esri/hub.js`](https://github.com/Esri/hub.js).
+
+### Example
+
+```bash
+npm install @esri/hub-content
+```
+
+```js
+import { getContent  } from '@esri/hub-content';
+
+// pass in a site id
+getContent("abc123")
+    .then(
+        contentModel => {
+            contentModel.title; // access content information
+        }
+    );
+```
+
+## Design
+
+Hub Content is the generic term for Hub-related items and integrated metadata. Content types include datasets, maps, documents, apps, surveys, as well as events, sites, pages, and initiatives. Content provides a common metadata model and interface for creating, viewing, and managing content - as well as content-specific attributes and associations such as map basemap, app datasets, site pages, and initiative site. 
+
+Content have common metadata, associated content or resources (e.g. owner member), and actions (update, delete). Content metadata is composed of the Portal Item + Data with content- or org-specific enrichments. For example, a Map content item has 
+- _literal metadata_ like `title`
+- _inferred metadata_ such as `coverage` (global, regional, local) derived from the geographic extent, 
+- _linked metadata_ `license` created through `licenseInfo`
+- _associated resources_ such as datasets via `operationalLayers` or `owner` (`as IHubMember`) via the `item.owner`
+
+### Methods
+
+- `getContent(id, associatedOptions, hubRequestOptions)`
+- `createContent(id, attributes, hubRequestOptions)`
+- `updateContent(id, attributes, hubRequestOptions)`
+- `deleteContent(id, hubRequestOptions)`
+
+`associatedOptions` is proposed to support external references that may require additional XHR to fetch. Similar to JSON-API or GraphQL. 
+For example `associatedOptions: { itemInfo: true, itemData: true, connectedContent: 5 }`
+
+- Question: should download formats + links be optional associations or require XHR via the download components?
+
+### Metadata
+
+Based on the [Hub Content UI Inventory](https://app.lucidchart.com/invitations/accept/dab82a25-2ed9-4f63-8075-3db8873087e3)
+
+Attemping a simple _Hungarian Notation_ for attributes suffix: `*Date`, `*Id`, `*Number`. Plural (e.g. `tags`) is an array.
+
+```ts
+export interface IContentModel extends IModel {
+    id: string; 
+    type: 'document' | 'dataset' | 'map' | 'app' | 'site' | 'initiative' | 'event' | 'template' ;
+    owner: IHubMember;
+    
+    access: {
+        visibility: 'private' | 'org' | 'public';
+        permission?: 'view' | 'edit' | 'admin'; // based on User requesting the info
+        updateGroups?: Array<IGroup>;
+        viewGroups?: Array<IGroup>;
+    };
+
+    metrics?: {
+        visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
+        viewNumber?: number; // available depending on visibility; 
+    };
+
+    // Common metadata
+    title: string;
+    summary: string;
+    description: string;
+    license: IHubLicense; // title, description, and optional link to license item with more info
+    source: IHubOrg | IHubTeam | IHubMember; // each of these has common metadata like `title`, `thumbnail` , and `link`
+    createdDate: Date; 
+    publishedDate: Date;
+    updatedDate: Date;
+    geography: IHubGeography; // {title, coverage (global|regional|local), extent (WSEN), geometry, source}
+    categories?: Array<string>;
+    tags: Array<string>;
+    language: string; // written locale e.g. en, es, cn
+    contentUrl: string; // Link the the raw content
+    thumbnailUrl: string; // Full URL
+    status?: string; // e.g. [in-development, maintained, archived, unreviewed, prototype]
+
+    // Unique or additional formal metadata that will be displayed in sidebar
+    metadata?: {
+        // Additional custom/formal metadata elements
+    };
+
+    // Content specific attributes such as 
+    attributes?: {
+        // Dataset, e.g.
+        fields?: [{name:string, type:string, statistics:Array<any>}];
+        recordNumber?: number;
+        // Map, e.g.
+        layers?: [];
+        basemap?: string;
+        // Document
+        format?: 'link' | 'PDF' | 'MS Word' | 'MS Excel' | 'Text';
+        // Initiative
+        stage?: string;
+        followersNumber?: number;
+        coreTeamId?: string; 
+        contentTeamId?: string;
+        supportingTeamsNumber?: number;
+        // Event
+        sponsors?: Array<any>;
+        attendeesNumber?: number;
+        limitNumber?: number;
+        startDate?: Date;
+        endDate?: Date;
+        videoUrl?: string;
+    };
+
+    // Depending on the association requests, e.g.
+    associations?: {
+        relatedContent?: Array<IContentModel>;  // Automatically determined through similar metadata, usages
+        connectedContent?: Array<IContentModel>; // Explicit through usage, e.g. a dataset is connected to the Maps that use it.
+    }
+
+    // Optional configured app links that replace "Create StoryMap" with links to specific apps/sites
+    // 'home' is special as the landing page for the item instead of /content/:id/about or ../explore
+    links?: {
+        home: string;
+    };
+
+    // [Future] View configuration options such as cartography, charts, table, etc.
+    contentDisplay?: 'thumbnail' | 'map'; 
+    // [Future] Optional configured home at `/content/:id` such as a Page or App 
+    home?: {
+        type: IContentType;
+        contentId?: string; // if type is a Hub content type
+        contentUrl?: string; // if type is an external link that will be iframed?
+    }
+
+    // Legacy / optional Portal items. TODO: Should these be in `associations`
+    item?: IItem; // Portal Item
+    data?: { // Portal item.data
+        [propName: string]: any;
+    };
+}
+
+```
+
+### Raw Info
+
+Content also include the original portal info via `content.item` and `content.data` for backwards compatibility and staging of new content model attributes.
--- a/packages/content/src/create.ts
+++ b/packages/content/src/create.ts
@@ -1,0 +1,81 @@
+import { IContentModel } from "./index";
+import { IHubRequestOptions } from "@esri/hub-common";
+import { createItem, IItem } from "@esri/arcgis-rest-portal";
+
+/**
+ * ```js
+ * createContent(content)
+ *  .then(result => {
+ *    // handle result including new item id
+ *  })
+ * ```
+ * Create the content in Portal and also ping Hub API to update index.
+ *
+ *
+ * @param content - Content Model
+ * @param requestOptions - Tequest options that may have authentication manager
+ * @returns A Promise that will resolve with the entire Content object including the new ID 
+ * @export
+ */
+export function createContent(
+  content: IContentModel,
+  hubRequestOptions?: IHubRequestOptions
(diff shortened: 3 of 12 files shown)


packages/content/src/get.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @returns A Promise that will resolve with the Content metadata
+ * @export
+ */
+export function getContent(
+  id: string,
+  hubRequestOptions?: IHubRequestOptions
+): Promise<any> {
+
+  if(hubRequestOptions.isPortal) {
+    return _getContentFromPortal(id, hubRequestOptions);
```
dbouwman, line 35: awaits needed on these

packages/content/src/get.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @returns A Promise that will resolve with the item and data
+ * @export
+ */
+export function _getContentFromPortal(
+  id: string,
+  hubRequestOptions?: IHubRequestOptions
+): Promise<any> {
+
+  return getModel(id, hubRequestOptions).then((item:IModel) => {
+    return new Promise(resolve => {
```
dbouwman, line 62: so - either you can `const model = await getModel(...)` and then later just `return model` or you run iwth promises and start the chain w/ `return getModel(...).then(..next step...)....then(foo => { return foo;})`

packages/content/test/_get-content.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      })
+    );
+
+    const content = await getContent("some-id", {isPortal: true} as hubModule.IHubRequestOptions);
+
+    expect(content.description).toBe("Content from Portal");
+  });
+
+  it("get content data from Hub", async () => {
+    spyOn(hubModule, "getFromHubAPI").and.returnValue(
```
dbouwman, line 19: jasmine spies have --> `.and.resolveTo(whatever)`. And usually you want to make assertions on the spy, so 
```
const mySpy = spyOn(....);
...later...
expect(mySpy.calls.argsFor(1)[0]).toBe('foo', 'was called with foo');
```

packages/common/src/api.ts
```diff
@@ -17,4 +19,37 @@ export function getHubApiUrl(requestOptions: IRequestOptions): string {
   
 }
 
+export function getFromHubAPI(id:string, hubRequestOptions?: IHubRequestOptions): Promise<any> {
```
dbouwman, line 22: We'll have to figure out how to handle an optional HubRequestOptions - b/c that is how we know if we're dealing with portal or not. And in fact, determining that for an un-authenticated is a real problem, which we get around by having different builds of UI which have that flag "built-in" as part of the app's config. For the "Content" aspect (and likely other aspects of Hub.js) we need some solution, which may end up being Classes composed from lower-level functions... and the constructor requires either a full `UserSession` or a more minimal hash with the portal url and a `isPortal` flag... something like that

packages/common/src/api.ts
```diff
@@ -17,4 +19,37 @@ export function getHubApiUrl(requestOptions: IRequestOptions): string {
...
+  
+  const url = `${hubUrl}/v3/datasets/${id}`;
+  const opts = {
+    method: "GET",
+    mode: "cors",
+    headers
+  } as RequestInit;
 
+  // TODO: handle Hub missing content by ID
+  return fetch(url, opts)
```
dbouwman, line 39: We should consider having a `hubFetch` general purpose wrapper to centralize `fetch` similar to rest-js...

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+- _linked metadata_ `license` created through `licenseInfo`
+- _associated resources_ such as datasets via `operationalLayers` or `owner` (`as IHubMember`) via the `item.owner`
+
+### Methods
+
+- `getContent(id, associatedOptions, hubRequestOptions)`
+- `createContent(id, attributes, hubRequestOptions)`
+- `updateContent(id, attributes, hubRequestOptions)`
+- `deleteContent(id, hubRequestOptions)`
+
```
dbouwman, line 39: Given the set of things that fall into `content`, I'm pretty sure we will want to delegate to some type-specific handlers - particularly when talking about apps/surveys/events/sites/pages - we should keep this in mind, esp re: the `create` and `update` fns, and for some, the `delete` (i.e. deleting a site should kill the teams, and domain record)

ajturner: @dbouwman yes. As I noted in the description I think it makes good sense to have type-specific methods. 

One aspect of this I'm curious about is if there is another content-facade. So something like...

```ts
let site = getContent(siteId);

deleteContent(site);

function deleteContent(content: IContentModel) {
  switch(content.type) {
    case 'dataset': 
      deleteDatasetById(content.id);
      break;
    case 'site':
      deleteSiteById(content.id);
      break;
    default:
      deleteContentById(content.id);
  }
}
```

dbouwman: @ajturner yeah - a facade is a good pattern. Something we need to look into is how to do this so that we're not tightly coupling everything - hopefully dynamic imports can solve aspects of that for us

mikeringrose: maybe something of the form:
```
const content = ContentFactory.get(type).create(params)
```
there's a couple benefits including making things easier for testing purposes.

packages/common/src/api.ts
dbouwman: That's doable, as long as we compose the `API` out of the free functions. Another wrinkle  is that the UI application uses two identities in a number of places, so we want to make it easy/cheap to spin up an instance (i.e. we don't fire off N xhrs in the factory)

packages/content/README.md
dbouwman: A challenge with this is figuring out if/how we do any checks on the `params`. Depending on what you are creating, these things can be massive - i.e. this is exactly what Solution.js solves by using a template with parameters. However that only gets you so far... the other alternative is that the API is dumb, takes sends the params to the Portal API and passes back the response

mikeringrose: it's easy when you are talking about a handful of types with a handful of parameters, but I think in this case it's easier and better to be dumber.

packages/content/src/create.ts
```diff
@@ -0,0 +1,81 @@
...
+  content: IContentModel,
+  hubRequestOptions?: IHubRequestOptions
+): Promise<any> {
+
+  // First we create the item in Portal, then also add to the Hub API index
+  // in the future the Hub API can use Hub.js to do the Portal proxy
+  const updatedContent = await _createContentInPortal(content);
+  // Add API and methods to directly update Hub
+  // hubModule.createContent(updatedContent);
+  return new Promise((resolve) => resolve(content));
```
mikeringrose, line 79: you can just do `return _createContentInPortal(content);`

packages/content/README.md
dbouwman: 👍 definitely easier

packages/content/src/get.ts
```diff
@@ -0,0 +1,164 @@
...
+ * This method works to call either the Portal (AGO) API or Hub API
+ * and return a common data model.
+ *
+ *
+ * @param id - Content Id
+ * @param requestOptions - Tequest options that may have authentication manager
+ * @returns A Promise that will resolve with the Content metadata
+ * @export
+ */
+export function getContent(
```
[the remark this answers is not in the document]
ajturner: @dbouwman I didn't investigate the rest of the library - so I wasn't sure if `async` was common across all methods. I agree that it's preferable.

packages/content/src/get.ts
ajturner: I'd prefer the `await` pattern as well.

packages/content/src/get.ts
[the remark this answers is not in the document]
tomwayson: Our current stance on async/await is that it should be removed (#254) b/c currently in the Hub app the `tslib` generator code is included for _each_ Hub.js package that uses async/await. This is likely more of a Hub app build problem (I suspect `ember-auto-import`), with broader implications (tslib's `Object.assign` polyfill is included about a dozen times even though we don't need a single one). It's also to some extent a JS ecosystem problem (no one compiles their dependencies). Until we fix that problem for at least the Hub app, I say no async/await in Hub.js.

packages/common/src/api.ts
ajturner: First, determining the environment would seem to remain as a deployment configuration / build config. When a deploy is finished it can copy/write the config.json file with the environment parameters that is fetched at app bootup. 

@mikeringrose within that, I think your proposal is then to wrap that state management/access into an `API` module. Is that accurate?

There might be various cases where there are different environment parameters such as calling to another Org (community org) or even to an Enterprise server. These would be handled within that specific workflow passing in unique options. For the community use-case the "Portal URL" and user credentials are fetched at operation time such as "search for all (private) community members"

packages/common/src/api.ts
ajturner: yes. This was _quicky + dirty_. Should we just use rest-js version of fetch?

packages/content/README.md
ajturner: can we use JSON Schema similar to what AGO is using internally for [Webmap and Webscene specs](https://devtopia.esri.com/jere5132/schema-tooling) (and [validator](https://devtopia.esri.com/juan9976/schema-validator))?

But do this at runtime within our API methods for create/update?

packages/common/src/api.ts
```diff
@@ -17,4 +19,37 @@ export function getHubApiUrl(requestOptions: IRequestOptions): string {
...
+      return response.json()
+    )
+    .then(contentData => {
+      return contentData;
+    })
+    .catch(err => {
+      throw Error(
+        `getFromHubAPI: Error requesting from Hub API: ${err}`
+      );
+    });
```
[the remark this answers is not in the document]
ajturner: when you say _client_ do you mean the `getContent` method itself?

packages/common/src/api.ts
```diff
@@ -17,4 +19,37 @@ export function getHubApiUrl(requestOptions: IRequestOptions): string {
...
+    .then(response => {
+      if (!response.ok) {
+        // TODO: handle passing up enumerated Hub API error codes
+        throw Error(response);
+      }
+      return response.json()
+    )
+    .then(contentData => {
+      return contentData;
+    })
```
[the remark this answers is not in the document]
ajturner: yep - quick code

packages/content/src/create.ts
ajturner: This won't follow my intent (in the comments). More pseudo-codey to be explicit

when Creating Content:
  first Create Item in ArcGIS Portal (Online/Enteprise)
  then if successful also Create Content in Hub (so that it's immediately indexed)
  

So I don't want to just return the Portal call, but do the additional XHR

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+- Question: should download formats + links be optional associations or require XHR via the download components?
+
+### Metadata
+
+Based on the [Hub Content UI Inventory](https://app.lucidchart.com/invitations/accept/dab82a25-2ed9-4f63-8075-3db8873087e3)
+
+Attemping a simple _Hungarian Notation_ for attributes suffix: `*Date`, `*Id`, `*Number`. Plural (e.g. `tags`) is an array.
+
+```ts
+export interface IContentModel extends IModel {
```
tomwayson, line 52: I don't think this should extend `IModel` - I don't think we should include the redundant `item`

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+    id: string; 
+    type: 'document' | 'dataset' | 'map' | 'app' | 'site' | 'initiative' | 'event' | 'template' ;
```
tomwayson, line 54: this should be `hubType` or something to keep from confusion w/ `item.type`

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+    owner: IHubMember;
+    
+    access: {
```
tomwayson, line 57: Just like my previous comments about `owner` being an `IHubMember`, I'm a little worried about nested properties like `access` and `metrics` introducing complexity that at the very least, not work w/ Ember's out of the box change detection.

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+        updateGroups?: Array<IGroup>;
+        viewGroups?: Array<IGroup>;
+    };
+
+    metrics?: {
+        visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
+        viewNumber?: number; // available depending on visibility; 
+    };
+
+    // Common metadata
```
tomwayson, line 69: It would be good to add comments for each of these to indicate where they come from - I keep bouncing from here to the fns that populate this interface. Ex `title: string; // item.title or dataset.name`

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+    thumbnailUrl: string; // Full URL
+    status?: string; // e.g. [in-development, maintained, archived, unreviewed, prototype]
+
+    // Unique or additional formal metadata that will be displayed in sidebar
+    metadata?: {
+        // Additional custom/formal metadata elements
+    };
+
+    // Content specific attributes such as 
+    attributes?: {
```
tomwayson, line 92: not sure about using `attributes` for this mainly out of concern for potential confusion w/ `dataset.attributes`. So this will be `dataset.attributes` stripped of the above properties?

tomwayson: it would be good to clarify w/ comments below which of these exist already

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+        videoUrl?: string;
+    };
+
+    // Depending on the association requests, e.g.
+    associations?: {
+        relatedContent?: Array<IContentModel>;  // Automatically determined through similar metadata, usages
+        connectedContent?: Array<IContentModel>; // Explicit through usage, e.g. a dataset is connected to the Maps that use it.
+    }
+
+    // Optional configured app links that replace "Create StoryMap" with links to specific apps/sites
```
tomwayson, line 122: where are these stored?

packages/content/README.md
```diff
@@ -0,0 +1,148 @@
...
+    // [Future] Optional configured home at `/content/:id` such as a Page or App 
+    home?: {
+        type: IContentType;
+        contentId?: string; // if type is a Hub content type
+        contentUrl?: string; // if type is an external link that will be iframed?
+    }
+
+    // Legacy / optional Portal items. TODO: Should these be in `associations`
+    item?: IItem; // Portal Item
+    data?: { // Portal item.data
```
tomwayson, line 139: we should keep this, even if we don't keep `item`

packages/content/README.md
tomwayson, line 52: Maybe `IContent`? I know it's not part of this PR, but `IModel` is not a great name for `item` & `data`.

packages/content/src/get.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @returns A Promise that will resolve with the item and data
+ * @export
+ */
+export function _getContentFromHub(
+  id: string,
+  hubRequestOptions?: IHubRequestOptions
+): Promise<any> {
+
+  return getFromHubAPI(id, hubRequestOptions).then((hubResponse: any) => {
+    return new Promise((resolve) => {resolve(hubResponse)})
```
tomwayson, line 117: TODO: transform the response into the `IContentModel` format.

packages/common/src/api.ts
tomwayson: @mikeringrose's suggestion is intriguing and worth investigating. Assuming the `hubAPI` instance would expose all fns w/ _overridable_ defaults, it's probably a less hacky version of what [we did in rest-js](https://esri.github.io/arcgis-rest-js/api/request/setDefaultRequestOptions/), which we don't yet take advantage of in the Hub app. Seems like it could be a lot of work/maintenance to attach all the fns to the instance tho. I'm curious to see what it looks like in practice.

packages/common/src/api.ts
tomwayson: If you mean `requestOptions.fetch`, then yes, we should probably favor that.

packages/common/src/api.ts
tomwayson: Thinking about this some more, @mikeringrose's suggestion (as I understand it) would likely prevent us from having tree shaking or lazy-loading certain portions of the app.

Full disclosure, it's very possible that the hack I use for rest-js' `setDefaultRequestOptions()` technically violates [webpack's `sideEffects: false` convention](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free), but we've tested and it seems to work, so 🤷‍♂️

tomwayson:
generally speaking, I've always wondered why the dataset records returned from the Hub API and/or the models we create from them in the Hub aren't a superset of the portal item schema. Is there a reason we wouldn't do it that way?

mjuniper:
I have not yet had a chance to look carefully at this or follow the conversation but this is the stuff I know I need for documents:

```
"id": this can be id or preferably slug
"title": currently mapped to item.title
"description": brolly says we actually want snippet if available so maybe we include both?
"extent": this is for drawing the extent on a map, so, more generically, this would maybe be geometry
"type": item.type - this is like Microsoft Excel, PDF, etc but it would also be great to have a more generic type like "document"
"metadata": for documents this is publisher, modified, created, license
"status": for documents this is `item.orgId !== site.orgId` ? 'Community' : ''
"tags": this one is weird - i'm calling it tags but it is not the item tags, it is `access`, `culture`
"thumbnailUrl": fully qualified
"size": for file size
```

tomwayson:
@mjuniper 

That's your view model though, which is a bit different (translated, formatted, etc). Also, it's missing `url` (from `item.url`), which for the view model would be `iframeUrl` (all the horrible stuff our app does to an unsuspecting URL that's bound for an iframe).

mjuniper:
Yes but that gives a good idea what I need in order to construct my viewmodel.

ajturner:
Good conversation - but not a mergable PR. Closing. Available in history for research and planning.

PR #274, opened 2020-06-05
[DNM] Added content model prototype

ajturner:
Here is a simplified content model recommendation. 

Generally, it's 
- stronger typing on `IItem` (e.g. `content.createdDate: Date = new Date(item.created)`)
- objects for cached relationships (e.g. `content.maintainer = getUser(item.owner)`) 
- new configuration parameters (e.g. `content.metrics.visibility`)
- additional metadata (`content.metadata`) 
- content-type specific values like event times, doc format, etc. (`content.attributes`)

Please review for clarity, not unnecessarily complex, and extensible. We need more than items, but less (and more organized than) v3/datasets.

### DIFF
--- a/packages/content/src/types.ts
+++ b/packages/content/src/types.ts
@@ -1,0 +1,137 @@
+import { IItem, IGroup } from "@esri/arcgis-rest-types";
+
+export type HubType = 'document' | 'dataset' | 'map' | 'app' | 'site' | 'initiative' | 'event' | 'template';
+
+// Based on the [Hub Content UI Inventory](https://app.lucidchart.com/invitations/accept/dab82a25-2ed9-4f63-8075-3db8873087e3)
+// Attempting a simple _Hungarian Notation_ for attributes suffix: `*Date`, `*Id`, `*Number`. Plural (e.g. `tags`) is an array.
+
+export interface IContent extends IItem {
+  
+  // Inherited metadata (noting relevant subset of all item attributes)
+  // id: string; // item.id
+  // title: string; // item.title
+  // snippet: string; // item.snippet
+  // description: string; // item.description
+  // categories?: Array<string>; // item.categories
+  // tags: Array<string>; // item.tags  
+  // culture: string; // written locale e.g. en, es, cn
+  // url?: string;
+  // size: number;
+
+  // Derived metadata
+  hubType: HubType;
+  maintainer: IHubMember; // item.owner with more user metadata
+  access: { // overrides item.access with more attributes. could flatten.
+    visibility: 'private' | 'org' | 'public'; // item.access
+    permission?: 'view' | 'edit' | 'admin'; // item.itemControl
+    groups?: Array<IGroup>; // item.sharing.groups via content/users/:username/items/:id
+  };
+
+  // Explicit data information since this is a common confusion + bug report
+  createdDate: Date; // formal metadata || new Date(item.created)
+  createdDateSource: string; // description of what was used for this attribute
+  publishedDate: Date; // formal metadata || new Date(item.created)
+  publishedDateSource: string; // description of what was used for this attribute
+  updatedDate: Date; // formal metadata || new Date(item.modified)
+  updatedDateSource: string; // description of what was used for this attribute
+  
+  contentUrl: string; // Link to the raw content. item.url in most (but not all) item types
+  thumbnailUrl: string; // Full URL. item.thumbnail with host + path 
+  
+  license: IHubLicense; // [Future] item.licenseInfo 
+  boundary: IHubGeography; // [Future] Inline, or boolean when stored at known location /resources/boundary.json
+  
+  // Additional metadata from custom/formal elements
+  metadata?: {
+      // Unique or additional formal metadata that will be displayed in sidebar
+  };
+
+  // Hub configuration metadata
+  actionLinks?: IActionLink[]; // item.properties.links
+  actionConfigs?: Object; // item.properties.actions - enable/disable standard actions like `createWebmap` or `createStorymap`
+
+  metrics?: { // Set visibility for telemetry metrics. Nested object future-proofing, but could flatten.
+      visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
+  };
+
+  contentDisplay?: 'thumbnail' | 'map'; // [Future] View configuration options such as cartography, charts, table, etc.
+  source: IHubSocial; // [Future] each of these has common metadata like `title`, `thumbnail` , and `link`
+  
+  // Content specific values. Combination of relevant item.data, layer info, enrichments, configuration settings
+  // could use values instead - which is common within item.data.values
+  attributes?: {
+      // Dataset, e.g.
+      fields?: [{name:string, type:string, statistics:Array<any>}];
+      recordNumber?: number;
+      // Map, e.g.
+      layers?: [];
+      basemap?: string;
+      // Document
+      format?: 'link' | 'PDF' | 'MS Word' | 'MS Excel' | 'Text';
+      // Initiative
+      stage?: string;
+      followersNumber?: number;
+      coreTeamId?: string; 
+      contentTeamId?: string;
+      supportingTeamsNumber?: number;
+      // Event
+      sponsors?: Array<any>;
+      attendeesNumber?: number;
+      limitNumber?: number;
+      startDate?: Date;
+      endDate?: Date;
+      videoUrl?: string;
+  };
+}
+
+// Common interface to people networks: Teams, Orgs, Members
+export interface IHubSocial {
+  id: string;
+  name: string;
+  thumbnailUrl?: string;
+  region: string; // user.region - US/Europe/etc.
+  culture: string; // e.g. user.culture
+  createdDate: Date;
+  updatedDate: Date;
+  url: string;
+}
+export interface IHubOrg extends IHubSocial { }
+
+export interface IHubTeam extends IHubSocial  { }
+
+// Simple user info - more could be added/cached
+export interface IHubMember extends IHubSocial {
+  username: string; // user.username
+  thumbnailUrl?: string; // user.thumbnail
+}
+
+// title, description, and optional link to license item with more info
+export interface IHubLicense {
+  id: string; // Licenses may be Items, so provide reference for links
+  title: string;
+  description: string;
+  link: string;
+  thumbnailUrl: string;
+  capabilities: {
+    reuse: boolean;
+    commercial: boolean;
+    derivatives: boolean;
+    attribution: boolean;
+    osm: boolean;
+  }
+}
+
+export interface IHubGeography {
+  title:string;
+  coverage?: 'global' | 'regional' | 'local'; // enrichment
+  geometry?: string; // serialized JSON, or should this refer to /resources/boundary.json
+  source: string; // item ID used for setting geometry
+}
+
+// Optional configured app links that replace "Create StoryMap" with links to specific apps/sites
+// per https://esriarlington.tpondemand.com/entity/96316-content-viewer-sees-associated-app-links
+// 'home' would be special as the landing page for the item instead of /content/:id/about or ../explore
+export interface IActionLink {
+  title: string;
+  url: string;
+}


packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
+import { IItem, IGroup } from "@esri/arcgis-rest-types";
+
+export type HubType = 'document' | 'dataset' | 'map' | 'app' | 'site' | 'initiative' | 'event' | 'template';
+
+// Based on the [Hub Content UI Inventory](https://app.lucidchart.com/invitations/accept/dab82a25-2ed9-4f63-8075-3db8873087e3)
+// Attempting a simple _Hungarian Notation_ for attributes suffix: `*Date`, `*Id`, `*Number`. Plural (e.g. `tags`) is an array.
+
+export interface IContent extends IItem {
```
mikeringrose, line 8: what about events? they are not items.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  
+  // Inherited metadata (noting relevant subset of all item attributes)
+  // id: string; // item.id
```
mikeringrose, line 11: not sure if this is confusing or not, but a dataset's id is different from its item id. there's a lot of place in the code base where we split the dataset id to get the item id.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  contentUrl: string; // Link to the raw content. item.url in most (but not all) item types
+  thumbnailUrl: string; // Full URL. item.thumbnail with host + path 
+  
+  license: IHubLicense; // [Future] item.licenseInfo 
+  boundary: IHubGeography; // [Future] Inline, or boolean when stored at known location /resources/boundary.json
+  
+  // Additional metadata from custom/formal elements
+  metadata?: {
+      // Unique or additional formal metadata that will be displayed in sidebar
+  };
```
mikeringrose, line 47: should this be defined as `metadata?: {[field: string]: any}`?

packages/content/src/types.ts
ajturner: Technically, they are not items, but for purposes of this interface I'm proposing they look + act like items. 

I'm also OK if we say "no, Events, Members, Teams" are different and should have their own interfaces. but I think we want them to have common structure and attributes.

packages/content/src/types.ts
ajturner: Good point - we should leave `content.id` to mean `item.id` and a few parameters

- `content.hubId` our own id, e.g. `content.hubId = '${item.id}_${item.values.layerIndex}'`
- `content.slug` e.g. `urban-forestry-trees`

packages/content/src/types.ts
ajturner: The proposal is that it's nested and based on [mdJSON example](https://github.com/adiwg/mdJson-schemas/blob/master/examples/resourceInfo.json)

mikeringrose: So is it actually an array?

ajturner: No - that link is an array because it's a set of examples. 

A full metadata is an object like [this example](https://gist.github.com/ajturner/d68270883a1fc729e6319d7fc5e5a4f2)

Looking at that makes a point that the full metadata.json would be a sidecar resource. The purpose of this `content.metadata` was to have a configured summary of "Maintainer wants this metadata to appear in result cards / about views". For example, some orgs might want their department phone number and hours to show up. Another might want the methodology or sensor type. 

We can pull this attribute for now from the baseline, with expectation that it would be added in the relatively near future with specifics on what it stores and can we be explicit about its schema.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  // Hub configuration metadata
+  actionLinks?: IActionLink[]; // item.properties.links
+  actionConfigs?: Object; // item.properties.actions - enable/disable standard actions like `createWebmap` or `createStorymap`
```
rweber-esri, line 51: `Object` will accept primitive & non-primitive types. It may be more appropriate to to use `object` so the compiler throws errors if we try to assign a `string`, `number`, etc

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  // culture: string; // written locale e.g. en, es, cn
+  // url?: string;
+  // size: number;
+
+  // Derived metadata
+  hubType: HubType;
+  maintainer: IHubMember; // item.owner with more user metadata
+  access: { // overrides item.access with more attributes. could flatten.
+    visibility: 'private' | 'org' | 'public'; // item.access
+    permission?: 'view' | 'edit' | 'admin'; // item.itemControl
```
rweber-esri, line 26: Is there a case where we could obtain a `IContent`  result and not have at least `view` permission?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    groups?: Array<IGroup>; // item.sharing.groups via content/users/:username/items/:id
```
rweber-esri, line 27: Additionally, should `groups` be optional? Is the thought that they might not apply to all `HubType`s?

dbouwman: Items can exist that are not shared to a group - so if we want/expect this structure to be used for UX's where we are choosing an item to be added to a Site's catalog... then `groups` could be null - that said, it would be *much* better to have it be an empty array vs being null.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  // Explicit data information since this is a common confusion + bug report
+  createdDate: Date; // formal metadata || new Date(item.created)
+  createdDateSource: string; // description of what was used for this attribute
+  publishedDate: Date; // formal metadata || new Date(item.created)
+  publishedDateSource: string; // description of what was used for this attribute
+  updatedDate: Date; // formal metadata || new Date(item.modified)
+  updatedDateSource: string; // description of what was used for this attribute
+  
+  contentUrl: string; // Link to the raw content. item.url in most (but not all) item types
+  thumbnailUrl: string; // Full URL. item.thumbnail with host + path
```
dbouwman, line 39: `thumbnailUrl` is a good example of a derived, read-only property. There is no way to update this property and change the thumbnail. Instead you have to POST w/ a Blob payload in the `thumbnail` property, and then, the backend api will return a new name in the thumbnail property, and we can re-compute the `thumbnailUrl` property.

packages/content/src/types.ts
dbouwman: I think Mike is saying that from a typescript perspective, this should be `metadata?: {[field: string]: any}` which means "accept an arbitrary hash"

packages/content/src/types.ts
dbouwman: @ajturner is the idea here to attach in functions? or is this more a hash of settings for well-known-actions?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  // Hub configuration metadata
+  actionLinks?: IActionLink[]; // item.properties.links
+  actionConfigs?: Object; // item.properties.actions - enable/disable standard actions like `createWebmap` or `createStorymap`
+
+  metrics?: { // Set visibility for telemetry metrics. Nested object future-proofing, but could flatten.
+      visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
+  };
+
+  contentDisplay?: 'thumbnail' | 'map'; // [Future] View configuration options such as cartography, charts, table, etc.
```
dbouwman, line 57: if this is going to get extended, then let's start w/ a hash vs an enum

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  // title: string; // item.title
+  // snippet: string; // item.snippet
+  // description: string; // item.description
+  // categories?: Array<string>; // item.categories
+  // tags: Array<string>; // item.tags  
+  // culture: string; // written locale e.g. en, es, cn
+  // url?: string;
+  // size: number;
+
+  // Derived metadata
```
dbouwman, line 21: If we think about the `update` flow for an `IContent`, and everything below this point are "derived" properties... maybe we should mark all of them as `readonly`? 
https://www.typescriptlang.org/docs/handbook/interfaces.html#readonly-properties

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+      visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
+  };
+
+  contentDisplay?: 'thumbnail' | 'map'; // [Future] View configuration options such as cartography, charts, table, etc.
+  source: IHubSocial; // [Future] each of these has common metadata like `title`, `thumbnail` , and `link`
+  
+  // Content specific values. Combination of relevant item.data, layer info, enrichments, configuration settings
+  // could use values instead - which is common within item.data.values
+  attributes?: {
+      // Dataset, e.g.
```
dbouwman, line 63: I'm not a huge fan of inherited classes or interfaces, but... for sanity and the usefulness of types, maybe `IContent` gets extended so that some of the `attributes` hash entries are required for specific types? maybe just for the HubTypes? Otherwise, the consuming code is going to have to do a lot of `if(getProp(model, 'attributes.fields')) {...}`

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    derivatives: boolean;
+    attribution: boolean;
+    osm: boolean;
+  }
+}
+
+export interface IHubGeography {
+  title:string;
+  coverage?: 'global' | 'regional' | 'local'; // enrichment
+  geometry?: string; // serialized JSON, or should this refer to /resources/boundary.json
```
dbouwman, line 127: To be useful, it should be a geometry - question is: geojson? or jsapi geometry?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  source: string; // item ID used for setting geometry
+}
+
+// Optional configured app links that replace "Create StoryMap" with links to specific apps/sites
+// per https://esriarlington.tpondemand.com/entity/96316-content-viewer-sees-associated-app-links
+// 'home' would be special as the landing page for the item instead of /content/:id/about or ../explore
+export interface IActionLink {
+  title: string;
```
dbouwman, line 135: i18n?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+
+// Simple user info - more could be added/cached
+export interface IHubMember extends IHubSocial {
+  username: string; // user.username
+  thumbnailUrl?: string; // user.thumbnail
+}
+
+// title, description, and optional link to license item with more info
+export interface IHubLicense {
+  id: string; // Licenses may be Items, so provide reference for links
```
dbouwman, line 110: How would we edit this? The item can only store a string in it's license field, and that's what's rendered in the Home App, so to make this work, it's really just the option to the license text, or a well-known-string (i.e. CC-BY etc) that cross-walks to preset versions of this hash. We'd have to have a UI that would present a license text editor or a well-known-license picker... we can't actually let them set these properties because where would we store it? and how would it manifest in the Home App UI?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  metadata?: {
+      // Unique or additional formal metadata that will be displayed in sidebar
+  };
+
+  // Hub configuration metadata
+  actionLinks?: IActionLink[]; // item.properties.links
+  actionConfigs?: Object; // item.properties.actions - enable/disable standard actions like `createWebmap` or `createStorymap`
+
+  metrics?: { // Set visibility for telemetry metrics. Nested object future-proofing, but could flatten.
+      visibility: 'private' | 'org' | 'updateGroups' | 'public'  ; // TODO: should we share with specific groups?
```
dbouwman, line 54: Since it's not an actual AGO feature, where is this stored? Item.properties seems like a good place, with some default value assigned if it's not present

packages/content/src/types.ts
dbouwman: also - how is this determined? from the type? or is this configurable... and if so where is it persisted?

packages/content/src/types.ts
mikeringrose: I think the problem when you have a function that works on Item and then somebody sends in an event, that could be misleading and confusing.

packages/content/src/types.ts
tomwayson: seems like we should also add `content.layerId` (or `content.layerIndex`)?

packages/content/src/types.ts
tomwayson: interesting idea. Along those same lines, I've been tempted to think about using an [abstract class](https://www.typescriptlang.org/docs/handbook/classes.html#abstract-classes) so we could actually implement things like `createdDate` etc as getters. I know that's a whole other can of worms b/c we get into the instance vs POJO debate.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  // description: string; // item.description
+  // categories?: Array<string>; // item.categories
+  // tags: Array<string>; // item.tags  
+  // culture: string; // written locale e.g. en, es, cn
+  // url?: string;
+  // size: number;
+
+  // Derived metadata
+  hubType: HubType;
+  maintainer: IHubMember; // item.owner with more user metadata
```
tomwayson, line 23: `ownerUser`?

packages/content/src/types.ts
tomwayson: good point.

I think a bigger question is how will `permission` be set? Do we expect the Hub API to return this? If not, then I think we should remove it and have this be something that's part of the view model at the application level.

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  access: { // overrides item.access with more attributes. could flatten.
```
tomwayson, line 24: I don't think derived properties should have the same name as native properties. In this case I think we should just return the item's `access` (see `permission` comment below)

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    visibility: 'private' | 'org' | 'public'; // item.access
+    permission?: 'view' | 'edit' | 'admin'; // item.itemControl
+    groups?: Array<IGroup>; // item.sharing.groups via content/users/:username/items/:id
+  };
+
+  // Explicit data information since this is a common confusion + bug report
+  createdDate: Date; // formal metadata || new Date(item.created)
+  createdDateSource: string; // description of what was used for this attribute
+  publishedDate: Date; // formal metadata || new Date(item.created)
```
tomwayson, line 33: what's the difference btw this and `createdDate`?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  maintainer: IHubMember; // item.owner with more user metadata
+  access: { // overrides item.access with more attributes. could flatten.
+    visibility: 'private' | 'org' | 'public'; // item.access
+    permission?: 'view' | 'edit' | 'admin'; // item.itemControl
+    groups?: Array<IGroup>; // item.sharing.groups via content/users/:username/items/:id
+  };
+
+  // Explicit data information since this is a common confusion + bug report
+  createdDate: Date; // formal metadata || new Date(item.created)
+  createdDateSource: string; // description of what was used for this attribute
```
tomwayson, line 32: examples?

packages/content/src/types.ts
tomwayson: again, I don't think we use the same keys as the item, in this case I propose something like `thumbnailFullUrl`

packages/content/src/types.ts
tomwayson: > We can pull this attribute for now from the baseline, with expectation that it would be added in the relatively near future with specifics on what it stores and can we be explicit about its schema.

sounds good to me

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+export interface IHubGeography {
+  title:string;
+  coverage?: 'global' | 'regional' | 'local'; // enrichment
+  geometry?: string; // serialized JSON, or should this refer to /resources/boundary.json
+  source: string; // item ID used for setting geometry
+}
+
+// Optional configured app links that replace "Create StoryMap" with links to specific apps/sites
+// per https://esriarlington.tpondemand.com/entity/96316-content-viewer-sees-associated-app-links
+// 'home' would be special as the landing page for the item instead of /content/:id/about or ../explore
```
tomwayson, line 133: You mean someone might supply { title: 'home', url: 'https://some.example.org/path/to/page' }` and the UI would know that b/c the title is `'home'` that's what it would use for the result's link in search results and such?

packages/content/src/types.ts
tomwayson: no

packages/content/src/types.ts
tomwayson: Yeah, is there a reason this would be serialized to a string?

Either way, it will be [ArcGIS Geometry](https://developers.arcgis.com/documentation/common-data-types/geometry-objects.htm)

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+    reuse: boolean;
+    commercial: boolean;
+    derivatives: boolean;
+    attribution: boolean;
+    osm: boolean;
+  }
+}
+
+export interface IHubGeography {
+  title:string;
```
tomwayson, line 125: seems like this should be optional

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  coverage?: 'global' | 'regional' | 'local'; // enrichment
+  geometry?: string; // serialized JSON, or should this refer to /resources/boundary.json
+  source: string; // item ID used for setting geometry
```
tomwayson, line 128: seems like this should be optional.

Also, what if the source was a feature? Maybe this should be a URL a la annotations `target`?

packages/content/src/types.ts
tomwayson: i hope it's the latter. If so, I propose `actionOptions`?

packages/content/src/types.ts
tomwayson: I'm not sure this belongs down at the content level unless we're talking about the option to persist a user configured override of the default behavior of the Hub application (i.e. if this is derived from `item.properties`

packages/content/src/types.ts
tomwayson: I think we should use inheritance for the "sub-types", but to clarify, we're still going to end up w/ a lot of `if(content.fields) {...}`, but what I believe @dbouwman  is saying is that at least we can avoid the `getProp()`

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  contentDisplay?: 'thumbnail' | 'map'; // [Future] View configuration options such as cartography, charts, table, etc.
+  source: IHubSocial; // [Future] each of these has common metadata like `title`, `thumbnail` , and `link`
+  
+  // Content specific values. Combination of relevant item.data, layer info, enrichments, configuration settings
+  // could use values instead - which is common within item.data.values
+  attributes?: {
+      // Dataset, e.g.
+      fields?: [{name:string, type:string, statistics:Array<any>}];
```
tomwayson, line 64: `[IField]`

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+      recordNumber?: number;
```
tomwayson, line 65: `recordCount`?

packages/content/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+      attendeesNumber?: number;
+      limitNumber?: number;
+      startDate?: Date;
+      endDate?: Date;
+      videoUrl?: string;
+  };
+}
+
+// Common interface to people networks: Teams, Orgs, Members
+export interface IHubSocial {
```
tomwayson, line 88: `IHubCommunity`? `IHubMember`?

packages/content/src/types.ts
ajturner: @tomwayson @mikeringrose _Events are a feature service_ is a logical implementation detail. Can you provide a concrete example of where the interface is fundamentally changed by this platform implementation? 

Events in particular may/should become items. but even before that the same domain model concepts apply (sharing, owner, etc.)

_arguing with myself_ if we conceptually say "Events are more like a Team than a Document" because Events have _members_ and _can attend_ actions then we could put events into the conceptual class of "Social Resources". Would that make more sense to an external developer and also our implementation?

packages/content/src/types.ts
ajturner: @tomwayson the `layerId` is very content type specific. It's nonsensical for documents/apps and confusing for maps!. 

Related _specifically_ to Feature Service layers there is a change to Online in 8.2 (July) that expands services into layers. We should [review the docs](https://confluencewikidev.esri.com/display/AGO/Sub-Layer+Pages) to understand how it might work in our API and also indexers.

tomwayson: OK, for `IFeatureLayerContent extents IContent`, seems like we should also add `featureLayerContent.layerId`

packages/content/src/types.ts
ajturner: `actionConfigs` is a hash of settings for links. 

for example
```js
actionConfigs: [
  { title: "Open GeoHive Viewer", url: "https://hosting.geohive.ie/custom-viewer?layer=4ef"},
]
```

packages/content/src/types.ts
ajturner: as noted in the comments, the value of `permission` comes _directly_ from the AGO Portal API.  [see Item Doc and search for 'itemControl'](https://developers.arcgis.com/rest/users-groups-and-items/item.htm). 

The common/default value will be `view`. The only reason view would not be available is if the item is blocked. e.g. `getContent('Art Vandelay')` . but that's not relevant as the response to that should be a better error IMO.

packages/content/src/types.ts
ajturner: I'm ok with `groups: Array<IGroup> = []`

packages/content/src/types.ts
ajturner: @dbouwman what if the implementation for WRITE of `thumbnailUrl` fetched the image and pushed it via the API? Alternatively, the update would return back either quietly that it didn't use the value or an explicit "worked but..."

This is the kind of implementation detail we want to isolate from end developers. If we think that someone should be able to set the thumbnail by URL then we can add that within the update method.

@tomwayson the original attribute is `item.thumbnail`. I explicitly did not use the same key by appending the "type" via `content.thumbnailUrl`. Since we extend `IItem` there still would be `content.thumbnail`.

packages/content/src/types.ts
ajturner: @tomwayson This attribute is user configurable. IT's optional which means we can default to whatever we think is best - but should be configurable.

Generally, I am proposing that we include (and possibly flatten) `item.properties` . As noted there are a few other "item maintainer" (owner or via update groups) configuration options.

packages/content/src/types.ts
ajturner: @dbouwman Generally "derived" can be bi-directional. So we could set `content.metadata.contactInfo.address` and have that go into the `content/items/:id/metadata` endpoint.

packages/content/src/types.ts
ajturner: @dbouwman I agree that it likely does make sense to have subclasses, such as `IDataset` and `IDocument`. The search results can return the correct type, but the subclasses can encapsulate various type-specific logic (e.g. download types).

packages/content/src/types.ts
ajturner: Yes, I wasn't aware of [ArcGIS IGeometry](https://github.com/Esri/arcgis-rest-js/blob/13bbea4ee995c1a7ba6936ea4ad16c3ff887927f/packages/arcgis-rest-types/src/geometry.ts) - but that probably is the right pattern.

packages/content/src/types.ts
ajturner:
```js
actionLinks: [
  { title: "Open GeoHive Viewer", url: "https://hosting.geohive.ie/custom-viewer?layer=4ef"},
]
```

packages/content/src/types.ts
ajturner: A component provides a UX to read/write info from the `item.licenseInfo`. If we need better license metadata, then we can add explicit parameters to `item.properties`. 

LicensePicker

ajturner: ^^ the above interface is intended to search for Items of type "Content License" that contains both the explicit legal text to put into `licenseInfo` and the likely `licenseItem.properties.capabilities` noted in the type.

packages/content/src/types.ts
ajturner: yes - this would be in `item.properties.metrics.shared=true` per [epic discussion](https://esriarlington.tpondemand.com/entity/107177-individual-content-usage-metrics-for-team) (and US or other doc link I can't find)

packages/content/src/types.ts
ajturner: Dataset measured January 15-March 24 (so created March 24)
Dataset processed + finally pushed to web April 10
Dataset data updated to fix bad valued on September 9

packages/content/src/types.ts
ajturner: @tomwayson that's what I'm suggesting. It's perhaps too implied that `'home'` is a well-known value. Instead, we could instead later add a more explicit `content.home = ...`

packages/content/src/types.ts
ajturner: yes. good point. generally, I would leave the final definition until we have a story to add this capability. The purpose here is to illustrate the plan.

```ts
export interface IHubGeography {
  title?:string;
  coverage?: 'global' | 'regional' | 'local'; // enrichment
  geometry?: string; // serialized JSON, or should this refer to /resources/boundary.json
  source: string; // feature layer + feature URL. e.g. "https://server.cityx.gov/FeatureService/0/53"
  item?: string; // item ID used for setting geometry
}
```

packages/content/src/types.ts
ajturner: yes, except that's _platform jargon_ whereas else where we use `*Number` to be for type number. Do we deviate here?

packages/content/src/types.ts
ajturner: Nice. I like `IHubCommunity`

ajturner:
Thanks for the great feedback. The comments seemed to focus on specific attributes for semantics (`IHubCommunity` is better than `IHubSocial`), clarity _what are `actionConfigs?_ and implementation details _where is `content.metrics.visibility` stored?_.

I didn't see any "this is the wrong direction."

I propose the updates based on comments here is the nominal _Design_ that can then be implemented pragmatically by @tomwayson @mjuniper as part of the initial prototype and then we can task other such as  @rweber-esri @MarkhamShofner etc. with implementing their relevant sections. 

How does that sound?

packages/content/src/types.ts
mikeringrose: Yes. This is a library meant to be used in a larger ecosystem where an Item has a certain meeting and what you can do with an Item's id, such creating links to its resources. A person writing a method can _NOT_ make those assumptions if they expect the item may not really be an item.

ajturner: @mikeringrose I'm not clear how to interpret your message. 

Based on the concerns here, it's probably conceptually & technically simpler to keep `IHubEvent` independent of `IContent`. I'll pull it from this model definition.

mikeringrose: @ajturner i'm not sure what part you don't interpret, but you can leave it as is

packages/content/src/types.ts
tomwayson: Many of the derived properties are not bi-directional, and ideally we'd want to keep them from being overwritten and diverging from their source. TS lures you into thinking you can actually do that, but at the end of the day JS consumers can do `content.createdDate = 'ha ha!'`, so we should be judicious about how much effort we put into trying to control this. Seems to me that flagging the ones we intend to not be bi-directional w/ `readonly` is a reasonable start.

packages/content/src/types.ts
tomwayson: that's not platform jargon at all - that's generic database jargon, if jargon at all. to me, `recordNumber` is the index of this record (as in this is numebr 35 out of 100 results). I don't see this as a deviation from the pattern. `Count` is not a qualifier, it's the value itself.

tomwayson:
FYI - see https://github.com/ArcGIS/opendata-ui/blob/d8a37a902bf0de2258e842c8bcaf18bcd6d6d61d/packages/opendata-ui/app/services/content.js 
for a stubbed implementation of an Emebr service's `getContent(id)` that returns an `IContent`. In that case, it's a facade around the existing Hub app's composer service that returns a dataset (Ember Data) record. Ultimately most of the logic in there should move to Hub.js.

tomwayson:
Closing this as most of these ideas were implemented in #303

PR #277, opened 2020-06-18
Downloads package.

rgwozdz:
Adds downloads package with methods for:
* getting metadata for a particular dataset download file from Hub API or Portal API (private/enterprise datasets)
* requesting the export of a dataset to a particular download file format from Hub API or Portal API (private/enterprise datasets)
* polling the status of download export from Hub API or Portal API (private/enterprise datasets)

### DIFF
--- a/packages/downloads/src/hub/format-converter.ts
+++ b/packages/downloads/src/hub/format-converter.ts
@@ -8,6 +8,9 @@ const HUB_FORMAT_LOOKUP: any = {
   'KML': 'kml'
 }
 
+/**
+ * @private
+ */
 export function convertToHubFormat(format: DownloadFormat ): string {
   if (!HUB_FORMAT_LOOKUP[format]) {
     throw new Error (`Unsupported Hub download format: ${format}`)
--- a/packages/downloads/src/hub/format-converter.ts
+++ b/packages/downloads/src/hub/format-converter.ts
@@ -1,0 +1,10 @@
+import { DownloadFormat } from "../download-format"
+
+export function convertToHubFormat(format: DownloadFormat ) {
+  if (format === 'Shapefile') return 'shp';
+  if (format === 'CSV') return 'csv';
+  if (format === 'File Geodatabase') return 'fgdb';
+  if (format === 'GeoJson') return 'geojson';
+  if (format === 'KML') return 'kml';
+  throw new Error (`Unsupported Hub download format: ${format}`)
+}
--- a/packages/downloads/src/hub/format-converter.ts
+++ b/packages/downloads/src/hub/format-converter.ts
@@ -1,10 +1,16 @@
 import { DownloadFormat } from "../download-format"
 
-export function convertToHubFormat(format: DownloadFormat ) {
-  if (format === 'Shapefile') return 'shp';
-  if (format === 'CSV') return 'csv';
-  if (format === 'File Geodatabase') return 'fgdb';
-  if (format === 'GeoJson') return 'geojson';
-  if (format === 'KML') return 'kml';
-  throw new Error (`Unsupported Hub download format: ${format}`)
+const HUB_FORMAT_LOOKUP: any = {
+  'Shapefile': 'shp',
+  'CSV': 'csv',
+  'File Geodatabase': 'fgdb',
+  'GeoJson': 'geojson',
+  'KML': 'kml'
+}
+
+export function convertToHubFormat(format: DownloadFormat ): string {
+  if (!HUB_FORMAT_LOOKUP[format]) {
+    throw new Error (`Unsupported Hub download format: ${format}`)
+  }
+  return HUB_FORMAT_LOOKUP[format];
 }
--- a/packages/downloads/src/hub/hub-poll-download-metadata.ts
+++ b/packages/downloads/src/hub/hub-poll-download-metadata.ts
@@ -32,7 +32,6 @@ export function hubPollDownloadMetadata (params:IHubDownloadMetadataPollParamete
         hubPollDownloadMetadata(params);
       }, pollingInterval);
     }).catch(error => {
-      console.log(error.message)
       return eventEmitter.emit(`${downloadId}PollingError`, { detail: { error } });
     });
 }
--- a/packages/downloads/src/hub/hub-poll-download-metadata.ts
+++ b/packages/downloads/src/hub/hub-poll-download-metadata.ts
@@ -1,0 +1,56 @@
+import { hubRequestDownloadMetadata, IHubDownloadMetadataRequestParams } from "./hub-request-download-metadata";
+import * as EventEmitter from 'eventemitter3';
+
+export interface IHubDownloadMetadataPollParameters extends IHubDownloadMetadataRequestParams {
+  downloadId: string;
+  eventEmitter: EventEmitter;
+  pollingInterval: number;
+}
+
+/**
+ * ```js
+ * import { hubPollDownloadMetadata } from "@esri/hub-downloads";
+ * const params = {
+ *   downloadId: 'abcdef0123456789abcdef0123456789_0::csv'
+ *   host: 'https://hub.arcgis.com,
+ *   datasetId: 'abcdef0123456789abcdef0123456789_0',
+ *   format: 'CSV'
+ * };
+ * hubPollDownloadMetadata(params);
+ * ```
+ * 
+ * Poll the Hub API for status of a dataset export until the download is ready or export/polling fails. Emits `<downloadId>ExportComplete` event when polling loop receives completion status. Emits `<downloadId>ExportError` event when the export fails. Emits `<downloadId>PollingError` event when polling fails.
+ * @param params - parameters that define the download
+ * @param eventEmitter an Event Emitter
+ * @param pollingInterval number of milliseconds for the polling interval
+ */
+export function hubPollDownloadMetadata (params:IHubDownloadMetadataPollParameters): void {
+  const {
+    downloadId,
+    eventEmitter,
+    pollingInterval
+  } = params;
+
+  hubRequestDownloadMetadata(params)
+    .then(metadata => {
+      if (isUpToDate(metadata)) {
+        return eventEmitter.emit(`${downloadId}ExportComplete`, { detail: { metadata } });
+      }
+      if (exportDatasetFailed(metadata)) {
+        return eventEmitter.emit(`${downloadId}ExportError`, { detail: { metadata } });
+      }
+      return setTimeout(() => {
+        hubPollDownloadMetadata(params);
+      }, pollingInterval);
+    }).catch(error => {
+      return eventEmitter.emit(`${downloadId}PollingError`, { detail: { error } });
+    });
+}
+
+function isUpToDate (metadata: any) {
+  return metadata && (metadata.status === 'ready');
+}
+
+function exportDatasetFailed (metadata: any) {
+  return metadata && (metadata.status === 'error_updating' || metadata.status === 'error_creating');
+}
--- a/packages/downloads/src/hub/hub-poll-download-metadata.ts
+++ b/packages/downloads/src/hub/hub-poll-download-metadata.ts
@@ -1,6 +1,9 @@
 import { hubRequestDownloadMetadata, IHubDownloadMetadataRequestParams } from "./hub-request-download-metadata";
 import * as EventEmitter from 'eventemitter3';
 
+/**
+ * @private
+ */
 export interface IHubDownloadMetadataPollParameters extends IHubDownloadMetadataRequestParams {
   downloadId: string;
   eventEmitter: EventEmitter;
@@ -8,21 +11,7 @@ export interface IHubDownloadMetadataPollParameters extends IHubDownloadMetadata
 }
 
 /**
- * ```js
- * import { hubPollDownloadMetadata } from "@esri/hub-downloads";
- * const params = {
- *   downloadId: 'abcdef0123456789abcdef0123456789_0::csv'
- *   host: 'https://hub.arcgis.com,
- *   datasetId: 'abcdef0123456789abcdef0123456789_0',
- *   format: 'CSV'
- * };
- * hubPollDownloadMetadata(params);
- * ```
- * 
- * Poll the Hub API for status of a dataset export until the download is ready or export/polling fails. Emits `<downloadId>ExportComplete` event when polling loop receives completion status. Emits `<downloadId>ExportError` event when the export fails. Emits `<downloadId>PollingError` event when polling fails.
- * @param params - parameters that define the download
- * @param eventEmitter an Event Emitter
- * @param pollingInterval number of milliseconds for the polling interval
+ * @private
  */
 export function hubPollDownloadMetadata (params:IHubDownloadMetadataPollParameters): void {
   const {
@@ -43,6 +32,7 @@ export function hubPollDownloadMetadata (params:IHubDownloadMetadataPollParamete
         hubPollDownloadMetadata(params);
       }, pollingInterval);
     }).catch(error => {
+      console.log(error.message)
       return eventEmitter.emit(`${downloadId}PollingError`, { detail: { error } });
     });
 }
--- a/packages/downloads/src/hub/hub-request-dataset-export.ts
+++ b/packages/downloads/src/hub/hub-request-dataset-export.ts
@@ -3,34 +3,20 @@ import { composeDownloadId } from '../utils';
 import { DownloadFormat } from "../download-format";
 import { convertToHubFormat } from "./format-converter";
 
+/**
+ * @private
+ */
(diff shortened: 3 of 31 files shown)


packages/downloads/src/hub-request-dataset-export.ts
```diff
@@ -0,0 +1,69 @@
...
+  const body = {
+    spatialRefId,
+    spatialRefWkt,
+    format,
+    geometry,
+    where
+  };
+
+  const url = requestBuilder(host, `/api/v3/datasets/${datasetId}/downloads`)
+  return new Promise((resolve, reject) => {
```
dbouwman, line 43: no need for this Promise - fetch itself is promise based

packages/downloads/src/hub-request-dataset-export.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   .then(response => {
+ *     // {
+ *     //   downloadId: 'abcdef0123456789abcdef0123456789_0::csv',
+ *     // }
+ *   });
+ * ```
+ * Request an export of a dataset to a particular file format.
+ * @param params - parameters defining a dataset export job
+ */
+export function hubRequestDatasetExport (params: IHubDatasetExportRequestParameters) {
```
dbouwman, line 23: Are we currently blocked from downloading private content? if not, I'd expect this to take some auth info

packages/downloads/src/hub-request-download-metadata.ts
```diff
@@ -0,0 +1,119 @@
...
+  const queryParams = {
+    spatialRefId,
+    spatialRefWkt,
+    formats: format,
+    geometry: geometry ? JSON.stringify(geometry) : undefined,
+    where
+  };
+  const url = urlBuilder({ host, route: `/api/v3/datasets/${datasetId}/downloads`, query: queryParams });
+
+  return new Promise((resolve, reject) => {
```
dbouwman, line 54: just `return fetch(...).then(...).catch(...)`

packages/downloads/src/portal-request-dataset-export.ts
```diff
@@ -0,0 +1,208 @@
...
+    format,
+    username,
+    token
+  } = params;
+  const [itemId, layerId = 0] = datasetId.split('_');
+  const exportParameters = {
+    layers: [{ id: layerId }]
+  }
+  let exportItemId: string;
+  return new Promise((resolve, reject) => {
```
dbouwman, line 41: ditto :)

packages/downloads/src/portal-request-dataset-export.ts
```diff
@@ -0,0 +1,208 @@
...
+ *   .then(response => {
+ *     // {
+ *     //   downloadId: 'abcdef0123456789abcdef0123456789_0::CSV',
+ *     // }
+ *   });
+ * ```
+ * Request an export of a dataset to a particular file format.
+ * @param params - parameters defining a dataset export job
+ */
+export function portalRequestDatasetExport (params: IPortalDatasetExportRequestParameters) {
```
dbouwman, line 28: Since this is working directly against portal, I'd expect this can handle private items, so... should we pass in auth/token?

packages/downloads/src/portal-request-dataset-export.ts
```diff
@@ -0,0 +1,208 @@
...
+  formData.append('itemId', itemId);
+  formData.append('exportFormat', exportFormat);
+  formData.append('exportParameters', JSON.stringify(exportParameters));
+  formData.append('f', 'json');
+ 
+  fetch(url, {
+    method: 'POST',
+    body: formData
+  })
+  return new Promise((resolve, reject) => {
```
dbouwman, line 82: ditto

packages/downloads/src/portal-dataset-export-request-parameters.ts
```diff
@@ -0,0 +1,7 @@
+export interface IPortalDatasetExportRequestParameters {
+  host: string;
+  datasetId: string;
+  username: string;
```
dbouwman, line 4: instead of username & token, accept an [IAuthenticationManager](https://esri.github.io/arcgis-rest-js/api/request/IAuthenticationManager/)

dbouwman: that will also allow you to skip sending `host` more info in my general comment

packages/downloads/src/portal-poll-export-job-status.ts
```diff
@@ -0,0 +1,233 @@
...
+      if (json.error) {
+        throw new RemoteServerError(json.error.message, url, json.error.code);
+      }
+      return resolve(json);
+    })
+    .catch(reject);
+  });
+}
+
+function createFolder (params: any): Promise<any> {
```
dbouwman, line 168: This exists in rest-js --> https://esri.github.io/arcgis-rest-js/api/portal/createFolder/

packages/downloads/src/portal-poll-export-job-status.ts
```diff
@@ -0,0 +1,233 @@
...
+      if (json.error) {
+        throw new RemoteServerError(json.error.message, url, json.error.code);
+      }
+      return resolve(json);
+    })
+    .catch(reject);
+  });
+}
+
+function shareItemAsOwner (params: any): Promise<any> {
```
dbouwman, line 201: This is also in rest-js, but it's called `setItemAccess` --> https://esri.github.io/arcgis-rest-js/api/portal/setItemAccess/

packages/downloads/src/portal-request-dataset-export.ts
```diff
@@ -0,0 +1,103 @@
...
+interface IExportItemResult {
+  downloadId: string;
+  jobId: string;
+  exportCreated: number;
+  status: string;
+}
+
+// Hoist to REST JS
+function exportItem (params: any): Promise<IExportItemResult> {
+  const {
```
dbouwman, line 62: One thought on this - should we fetch the item and do some type checking on it before we fire off the request to export it? That way we can return a descriptive error, vs whatever the API returns.

dbouwman:
Sorry - did not realize this was WIP - but some thoughts:
- `fetch` returns a promise, no need for extra `new Promise` wrapping :)
- We should pass `authentication` into the export methods or extend the interfaces. `IAuthenticationManager` has a `portal` property, which is the url to the sharing api backing the user session, from that we can derive the Hub API end-point using [getHubApiUrl](https://esri.github.io/hub.js/api/common/getHubApiUrl/). If a user is un-authenticated, calling code can pass in `{portal: 'https://qaext.arcgis.com/sharing/rest' }'. I think this will also work better with the various rest-js functions that want an authentication object
- having some small private functions in modules seems ok to me, but in some of these, we have a number of non-trivial functions. My preference is to export them as `_myInternalFunction`, which allows us to beat up that specific code with tests. Currently the auto-generated docs **do not** exclude functions marked as `@private` but that's the plan for Rest-js, Hub.js and Solution.js
- when we are making calls to the Portal API, we should use [request](https://esri.github.io/arcgis-rest-js/api/request/request/) as that handles all manner of portal idiosyncrasies - from auth federation to error parsing. 
- I will be adding a few new packages to Hub.js today/tomorrow - one will be to hold the types, so plan on moving those over. Another will be for "hub-content" - thoughts on putting these functions into that package?

mikeringrose:
> * having some small private functions in modules seems ok to me, but in some of these, we have a number of non-trivial functions. My preference is to export them as `_myInternalFunction`, which allows us to beat up that specific code with tests. Currently the auto-generated docs **do not** exclude functions marked as `@private` but that's the plan for Rest-js, Hub.js and Solution.js

@dbouwman I strongly disagree with this idea. We should program by interface (by contract) and any private method should be exercised through its public interface.

mikeringrose:
> * I will be adding a few new packages to Hub.js today/tomorrow - one will be to hold the types, so plan on moving those over. Another will be for "hub-content" - thoughts on putting these functions into that package?

I have no issues adding this to "hub-content" but only only works for a limited number of types atm. We need to think about the interface a bit more before expanding it to other types.

rgwozdz:
@dbouwman - this is ready for another look.

I ended up having to use `UserSession` instead of `IAuthenticationManager`; at least one RESTJS method complained about IAuthenticationManger (I believe it was `getItemStatus`).

Waiting on RESTJS merge and publish to hoist the remaining methods from here to there;  Will get code coverage up once that is complete.

packages/downloads/src/hub/format-converter.ts
```diff
@@ -0,0 +1,10 @@
+import { DownloadFormat } from "../download-format"
+
+export function convertToHubFormat(format: DownloadFormat ) {
+  if (format === 'Shapefile') return 'shp';
+  if (format === 'CSV') return 'csv';
+  if (format === 'File Geodatabase') return 'fgdb';
+  if (format === 'GeoJson') return 'geojson';
+  if (format === 'KML') return 'kml';
```
mikeringrose, line 8: These could just be put into a dict
```
const hubFormat = HUB_FORMATS[format]

if (!hubFormat) {
 throw ...
}

return hubFormat
```

packages/downloads/src/hub/hub-request-dataset-export.ts
```diff
@@ -0,0 +1,74 @@
...
+import { composeDownloadId } from '../utils';
+import { DownloadFormat } from "../download-format";
+import { convertToHubFormat } from "./format-converter";
+
+export interface IHubDatasetExportRequestParams {
+  host: string;
+  datasetId: string;
+  format: DownloadFormat;
+  spatialRefId?: string;
+  spatialRefWkt?: string;
```
dbouwman, line 11: Unclear what `spatialRefWkt` is for as it does not seem to be used in the fn below...

packages/downloads/src/hub/hub-request-dataset-export.ts
```diff
@@ -0,0 +1,74 @@
...
+ *   });
+ * ```
+ * Request an export of a dataset to a particular file format.
+ * @param params - parameters defining a dataset export job
+ */
+export function hubRequestDatasetExport (params: IHubDatasetExportRequestParams) {
+  const {
+    host,
+    datasetId,
+    spatialRefId,
```
dbouwman, line 39: Since `spatialRefId` is spread out of the params, and then used in `body` it seems like this prop is not actually optional?

dbouwman: or is everything downstream from this `null` tolerant?

packages/downloads/src/hub/hub-request-dataset-export.ts
```diff
@@ -0,0 +1,74 @@
+import { RemoteServerError } from '../remote-server-error';
+import { composeDownloadId } from '../utils';
+import { DownloadFormat } from "../download-format";
+import { convertToHubFormat } from "./format-converter";
+
+export interface IHubDatasetExportRequestParams {
```
dbouwman, line 6: doc comment for interface

packages/downloads/src/hub/hub-poll-download-metadata.ts
```diff
@@ -0,0 +1,56 @@
+import { hubRequestDownloadMetadata, IHubDownloadMetadataRequestParams } from "./hub-request-download-metadata";
+import * as EventEmitter from 'eventemitter3';
+
+export interface IHubDownloadMetadataPollParameters extends IHubDownloadMetadataRequestParams {
```
dbouwman, line 4: doc comment for interface

packages/downloads/src/portal/portal-poll-export-job-status.ts
```diff
@@ -0,0 +1,178 @@
...
+  jobId: string;
+  exportCreated: number;
+  eventEmitter: EventEmitter;
+  pollingInterval: number;
+  spatialRefId?: string;
+  geometry?: string;
+  where?: string;
+}
+
+export function portalPollExportJobStatus (params:IPortalPollExportJobStatusParams): void {
```
dbouwman, line 51: doc comment for fn

packages/downloads/src/portal/portal-request-dataset-export.ts
```diff
@@ -0,0 +1,43 @@
+import { exportItem } from "@esri/arcgis-rest-portal";
+import { DownloadFormat } from "../download-format";
+
+export interface IPortalDatasetExportRequestParams {
```
dbouwman, line 4: interface doc

packages/downloads/src/utils.ts
```diff
@@ -0,0 +1,40 @@
+interface IQueryParams {
+  [key: string]: string;
+}
+
+export function urlBuilder (params: { host: string, route: string, query?: IQueryParams }): string {
```
dbouwman, line 5: doc comments for exported functions :)

packages/downloads/test/hub/hub-poll-download-metadata.test.ts
```diff
@@ -0,0 +1,219 @@
...
+      expect(err).toEqual(undefined);
+    } finally {
+      done();
+    }
+  });
+
+  it('handle failed export', async done => {
+    try {
+      fetchMock.mock('http://hub.com/api/v3/datasets/abcdef0123456789abcdef0123456789_0/downloads?spatialRefId=4326&formats=csv', {
+        status: 200,
```
dbouwman, line 44: consider moving large json fixtures to /tests/fixtures, exporting them as `const fooFixture = {...}`. in my experience, this makes the tests are easier to read... but at a cost of not seeing the data... so your call. Or I've also had a "base" fixture that I'd clone and then modify the parts that are important for the test so that's "in" the test

packages/downloads/test/utils.test.ts
```diff
@@ -0,0 +1,23 @@
+import { urlBuilder } from "../src/utils";
+
+describe("utils", () => {
```
dbouwman, line 3: Although trivial, and coverage is provided by other tests, `composeDownloadId` is a function exported from utils, so it would be good to have a test here

packages/downloads/src/hub/hub-request-dataset-export.ts
rgwozdz: Ah, we were talking about supporting `spatialRefWkt` in the future, and the first draft of the code had it in there.  Since the current hub API doesn't support I decide to remove, but missed this one. Will remove.

dbouwman: 👍

packages/downloads/src/hub/format-converter.ts
rgwozdz: ✅

packages/downloads/src/hub/hub-request-dataset-export.ts
rgwozdz: Thanks for catching this.  `spatialRefId` is required for Hub export requests; `geometry` and `where` are optional and the API is tolerant of undefined.  Fixed in c735ae5.

packages/downloads/src/hub/hub-poll-download-metadata.ts
rgwozdz: Made private in c735ae5.

packages/downloads/src/utils.ts
rgwozdz: Made private in c735ae5.

packages/downloads/test/hub/hub-poll-download-metadata.test.ts
rgwozdz: Extracted to fixture in c735ae5.

packages/downloads/src/portal/portal-poll-export-job-status.ts
rgwozdz: Made private in c735ae5.

packages/downloads/src/hub/hub-request-dataset-export.ts
rgwozdz: Made private in c735ae5.

packages/downloads/test/utils.test.ts
rgwozdz: Added in c735ae5.

packages/downloads/src/hub/hub-poll-download-metadata.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+      return setTimeout(() => {
+        hubPollDownloadMetadata(params);
+      }, pollingInterval);
+    }).catch(error => {
+      return eventEmitter.emit(`${downloadId}PollingError`, { detail: { error } });
+    });
+}
+
+function isUpToDate (metadata: any) {
```
mikeringrose, line 39: can metadata be defined as
```
metadata: { [key: string]: any }
```
or
```
metadata: object
```
to prevent people passing in a number

PR #280, opened 2020-06-25
Add a logger wrapper class

sonofflynn89:
1. [130847](https://esriarlington.tpondemand.com/entity/130847-add-a-hub-logger-wrapper-method)
2. Description:
Create a isomorphic logger that only logs when a corresponding log level is set.
3. [x] updated `opendata-ui` changelog
4. [x] did not copy code thoughtlessly ([see this issue](https://esriarlington.tpondemand.com/entity/84512-faster-loading-due-to-smaller-build))


ajturner:
@sonofflynn89 Does this logger only work in a browser? we likely would want Hub.js to be isomorphic to node + browser and be able to support the same logger calls.

instead of the core `_isDebugEnabled(Window)` this may be better to have a state parameter that is shared across the library and set in the right context of node via env variable or API query (up to app) or browser can set via the window query.

ajturner:
I would also recommend rather than merely a _truthy_ state for `debug` there is a enumerated `logLevel: "none" | "error" | "warning" | "all"` or similar.

sonofflynn89:
Changed Logger to be Isomorphic

PR #283, opened 2020-06-29
136998 - Hoist Survey sharing logic to hub.js

rweber-esri:
[136998](https://esriarlington.tpondemand.com/entity/136998-chore-hoist-all-sharing-logic-into)

This PR hoists Survey sharing logic to hub.js. Additionally, it adds transactional/rollback support for group sharing similar to what was done for setting Survey item access since we're working with many items.


packages/surveys/src/sharing/get-input-feature-service-model.ts
```diff
@@ -0,0 +1,33 @@
+/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
```
ajturner, line 1: Update the copyright year

packages/surveys/README.md
```diff
@@ -0,0 +1,59 @@
...
+# @esri/hub-surveys
+
+> Module to interact with ArcGIS Hub Surveys in Node.js and modern browsers.
+
+### Example
+
+```bash
+npm install @esri/hub-surveys
+```
+
```
ajturner, line 24: Can you provide some common external usage patterns? The API docs will have all the internals, but it may be difficult for most devs to know the few minimal functions to use when working with Hub feedback domain objects.

packages/common/src/types.ts
```diff
@@ -106,3 +106,50 @@ export type IBBox = number[][];
...
+};
+
+export interface IFormModel extends IModel {
+  item: IFormItem;
+  data?: {
+    [propName: string]: any;
+  }
+};
+
+export interface IFeatureServiceItem extends IItem {
```
ajturner, line 121: Should some of these types and methods be pushed into arcgis-rest-js or solutions.js?

packages/surveys/README.md
rweber-esri: Thanks for pointing out, meant to circle back to this before submitting my PR, will add some helpful docs.

packages/surveys/src/sharing/get-group-sharing-details.ts
```diff
@@ -0,0 +1,56 @@
...
+        // published surveys have a form, feature service, fieldworker, and potentially a stakeholder
+        // if a view group, only share the form & fieldworker
+        // if an edit/update group, share form, feature service, fieldworker, and stakeholder (if it exists)
+        models = (isUpdateGroup && [form, featureService, fieldworker, stakeholder]) || [form, fieldworker];
+      } else {
+        // drafts only have a form & feature service
+        models = [form, featureService];
+      }
+
+      const modelsToShare = models.filter(_ => _);
```
MarvinPerry, line 52: Silly question, but I don't understand this line. What's different between `models` and `modelsToShare`?

rweber-esri: @MarvinPerry 

The `getSurveyModels` method returns a dictionary of all survey models, e.g:

```js
{
  form: { ... },
  featureService: { ... },
  fieldworker: { ... },
  stakeholder: { ... }
}
```

However, _some_ of those items may not exist in the even the user has not yet published the survey, or if someone manually deleted, in which case it may look something like:

```js
{
  form: { ... },
  featureService: { ... },
  fieldworker: undefined,
  stakeholder: undefined
}
```

In which case `models` array could wind up with falsey values, which we need to filter out to prevent API errors. The filter statement just makes sure we weed out any falsey values for items that don't actually exist.

MarvinPerry: This makes all the sense in the world now. Thanks!

packages/common/src/types.ts
rweber-esri: `Hub.js` is a dependency of `Solutions.js`, so probably can't move these types there. I could see an argument for moving `IFormItem` and `IFeatureServiceItem` to `arcgis-rest-js`, although these survey sharing methods would be the only thing using them at the moment. I'd need to take a closer look at any `arcgis-rest-portal` and `hub.js` peer dependencies, suspect we may need to account for breaking API changes by updating to latest.

rweber-esri: Additionally, this level of granularity wrt typing is probably not necessary, I _could_ also just refactor to just use the existing `IModel` and `IItem` interfaces vs `IFormItem`, `IFormModel`, `IFeatureServiceItem`, & `IFeatureServiceModel`.

packages/common/src/utils/is-update-group.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { IGroup } from "@esri/arcgis-rest-types";
+
+/**
+ * Determines if a given IGroup is an update group
+ * @param {IGroup} group The group to evaluate
+ */
+export const isUpdateGroup = (
+  group: IGroup
+): boolean => group.capabilities.indexOf("updateitemcontrol") > -1;
```
mjuniper: Why not `.includes`?

rweber-esri: > Why not `.includes`?

Typically I do, I _thought_ the TS compiler was complaining about `includes`, but I just built without issue. Will change.

PR #286, opened 2020-07-01
136998 - Adjust method signatures to accept ids instead of models

rweber-esri:
A follow-up PR to [my original PR](https://github.com/Esri/hub.js/pull/283) for [hoisting survey sharing logic](https://esriarlington.tpondemand.com/entity/136998-chore-hoist-all-sharing-logic-into).

1) Addresses some cleanup in the `surveys` package, e.g. moving `get-*` methods out of `sharing` dir and into a new `items` dir
2) Changes some of the `get-*` method signatures to accept an `ID` string vs `IModel`.
3) Changes `getStakeholderModel` to use the `Survey2Data` relationship and accept the Form ID instead of a Fieldworker ID
4) Improved `getSurveyModels` performance by make the `getInputFeatureServiceModel` and `getStakeholderModel` calls concurrent (point of #3 above).


rweber-esri:
> Nice. Usually we have used `getModelFromOptions` to allow the client to use _either_ an ID or a model in functions that require an item model, but it looks like it might not make sense in your case(?) @rweber-esri

Good observation @drewctate. Admittedly, this is something I went back and forth on. Ultimately, these methods don't benefit from accepting a full model object (we don't need groups, data, etc) and the existing `shareItemToGroups` and `unshareItemFromGroups` methods from the `hub-common` package use a simplified interface that accepts id strings, so went this direction to be consistent on the sharing front.

PR #290, opened 2020-07-06
Add Teams package, Fix Arrays in Doc

### LINKED ISSUE #285: Docs are wrong anywhere we use shorthand Array notation
Generics Array notation (e.g. `IModel[]`) and shorthand Array notation (e.g. `Array<IModel>`) are semantically equal in TypeScript. Apparently this is not the case, though, for our documentation. The more verbose generics syntax compiles to the expected documentation type of `IModel[]`, however shorthand notation compiles to something like `IModel[] | IModel` which is not right. We use [typedoc](https://github.com/TypeStrong/typedoc) to generate our documentation. Looks like that lib is up to `v0.17.8`, we're currently using `v0.14.2`. We should see if this inconsistency can be corrected by updating the `typedoc` lib and/or tweaking it's configuration. If those efforts fall flat, we should replace all instances of shorthand Array notation with generics notation and investigate enforcing via linting rules.

You can see this issue in action by viewing [IModelTemplate docs](https://esri.github.io/hub.js/api/common/IModelTemplate/). See `assets`, `dependencies` & `resources` properties.

dbouwman:
As it says - just adds the "empty" package for teams.

Over time, we will migrate Team functionality currently in the sites package into this package.

Closes #285

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,11 +75,14 @@ Some useful commands include:
 * [`@esri/hub-common`](./packages/common) - Common TypeScript types and utility functions.
 * [`@esri/hub-initiatives`](./packages/initiatives) - Helper functions for working with [Initiatives](http://doc.arcgis.com/en/hub/initiatives/initiatives-overview.htm).
 * [`@esri/hub-annotations`](./packages/annotations) - Helper functions for working with Annotations.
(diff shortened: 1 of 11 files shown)


rweber-esri:
@dbouwman Should we update the [root README](https://github.com/Esri/hub.js/blob/master/README.md#packages) to include the new package?

PR #303, opened 2020-08-01
getContent()

tomwayson:
@ajturner 

Here's my first pass at `getContent()` - take a look at the `TODO:` and `TW:` in the code comments.


packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+} from "@esri/hub-common";
 
+function buildUrl (base: string, path: string): string {
+  const baseUrl = base.endsWith('/') ? base : `${base}/`;
+  const url = new URL(path, baseUrl);
+  return url.toString();
+}
+
+// TODO: make this real; use request() under the hood?
+export function hubRequest (url: string, requestOptions?: IHubRequestOptions) {
```
mikeringrose, line 19: maybe passing the method as the first parameter since it's required. we could always do partial application and have more specific methods such as `hubGet` and `hubPost`

```
export function hubRequest (method: string, url: string | URL, requestOptions?: IHubRequestOptions) { ... }
export const hubGet = hubRequest.bind(null, 'GET')
```
we should define the return type.

packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+  return fetch(url, {
```
mikeringrose, line 20: I know this is a bigger change, but if we're to be isomorphic, it'd be nice if we didn't require fetch.

packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+      // TODO: base on request requestOptions?
+      'Content-Type': 'application/json'
+    },
+    // TODO: base on requestOptions.params?
+    // body: JSON.stringify(body)
+  }).then(resp => {
+    if (resp.ok) {
+      return resp.json();
+    } else {
+      throw resp.statusText;
```
mikeringrose, line 32: i'm not a huge fan of throwing strings.

packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+ * @export
+ */
+function datasetToContent(
+  // TODO: IDataset
+  dataset: any,
+  portalApiUrl: string
+): IHubContent {
+  if (!dataset) {
+    return;
+  }
```
mikeringrose, line 152: part of the contract should be a non-null object. the caller should do:
```
return dataset && datasetToContent(...)
```

packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+  // TODO: should we return a different subtype of IHubContent for this?
+  // TODO: should this be based on existence of attributes instead of hubType?
+  if (content.hubType === 'dataset') {
+    content.recordCount = recordCount;
+  }  
+  // TODO: any remaining enrichments
+  return content;
+}
+
+function datasetToItem (dataset:any):IItem {
```
mikeringrose, line 183: use a better type than any. any allows string, numbers, etc.

packages/common/src/types.ts
```diff
@@ -108,26 +108,138 @@ export type IBatch = any[];
...
+  | "initiative"
+  | "template"
+  | "organization";
+export type VisibilityOptions = "private" | "org" | "public";
+export type ControlOptions = "view" | "edit" | "admin";
+
+// TW: basing this on https://opendata.arcgis.com/api/v3/datasets/7a153563b0c74f7eb2b3eae8a66f2fbb_0/
+// but that may not be accurate
+export interface IHubGeography {
+  size?: number; // what is this?
```
ajturner, line 158: I don't recall the details of why this was included (Dmfenton, pranav, alex, other interns). I don't see any notes in emails. I think we can drop it, unless you see a reference in live code.

This is similar to your dropping other speculative geography attributes (name, source) until they're explicitly added.

packages/common/src/types.ts
```diff
@@ -108,26 +108,138 @@ export type IBatch = any[];
...
+    // overrides item.access with more attributes. could flatten.
+    visibility: VisibilityOptions; // item.access
+    control?: ControlOptions; // itemControl
+    groups?: IGroup[]; // item.sharing.groups via content/users/:username/items/:id
+  };
 
-export type IRevertableTaskResult = IRevertableTaskSuccess | IRevertableTaskFailed;
+  // Explicit data information since this is a common confusion + bug report
+  createdDate: Date; // formal metadata || new Date(item.created)
+  createdDateSource?: string; // description of what was used for this attribute
```
ajturner, line 180: comment may be explicit that this is meant to be the attribute key, e.g. `item.modified` or `item.metadata.created_date` which would likely match to a UI string key.

packages/content/src/content.ts
ajturner: is there an error/response type we can use consistently that includes both code and strings? @mikeringrose @tomwayson

packages/content/src/content.ts
```diff
@@ -1,8 +1,292 @@
...
+  const url = buildUrl(requestOptions && requestOptions.hubApiUrl, `/datasets/${id}`);
+  return hubRequest(url, requestOptions)
+  .then(resp => {
+    const dataset = resp && resp.data && resp.data[0];
+    const portalApiUrl = getPortalApiUrl(requestOptions && requestOptions.portalSelf);
+    return datasetToContent(dataset, portalApiUrl);
+  });
+}
+
+function getContentFromAgo(id: string, requestOptions?: IHubRequestOptions): Promise<IHubContent> {
```
ajturner, line 52: this method is mis-named since it would also apply to getting content from enterprise. Can we keep the name as `getContentFromPortal` @tomwayson

packages/common/src/types.ts
tomwayson: thanks @ajturner. @bcamper do you know what the `boundary.size` property is?

bcamper: I think it is the actual size of the ring data?? Not sure why it is being stored and don't think it has a functional purpose.

But also, I believe the data structure for the concave hulls is changing (I had filed a bug about how it's neither compliant Esri JSON geometry nor GeoJSON) - @mikeringrose, do you know that status?

packages/content/src/content.ts
tomwayson: that was just temporary, I've added `DatasetResource` using https://github.com/mike-north/jsonapi-typescript

packages/content/src/content.ts
tomwayson: This now throws `RemoteServerError`, which I've copied over from the downloads package and added a chore to remove the duplicate next breaking change to #305

packages/common/src/types.ts
tomwayson: I've removed `size` for now

tomwayson: @mikeringrose @bcamper  we need to at least figure out what format the Hub API _intends_ to return the `boundary` before we merge this PR.

packages/content/src/content.ts
tomwayson: I think we can get this from https://esri.github.io/hub.js/api/common/IHubRequestOptions/#fetch the way rest-js does

packages/content/src/content.ts
tomwayson: I think we can get this from https://esri.github.io/hub.js/api/common/IHubRequestOptions/#httpMethod the way rest-js does

packages/common/src/types.ts
bcamper: TP issue I filed is here: https://esriarlington.tpondemand.com/entity/138758-incorrectly-formatted-concave-hull-geometries
PR in Hub API: https://github.com/ArcGIS/hub-indexer/pull/2242

I ran into this a few months ago when reusing some of the graphics sorting code for map notes, but abandoned that change because of this issue.

But I did want to ensure we're storing compliant geometries (and there were other oddities in the current format) before we expand the use of additional geometries with things like boundary picker UI etc.

packages/content/src/content.ts
tomwayson: unfortunately that contract is not enforceable in JS, only in TS.

the way I've dealt w/ that in the past is to check for the presence of required arguments w/in the fn and return early if they're not there. that keeps all consuming code from having to use `&&` and avoids those ambiguous "cannot read property of undefined" errors. The downside is that it is the consuming code that knows what to do when the arg is missing, and just returning `undefined` might cause silent errors.

I suppose a better way is to explicitly throw a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) when required arguments are missing. If we want to adopt that style I'll add it to this PR.

packages/common/src/resources/get-item-thumbnail-url.ts
```diff
@@ -11,14 +11,17 @@ import { getPortalApiUrl } from "../urls";
...
-  hubRequestOptions: IHubRequestOptions
+  portalApiUrlOrHubRequestOptions: string | IHubRequestOptions
 ): string | null {
-  if (item.thumbnail) {
-    const portalRestUrl = getPortalApiUrl(hubRequestOptions.portalSelf);
-    const itemUrl = `${portalRestUrl}/content/items/${item.id}`;
-    const url = `${itemUrl}/info/${item.thumbnail}`;
-    return url;
-  } else {
+  if (!item || !item.thumbnail) {
```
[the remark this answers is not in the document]
tomwayson: see https://github.com/Esri/hub.js/pull/303#discussion_r466555191

packages/common/src/types.ts
tomwayson: Thanks @bcamper. So it looks like the intent at least is to store/return Esri geometries, so I'll consider things resolved on this end - they should "just work" once the bug is fixed on the server.

tomwayson:
I'm working on improving comments for the docs site. I'm going to leave @mikeringrose's suggestions for `hubRequest()` to future PRs since they can be implemented w/ the signature that I put forth in this PR.

The only remaining issue is https://github.com/Esri/hub.js/pull/303#discussion_r466555191

tsconfig.json
```diff
@@ -39,6 +39,7 @@
     // "baseUrl": "./",                       /* Base directory to resolve non-absolute module names. */
     // "paths": {},                           /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
     // "rootDirs": [],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
+    "resolveJsonModule": true,
```
tomwayson, line 42: I wanted to apply this to only the `content/test` folder so we could `import` mock JSON responses directly. I put at tsconfig.json file in that folder that extends this one and added this one rule, but it didn't seem to work. I don't _think_ this affects the runtime code, but I'm open to other suggestions as to how to scope it to just the test files.

packages/content/src/content.ts
mikeringrose: when i meant by contract, i was more referring to https://en.wikipedia.org/wiki/Design_by_contract, specifically,
>The DbC approach assumes all client components that invoke an operation on a server component will meet the preconditions specified as required for that operation.
i mean the caller has to know if they have a dataset. this is a minor point, so i'm fine leaving the check

Issue #309, opened 2020-08-11
cloneObject doesn't work on dates


tomwayson:
Add a `Date` to the "can clone a shallow object" as follows:

```ts
  it("can clone a shallow object", () => {
    const obj = {
      color: "red",
      length: 12,
      startDate: new Date() 
    } as any;
    const c = cloneObject(obj);
    expect(c).not.toBe(obj);

    ["color", "length", "startDate"].map(prop => {
      expect(c[prop]).toEqual(obj[prop]);
    });
  });
```

The test will fail w/ `Expected Object({  }) to equal Date(Tue Aug 11 2020 12:40:13 GMT-0700 (Pacific Daylight Time)).`

PR #310, opened 2020-08-11
fix(cloneObject): fix cloneObject's handling of Date properties

tomwayson:
AFFECTS PACKAGES:
@esri/hub-common

ISSUES CLOSED: #309

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -19,10 +19,15 @@ export function cloneObject<T>(obj: T): T {
     clone = obj.map(cloneObject);
   } else if (typeof obj === "object") {
     for (const i in obj) {
-      if (obj[i] != null && typeof obj[i] === "object") {
-        clone[i] = cloneObject(obj[i]);
-      } else {
-        clone[i] = obj[i];
+      /* istanbul ignore next no need to deal w/ other side of hasOwnProperty() */
+      if (obj.hasOwnProperty(i)) {
+        const value = obj[i];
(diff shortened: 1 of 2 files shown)


tomwayson:
@drspacemanphd 

Once this is approved, can I cut a patch release w/ your recently merged revert: #304

packages/common/src/util.ts
```diff
@@ -19,10 +19,15 @@ export function cloneObject<T>(obj: T): T {
...
-      } else {
-        clone[i] = obj[i];
+      /* istanbul ignore next no need to deal w/ other side of hasOwnProperty() */
+      if (obj.hasOwnProperty(i)) {
+        const value = obj[i];
+        const isDate = value instanceof Date;
+        if (value != null && typeof value === "object" && !isDate) {
+          clone[i] = cloneObject(value);
+        } else {
+          clone[i] = value;
```
dbouwman, line 29: Since Date is an object I'd expect the clone and original to share a ref to the same Data instance... I think we'd want to convert it to a timestamp then create a new date instance from that... no?

tomwayson: good catch @dbouwman - i updated the code as you suggest and updated the test to ensure it's not a reference to the the original

drspacemanphd:
> @drspacemanphd
> 
> Once this is approved and merged, can I cut a patch release w/ your recently merged revert: #304

@tomwayson I do not see a reason why we can't cut a release.

PR #315, opened 2020-08-12
Fix search for cached download files on Portal.

rgwozdz:
Fixed broken search for previously cached download files with a format-name that includes whitespace (e.g., `Feature Collection`). Keywords actually don't require inclusion of the export-format type, since the search already includes a "type", e.g. `type:"Feature Collection"`.

This PR removes `exportFormat` from cached file search and removes it from the item update that tags the cached file with keywords.


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased][HEAD]
 
+* Bug Fixes
+   * **downloads**: fix broken search for previously cached download files with a format-name including whitespace. Keywords actually don't require inclusion of the export-format type, since the search already includes a "type" which refers to the export-format.
```
tomwayson, line 11: you shouldn't need to modify the CHANGELOG if you run `npm run c`

rgwozdz: Ah, thanks.  I didn't know that!

PR #317, opened 2020-08-13
feat(request): add hubApiRequest() for making requests to the Hub API

tomwayson:
AFFECTS PACKAGES:
@esri/hub-common

### DIFF
--- a/packages/common/src/request.ts
+++ b/packages/common/src/request.ts
@@ -38,11 +38,12 @@ export function hubApiRequest(
   const options: IHubRequestOptions = {
     ...{
       hubApiUrl: "https://opendata.arcgis.com/api/v3/",
-      httpMethod: "POST",
-      fetch
+      httpMethod: "GET"
     },
     ...requestOptions
   };
+  // use fetch override if any
+  const _fetch = options.fetch || fetch;
   // merge in default headers
   const headers = {
     ...{ "Content-Type": "application/json" },
@@ -54,17 +55,15 @@ export function hubApiRequest(
     host: options.hubApiUrl,
     path: route
   });
-  return options
-    .fetch(url, {
-      method: options.httpMethod,
-      headers
-      // body: JSON.stringify(body)
-    })
-    .then(resp => {
-      if (resp.ok) {
-        return resp.json();
-      } else {
-        throw new RemoteServerError(resp.statusText, url, resp.status);
-      }
-    });
+  return _fetch(url, {
+    method: options.httpMethod,
+    headers
+    // body: JSON.stringify(body)
+  }).then(resp => {
+    if (resp.ok) {
+      return resp.json();
+    } else {
+      throw new RemoteServerError(resp.statusText, url, resp.status);
+    }
+  });
 }
--- a/packages/common/src/request.ts
+++ b/packages/common/src/request.ts
@@ -1,4 +1,4 @@
-import { IHubRequestOptions } from "./index";
+import { IHubRequestOptions, buildUrl } from "./index";
 
 /**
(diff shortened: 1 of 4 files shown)


tomwayson:
image

packages/common/src/request.ts
```diff
@@ -17,34 +17,54 @@ export class RemoteServerError extends Error {
...
-    }
+export function hubApiRequest(
+  route: string,
+  requestOptions?: IHubRequestOptions
+) {
+  // merge in default request options
+  const options: IHubRequestOptions = {
+    ...{
+      hubApiUrl: "https://opendata.arcgis.com/api/v3/",
+      httpMethod: "POST",
```
rgwozdz, line 41: Why is `POST` the default method?

packages/common/src/request.ts
```diff
@@ -17,34 +17,54 @@ export class RemoteServerError extends Error {
...
+  // build Hub API URL
+  const url = buildUrl({
+    host: options.hubApiUrl,
+    path: route
   });
+  return options
+    .fetch(url, {
+      method: options.httpMethod,
+      headers
+      // body: JSON.stringify(body)
```
rgwozdz, line 61: Is the commented-out line intentional?  Seems like it would prevent POST body from being sent.

packages/common/src/request.ts
```diff
@@ -17,34 +17,54 @@ export class RemoteServerError extends Error {
...
+export function hubApiRequest(
+  route: string,
+  requestOptions?: IHubRequestOptions
+) {
+  // merge in default request options
+  const options: IHubRequestOptions = {
+    ...{
+      hubApiUrl: "https://opendata.arcgis.com/api/v3/",
+      httpMethod: "POST",
+      fetch
```
rgwozdz, line 42: Why package `fetch` inside `options`?  Couldn't you just do something like:

`const fetchMethod = requestOptions.fetch || fetch;`

packages/common/src/request.ts
tomwayson: that's a TODO, I'm just implementing the bare minimum that I need to ship `getContent()`

packages/common/src/request.ts
tomwayson: That's what we did in REST JS, b/c most of the requests made by the lib were POST (CUD vs R), and all the ArcGIS REST API "query" end points accept POST as well as GET.

I'd love your insight re: the Hub API. If you think GET is the better default, we can roll w/ that

packages/common/src/request.ts
tomwayson: you'd have to at least do `const fetchMethod = (requestOptions && requestOptions.fetch) || fetch;`. I don't see how that's any better than the current code

packages/common/src/request.ts
rgwozdz: That's a tough one then.  I'd venture that most of the requests going to Hub API are going to be GETs - check out the API docs here: https://github.com/ArcGIS/hub-indexer/blob/master/docs/api.md.  Also, since node-fetch's default is `GET` I generally assume that it is the default of request methods.  If it were up to me I would default to GET.

packages/common/src/request.ts
rgwozdz: Just readability, but that just opinion, so I will leave it up to you.  I saw that the `options.fetch` was possibly being modified, so I assumed `options` was going to be passed on to another method rather than just accessed and executed in the current block.

tomwayson: gotcha, I've made that change for readability's sake.

PR #318, opened 2020-08-14
fix(hubApiRequest): url should include /api/v3

tomwayson:
AFFECTS PACKAGES:
@esri/hub-common


rgwozdz:
Just thinking, we should probably be setting the api version, e.g. `v3`, with a config setting.

PR #324, opened 2020-08-19
147718 Port over workflow for adding users to teams, events and groups

sonofflynn89:


packages/common/src/groups/add-users-workflow/utils/_get-invite-users.ts
```diff
@@ -0,0 +1,15 @@
+import { IUser } from "@esri/arcgis-rest-auth";
+import { _getAutoAddUsers } from "./_get-auto-add-users";
+
+/**
+ * @private
```
dbouwman, line 5: This is the sort of function that could really use some explanation... I *think* this returns users that we can't auto-add?

sonofflynn89: I'll add descriptions to this and the other helpers to make them more clear

PR #329, opened 2020-08-24
Proposed npm updates

MikeTschudi:
* ensure all projects use tslib ^1.13.0
* updated rollup-* to latest to get latest iltorb
* removed circular dependency


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased][HEAD]
 
+### General
+* ensure all projects use tslib ^1.13.0
+* updated rollup-* to latest to get latest iltorb
```
tomwayson, line 12: looks like [iltorb has been deprecated](https://www.npmjs.com/package/iltorb). How is it used in this repo?

tomwayson, line 12: Rather than manually add comments to the changelog we should use `npm run c`

MikeTschudi: Thanks!

MikeTschudi: A fresh install of arcgis-rest-js and hub.js failed due to a reference to __spreadArrays. Dave figured it out and [solved it for arcgis-rest-js](https://github.com/Esri/arcgis-rest-js/pull/732)
> ok... that was nasty, but the problem was that the rollup plugin used to generate the umd output was using tslib 1.9.3. It's unclear to me how/why it worked with the package-lock file, but the only way I could get a clean build from a fresh npm i was to bump the rollup-plugin-typescript2 to v0.22.0 which depends on tslib 1.10.0 - which is when tslib added __spreadArrays

`iltorb` was one of the npm complaints after getting past `__spreadArrays`.  Updating rollup-* stopped the complaining, so the comment in the changelog is misleading.

My goal was to apply the same modifications to hub.js, and I was able to get a successful fresh install as a result.

MikeTschudi:
Thanks, @tomwayson!

tomwayson:
this has been released in `6.4.0`

PR #333, opened 2020-08-26
add overloads to getPortalUrl and consistency to signatures of fns that use it

tomwayson:
Under the hood `getPortalUrl()` now falls back to calling [`getPortalUrl`](https://esri.github.io/arcgis-rest-js/api/portal/getPortalUrl/), which means it:
- will take an `IRequestOptions` in addition to `IHubRequestOptions`
- will default to the `https://www.arcgis.com`, making the argument optional

image

The signatures of all fns that use this have been updated accordingly as well.

Finally, some fns that called `getPortalUrl()` from `@esri/arcgis-rest-request` have been updated to use the Hub.js version.


tomwayson:
Also, `getItemThumbnailUrl()` has been updated to take an optional `token` argument and append it to the URL if needed.

packages/content/src/portal.ts
```diff
@@ -44,17 +44,17 @@ export function withPortalUrls(
   requestOptions: IHubRequestOptions
 ) {
   const newContent = cloneObject(content);
-  const portal = requestOptions.portalSelf;
+  // const portal = requestOptions.portalSelf;
```
rgwozdz, line 47: Is this commented code still needed?

PR #335, opened 2020-08-26
Add Site ID to Initiative if it was created with the Site

dbouwman:
Convoluted, but basically when we create a site, if the user has Hub Premium licensing, we also create an Initiative item. But we create that _before_ we create the site item... so we can't interpolate the siteId into the Initiative.

This adds a step to the createSite fn to lookup the initiative and add the siteId to it.

That step is backed by a function that we will use in opendata-ui to "heal" initiatives that do not have a siteId in them

### DIFF
--- a/packages/initiatives/src/activate.ts
+++ b/packages/initiatives/src/activate.ts
@@ -27,6 +27,7 @@ import { getProp } from "@esri/hub-common";
  * @param {any} groupIds hash of group props and ids
  * @param {IRequestOptions} requestOptions
  * @returns {Promise<IInitiativeModel>}
+ * @private
  */
 export function activateInitiative(
   template: string | any,
--- a/packages/initiatives/src/update-initiative-site-id.ts
+++ b/packages/initiatives/src/update-initiative-site-id.ts
@@ -1,0 +1,62 @@
+/* Copyright (c) 2020 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+import { IItem } from "@esri/arcgis-rest-portal";
+import { IModel, IHubRequestOptions, getProp, isGuid } from "@esri/hub-common";
+import { updateItem, getItem } from "@esri/arcgis-rest-portal";
+
+/**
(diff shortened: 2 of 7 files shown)


packages/initiatives/src/update-initiative-site-id.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Used during createSite(...) and any time we need to
+ * connect a different Site to an Initiative
+ *
+ * @param initiativeItemId string | IItem | IModel
+ * @param siteId string
+ * @param hubRequestOptions IHubRequestOptions
+ */
+export function updateInitiativeSiteId(
+  maybeModel: string | IItem | IModel,
```
drewctate, line 19: Why the flexibility?

packages/initiatives/src/activate.ts
```diff
@@ -27,6 +27,7 @@ import { getProp } from "@esri/hub-common";
  * @param {any} groupIds hash of group props and ids
  * @param {IRequestOptions} requestOptions
  * @returns {Promise<IInitiativeModel>}
+ * @private
```
drewctate, line 30: What is the thinking behind this?

packages/initiatives/src/update-initiative-site-id.ts
dbouwman: in Hub.js we will be passing in a Guid, but from w/in Hub, where we want to "heal" this, we may have the Item or Model already - thus allowing us to skip the fetch

packages/initiatives/src/activate.ts
dbouwman: At this point, Initiative activation is not used, and marking it private keeps it out of the docs. I've also gotten more draconian and the whole initiaives package is hidden from the docs... so... I guess this is not strictly needed

tomwayson:
I see there are changes in _both_ initiatives and sites, so when I do a release, will I have to bump site's `peerDependencies`?

PR #337, opened 2020-08-26
Support where filter on portal downloads.

rgwozdz:
Adds support for a `where` filter on download requests using the Portal API.


packages/downloads/test/portal/portal-request-dataset-export.test.ts
```diff
@@ -35,7 +35,7 @@ describe("portalRequestDatasetExport", () => {
         {
           id: "abcdef0123456789abcdef0123456789",
           exportFormat: "CSV",
-          exportParameters: { layers: [Object({ id: 0 })] },
+          exportParameters: { layers: [Object({ id: 0, where: undefined })] },
```
tomwayson, line 38: is this needed? will clients need to do this?

PR #343, opened 2020-08-31
regression web mapping applications cant save drafts

drewctate:


packages/sites/src/drafts/_include-list-from-item-type.ts
```diff
@@ -1,27 +1,26 @@
...
+import { isSite } from "../is-site";
+import { isPage } from "../pages";
+import { IItem } from "@esri/arcgis-rest-portal";
 
 /**
  * Returns the right include list for the item type.
  * @param siteOrPageModel - the site or page model
  * @param isPortal - if we're on portal or not
  * @private
  */
```
benstoltz, line 12: Update comment to take into consideration function changes.

packages/sites/src/is-site.ts
```diff
@@ -0,0 +1,15 @@
+import { IItem } from "@esri/arcgis-rest-portal";
+import { includes } from "@esri/hub-common";
+
+/**
+ * Determines whether an item is a site item or not
+ * @param item - the item
+ * @param isPortal - whether or not we are on portal
+ */
+export function isSite(item: IItem) {
```
benstoltz, line 9: Either remove isPortal param from comment....or perhaps you forgot something?

packages/sites/src/pages/is-page.ts
```diff
@@ -0,0 +1,15 @@
+import { IItem } from "@esri/arcgis-rest-portal";
+import { includes } from "@esri/hub-common";
+
+/**
+ * Determines whether an item is a page item or not
+ * @param item - the item
+ * @param isPortal - whether or not we are on portal
+ */
```
benstoltz, line 8: ditto down here for isPortal

Issue #341, opened 2020-08-28
getContent() should handle un-indexed (usually private) content


tomwayson:
The hub indexer cannot index private items, so we need to fetch them at runtime from Online.

@drewctate took a glance and it appears the way the composer service in the Hub app works is to try and fetch from the Hub API, and if it gets a 404 then it tries to fetch from Online w/ the user's credentials.

Issue #322, opened 2020-08-17
getContent() returns IHubDataset for dataset types


tomwayson:
Define `IHubDataset` that extends `IHubContent` w properties like `fields`, `geometryType`, `layer`?, `server`?, etc, as needed by the Hub app, and have `getContent()` return that when the content's `hubType` is `dataset`.

ajturner:
Should this method provide an option to include or not include these additional metadata? When these extended properties are not cached via harvesting (e.g. public datasets) then there could be significant XHR to fetch these from different endpoints. 

Relatedly, should the extended properties be in a type-specific property to make it clear they're type specific, maybe optional, and keep the primary object readable/coherent. For example `dataset.extended.fields`

tomwayson:
Some things we discussed offline:
- we won't used a property bag like `content.extended` - we'll just add the properties directly to `content` (i.e. `IHubDataset extends IHubContent`)
- we'll use an request option like `include` (or `exclude`, depending on what the default behavior is)
- should default behavior be to perform whatever additional async fetches are needed to return a complete dataset and users would have to opt-out, or to not perform any (opt-in)?

ajturner:
I’d like to hear @mikeringrose thoughts on default includes.

I recall seeing an update to Urban GraphQL API only default return an ID unless fields were explicitly include. 

However, that would be onerous request to include all common Resources or Content metadata. We don’t want to require knowledge of the data store implementation (what’s in Portal vs Server) - but is it sensible to say “by default the model includes [...] attributes”

tomwayson:
To me this option is not going to specify which fields are returned, but rather which additional xhrs are going to be made _if needed_ (i.e. the content is coming from the portal API instead of the Hub API). I don't think we should derive the list of xhrs from a list of fields (i.e. "consumer wants `advancedQueryCapabilities`, therefore I need to fetch the layer info"). In other words `getContent()` should always return all the properties it has fetched, and what the consumer will have  control over is what additional fetches should be made if needed.

To me that implies a better name than `include` would be `ensure`. For example, if a consumer specifies `ensure: ['service', 'recordCount']`:
- when fetching from portal (i.e. private item, or in enterprise), additional xhrs would be made to get the service info (`fields`, `advancedQueryCapabilities`, `maxRecordCount`, etc) and query the layer's `recordCount`, and the corresponding properties would be populated
[...]

PR #345, opened 2020-08-31
fix(getContent): fix handle un-indexed (usually private) content

tomwayson:
AFFECTS PACKAGES:
@esri/hub-content

ISSUES CLOSED: #341

### DIFF
--- a/packages/content/src/content.ts
+++ b/packages/content/src/content.ts
@@ -19,7 +19,12 @@ export function getContent(
   if (requestOptions && requestOptions.isPortal) {
     return getContentFromPortal(id, requestOptions);
   } else {
-    return getContentFromHub(id, requestOptions);
+    return getContentFromHub(id, requestOptions).catch(() => {
+      // TODO: inspect error?
+      // dataset is not in index (i.e. might be a private item)
+      // try fetching from portal instead
+      return getContentFromPortal(id, requestOptions);
+    });
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/content/src/content.ts
```diff
@@ -19,7 +19,12 @@ export function getContent(
   if (requestOptions && requestOptions.isPortal) {
     return getContentFromPortal(id, requestOptions);
   } else {
-    return getContentFromHub(id, requestOptions);
+    return getContentFromHub(id, requestOptions).catch(() => {
+      // TODO: inspect error?
+      // dataset is not in index (i.e. might be a private item)
+      // try fetching from portal instead
+      return getContentFromPortal(id, requestOptions);
```
drewctate, line 26: I don't love the fact that we use `portal` to refer both to enterprise sites and to the portal API here (and everywhere else in the app). Maybe we could consider calling this function `getContentFromPortalApi` or something. Anyway, your call.

drewctate:
Do we need a changelog entry?

tomwayson:
Thanks @drewctate, changelog will be _generated_ ;)

packages/content/src/content.ts
tomwayson: Fair point. I would suggest we also rename `getContentFromHubApi` as well then.

mikeringrose, line 26: is this supported for all item types? if somebody is selecting a feature layer dataset, a failed request would return a feature service item. is that acceptable?

tomwayson: good point, but I think that is out of scope of this PR, b/c `getContent()` doesn't yet support feature layer/service datasets.

We should address this as part of #322

PR #352, opened 2020-09-01
getContent() handles slugs

tomwayson:
The slug handling logic is copied (roughly) from the composer service in the Hub app.

I also had to add the ability for `hubRequest()` to pass params, either in the query string or body of the request, and I'd like @rgwozdz to review that b/c we'll eventually want to have all the download code start using that fn.

### DIFF
--- a/packages/common/src/request.ts
+++ b/packages/common/src/request.ts
@@ -50,16 +50,26 @@ export function hubApiRequest(
     ...{ "Content-Type": "application/json" },
     ...options.headers
   };
-  // TODO: build query params/body based on requestOptions.params?
+  // build query params/body based on requestOptions.params
+  let query;
+  let body;
+  if (options.httpMethod === "GET") {
+    // pass params in query string
+    query = options.params;
+  } else {
+    // pass params in request body
+    body = JSON.stringify(options.params);
+  }
   // build Hub API URL
   const url = buildUrl({
     host: options.hubApiUrl,
-    path: `/api/v3/${route}`.replace(/\/\//g, "/")
+    path: `/api/v3/${route}`.replace(/\/\//g, "/"),
+    query
   });
   return _fetch(url, {
     method: options.httpMethod,
-    headers
-    // body: JSON.stringify(body)
+    headers,
+    body
   }).then(resp => {
     if (resp.ok) {
       return resp.json();
--- a/packages/common/src/request.ts
+++ b/packages/common/src/request.ts
@@ -37,17 +37,15 @@ export function hubApiRequest(
 ) {
   // merge in default request options
   const options: IHubRequestOptions = {
-    ...{
-      hubApiUrl: "https://opendata.arcgis.com/api/v3/",
-      httpMethod: "GET"
-    },
+    hubApiUrl: "https://opendata.arcgis.com/api/v3/",
+    httpMethod: "GET",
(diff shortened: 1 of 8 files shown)


tomwayson:
ah dangit - forgot to add context (`orgKey`), will add and re-push

packages/common/src/request.ts
```diff
@@ -50,16 +50,26 @@ export function hubApiRequest(
...
-  // TODO: build query params/body based on requestOptions.params?
+  // build query params/body based on requestOptions.params
+  let query;
+  let body;
+  if (options.httpMethod === "GET") {
+    // pass params in query string
+    query = options.params;
+  } else {
+    // pass params in request body
+    body = JSON.stringify(options.params);
```
tomwayson, line 61: @rgwozdz can you verify that for POST operations the API expects the body to be stringified JSON?

packages/content/src/hub.ts
```diff
@@ -6,6 +6,12 @@ import {
   hubApiRequest
 } from "@esri/hub-common";
 import { itemToContent, withPortalUrls } from "./portal";
+import { isSlug, addContextToSlug } from "./slugs";
+import { cloneObject } from "@esri/hub-common";
+
+export interface IGetContentOptions extends IHubRequestOptions {
+  siteOrgKey?: string;
```
tomwayson, line 13: @dbouwman I contemplated putting something like `site` on `IHubRequestOptions` as a place for the global site/domain data we store in appSettings in the Ember app b/c I suspect we'll need that stuff in a few places, but ended up just stabbing the one prop I need onto the local options for this fn.

packages/common/src/request.ts
mikeringrose: the hub api expects json, so this is stringifying a javascript object into json. in short this should be fine.

packages/content/src/slugs.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @returns true if the identifier is valid _and_ is **not** a record id
+ */
+export function isSlug(identifier: string): boolean {
+  const { itemId } = parseDatasetId(identifier);
+  if (!itemId || isGuid(itemId)) {
+    // it's either invalid, or an item id, or a dataset id
+    return false;
+  }
+  // otherwise assume it's a slug
+  return true;
```
mikeringrose, line 18: just return the condition `return (itemId && isGuid(itemId);`

tomwayson: ordinarily that's what I'd do, but here I wanted a place to hang the comments

packages/content/src/hub.ts
dbouwman: I'd be ok adding it as an optional on `IHubRequestOptions` but we def can't make it required as there are scenarios where this is no "site" context... and then there's the issue that sometimes the siteId is enough, others you'd really need the whole model, and in some env's you'd have it (i.e. Ember), others (i.e. scripting in node, or in some rando component) you may not... it's all messy messy with few clear "winner" solutions

tomwayson: yeah, that's sort of where I got to, and why I decided to just punt and add only exactly what I needed for this fn.

dbouwman (approved):
ALL YOUR CONTENT ARE BELONG TO HUB

Issue #305, opened 2020-08-03
things to remove at next breaking change


tomwayson:
## @esri/hub-content

https://github.com/Esri/hub.js/blob/54d33cd0251d31eb8f099827f530c288625d4ec8/packages/content/src/content.ts#L10

tomwayson:
In addition to above, search for "DEPRECATED" in the code base.

drewctate:
We should also separate the current "template" collection into something like "solution" and "template" or maybe something like "solutionTemplate" and "initiativeTemplate"

tomwayson:
@andrewctate does your last comment still stand?

andrewctate:
@tomwayson good question. I believe that I was just looking for a way to get solutions to send you the solution content view and initiative templates to send you to that content view. IIRC, I worked it out in the Ember app.

So, the question becomes: are solutions and initiative templates conceptually in the same collection even if we are partitioning that collection between two content views? I'm not sure. I'm not even really sure what the function of these groupings are.

Any insights?

tomwayson:
_backs away slowly_

andrewctate:
😂 Okay, then let's not touch it for now.

PR #357, opened 2020-09-03
update page func

drewctate:


packages/sites/src/pages/update-page.ts
```diff
@@ -23,9 +23,18 @@ import { removeUnusedResources } from "../layout/remove-unused-resources";
  */
 export function updatePage(
   model: IModel,
-  patchList: string[],
-  requestOptions: IHubRequestOptions
+  maybePatchList: string[] | Record<string, any>,
+  maybeRequestOptions?: IHubRequestOptions
 ): Promise<IUpdateItemResponse> {
+  let patchList = maybePatchList as string[];
```
tomwayson, line 29: add a comment above this: `// by default assume the deprecated signature was used`

packages/sites/src/pages/update-page.ts
```diff
@@ -23,9 +23,18 @@ import { removeUnusedResources } from "../layout/remove-unused-resources";
...
+  let requestOptions = maybeRequestOptions;
+
+  // support old call signature
```
tomwayson, line 32: this is the _new_ signature (where the second arg potentially _has_ a `.patchList` as opposed it _being_ a patch list).  It's really important to be clear in the comments here and in the names/docs of the args what will be supported going forward and what will be dropped at the next breaking change.

packages/sites/src/pages/update-page.ts
```diff
@@ -23,9 +23,18 @@ import { removeUnusedResources } from "../layout/remove-unused-resources";
...
+  if (!Array.isArray(maybePatchList)) {
+    patchList = getWithDefault(maybePatchList, "patchList", []) as string[];
```
tomwayson, line 34: the second arg is required so why not just use `maybePatchList.patchList || []` here?

dbouwman (approved):
LGTM

packages/sites/test/pages/update-page.test.ts
```diff
@@ -92,6 +92,45 @@ describe("updatePage", () => {
     );
   });
 
+  it("handles old call signature", async () => {
```
tomwayson, line 95: again, this looks to me like the _new_ signature b/c of `requestOptions.patchList`. if so this would be the test we'd want to keep. Is there another test for the deprecated signature (model, patchList, options)? If so _that_ one needs to be marked as such w/ explicit comments like `// this should be dropped when we bump major`

tomwayson (changes requested):
In addition the changes I suggest below, we'll need to add this to #305

tomwayson (approved):

PR #359, opened 2020-09-03
Bump rest.js versions to 2.17.0 for new sharing capabilities

rweber-esri:
[144791](https://esriarlington.tpondemand.com/entity/144791-search-filter-and-select-groups-to)

Bump rest.js versions to 2.17.0 for new sharing capabilities.

### DIFF
--- a/packages/sites/src/drafts/save-published-status.ts
+++ b/packages/sites/src/drafts/save-published-status.ts
@@ -2,6 +2,7 @@ import { IModel, IHubRequestOptions } from "@esri/hub-common";
 import { updateSite } from "../update-site";
 import { updatePage, isPage } from "../pages";
 import { isSite } from "../is-site";
+import { IUpdateItemResponse } from "@esri/arcgis-rest-portal";
 
 /**
  * Saves the published status of a site or page model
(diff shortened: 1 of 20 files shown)


dbouwman (approved):
This looks right to me... but human-yarn @tomwayson should double check the math

rweber-esri:
@tomwayson 

Only things that depend on _new_ functionality (i.e. I added removeUsers to REST.js) should have their `peerDependencies` bumped, which nothing depends on _new_ functionality, so `peerDepencendies` left as-is. Looks like `common`, `initiatives`, `sites` & `surveys` use REST.js `shareItemWithGroup` and `unshareItemWithGroup`, so bumped their `devDependencies` to latest REST.js `2.17.0`.

packages/sites/src/drafts/save-published-status.ts
```diff
@@ -2,6 +2,7 @@ import { IModel, IHubRequestOptions } from "@esri/hub-common";
 import { updateSite } from "../update-site";
 import { updatePage, isPage } from "../pages";
 import { isSite } from "../is-site";
+import { IUpdateItemResponse } from "@esri/arcgis-rest-portal";
```
tomwayson, line 5: if this interface is new in 2.17, we need to bump `@esri/arcgis-rest-portal` in the sites' `peerDependencies` as well as the `devDependencies`

tomwayson (changes requested):

packages/sites/src/drafts/save-published-status.ts
tomwayson: sorry, I didn't mean to block this - I meant to add that as a comment, not request for change.

And yes, you were on exactly the right track.

I'm going to approve this PR, it's good as is as long as `IUpdateItemResponse` was not added in 2.17

Issue #367, opened 2020-09-08
add `normalizedType` to content objects


drewctate:
This will make it easier for the content system in the app which relies upon a consistent item type.

Issue #366, opened 2020-09-08
Add type normalization for solution template items


drewctate:
Solution template items can be of type "Web Mapping Application" with a `typeKeyword` of `hubSolutionTemplate` instead of the standard `Solution` item type.

Gotta extend this guy: https://github.com/Esri/hub.js/blob/7019225322a28d7fc5224ed587c765d2ddb08758/packages/common/src/content.ts#L81

Also, probably need to separate the solution template and initiative template types into separate collections.

PR #368, opened 2020-09-08
feat(normalizeItemType): solution type normalization, normalizedType …

drewctate:
…in content

AFFECTS PACKAGES:
@esri/hub-common
@esri/hub-content

ISSUES CLOSED: #367, #366

### DIFF
--- a/packages/content/src/content.ts
+++ b/packages/content/src/content.ts
@@ -2,7 +2,7 @@
  * Apache-2.0 */
 
 import { request } from "@esri/arcgis-rest-request";
-import { IHubContent } from "@esri/hub-common";
+import { IHubContent, normalizeItemType } from "@esri/hub-common";
 import { IGetContentOptions, getContentFromHub } from "./hub";
 import { getContentFromPortal } from "./portal";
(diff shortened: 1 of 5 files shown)


benstoltz (approved):

packages/content/src/content.ts
```diff
@@ -29,6 +30,12 @@ export function getContent(
       return Promise.reject(e);
     });
   }
+  return ret.then(content => applyNormalizedType(content));
```
tomwayson, line 33: this should be done in `itemToContent()`, not here

tomwayson (commented):
please make the change I've requested before releasing

packages/content/src/content.ts
drewctate: Won't that make it so that it's only attached when we're using the portal API?

tomwayson: nope, `itemToContent()` is called by `datasetToContent()` as well

drewctate: Nevermind. I see

PR #369, opened 2020-09-08
fix(itemToContent): moved normalizedType generation to better spot

drewctate:
AFFECTS PACKAGES:
@esri/hub-common
@esri/hub-content

### DIFF
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -253,7 +253,7 @@ export interface IHubContent extends IHubResource, IItem {
    * The normalized item type (we run normalizeItemType on
    * the item in order to compute this prop)
    */
-  normalizedType?: string;
+  normalizedType: string;
   /**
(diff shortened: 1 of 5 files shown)


packages/common/src/types.ts
```diff
@@ -253,7 +253,7 @@ export interface IHubContent extends IHubResource, IItem {
    * The normalized item type (we run normalizeItemType on
    * the item in order to compute this prop)
    */
-  normalizedType?: string;
+  normalizedType: string;
```
tomwayson, line 256: this still needs to be optional

tomwayson (approved):

tomwayson (changes requested):
one small change

packages/common/src/types.ts
drewctate: Updated

tomwayson (approved):

Issue #353, opened 2020-09-02
IHubContent should have metadata, and provide a way to fetch it


tomwayson:
follow up from #330 and related to #322 (but more limited in scope).

As I indicated in https://github.com/Esri/hub.js/issues/322#issuecomment-685131319, I think `getContent()` should return `metadata` when it exists (i.e. when fetching from the Hub API). I'm not sure that it should proactively make an async request to get it otherwise. We can start by adding something like `getContentMetadata()` that would fetch it from wherever the enrichment currently fetches it from and the client can call that as needed.

drewctate:
@tomwayson This may be the key: https://www.arcgis.com/sharing/rest/content/items/d9ea4896eb7b4471b6839771c51de091/info/metadata/metadata.xml

drewctate:
Even better: https://www.arcgis.com/sharing/rest/content/items/1ce4060e854747038b153c5c8f8c894d/info/metadata/metadata.xml

ajturner:
This is a good plan. However, please be aware that the purpose of `getContent()` and other Hub.js functions is to provide a clear, consistent interface to the Hub models. it's not just a wrapper for promise chains :)

Specifically, in the near future item editors will be able to configure their metadata display. By far the most common request is to choose what shows in "Last Updated" because it's often not the "ArcGIS Item edited date" but instead some real-world date when a surveyor went into the field. Therefore, the [`Content.publishedDate`](https://github.com/Esri/hub.js/blob/fbfcf0ddb74a2dad215583163b44f530f0e38e03/packages/common/src/types.ts#L256) may be coming from `Content.publishedDateSource = 'metadata.dataIdInfo.idCitation.date.pubDate'` such as from [DC Trees](https://www.arcgis.com/sharing/rest/content/items/f6c3c04113944f23a7993f2e603abaf2/info/metadata/metadata.xml). 

So the `getContent()` call will look at it's display configuration (likely in either `item.properties` or another sidecar in resources) to know it should fetch this information and merge specific results into the core content model. @tomwayson @drewctate @mikeringrose

tomwayson:
[...]

PR #375, opened 2020-09-12
F/content metadata

tomwayson:
introduces `getContenMetadata()` that fetches item metadata and parses it as JSON.

@mikeringrose as mentioned in https://github.com/Esri/hub.js/issues/353#issuecomment-689826285, this uses same library (although a more recent version) and equivalent configuration as TheDuke does to parse the metadata as JSON. The major difference is that `getContentMetadata()` uses `@esri/arcgis-rest-portal` under the hood to fetch the metadata from portal.  The long term hope is that we could use this code in TheDuke (someday). Can you please look for anything that would prohibit that.


packages/content/src/metadata.ts
```diff
@@ -0,0 +1,28 @@
+import { getItemMetadata } from "@esri/arcgis-rest-portal";
+import { IHubRequestOptions } from "@esri/hub-common";
+import { parse } from "fast-xml-parser";
```
drewctate, line 3: Is this what we use in the indexer @mikeringrose ?

drewctate (approved):
LGTM

packages/content/src/metadata.ts
tomwayson: it is, though this PR is using v3.x, but indexer uses 2.x, and there are a few differences in the configuration settings and default behavior between the versions. I've made the equivalent configuration in this PR, but there are still slight differences in how the XML is parsed. For example, the [`peXml` node](https://github.com/Esri/hub.js/pull/375/files#diff-720828cca6aaaac8de88dda7278ca80fR35) is not parsed by v2 b/c it contains a nested XML document, but v3 parses that XML in place. See https://github.com/Esri/hub.js/pull/375/files#diff-720828cca6aaaac8de88dda7278ca80fR317-R324

mikeringrose: This is what we use. I wish we would have turned to it into nicer json, but it is what it is.

PR #377, opened 2020-09-15
F/get content from portal metadata

tomwayson:


packages/content/package.json
```diff
@@ -10,7 +10,7 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
-    "fast-xml-parser": "^3.17.4",
+    "fast-xml-parser": "~3.2.4",
```
tomwayson, line 13: this is the last version before they introduced `const` into their code.

packages/content/types/fast-xml-parser.d.ts
```diff
@@ -0,0 +1 @@
+declare module "fast-xml-parser";
```
tomwayson, line 1: this is needed b/c the older version of this library doesn't include types

dbouwman (approved):
Approved, but I think you need to add jsdoc to the `getContentFromPortal` function as it is exported

PR #378, opened 2020-09-16
[WIP] adds survey status utils to hub-surveys

cpgruber:


packages/common/src/types.ts
```diff
@@ -114,6 +114,21 @@ export interface IGetSurveyModelsResponse {
   stakeholder: IModel;
 }
 
+export interface IGetSurveyFormJson {
```
ajturner, line 117: I'm relatively new to typescript, but it seems odd to create interfaces based on methods and definitely not on encoding (JSON). Instead, this seems like there should be a model interface like `ISurveyForm`

packages/common/src/types.ts
```diff
@@ -114,6 +114,21 @@ export interface IGetSurveyModelsResponse {
...
+  settings: {
+    [propName: string]: any,
+    openStatusInfo: {
+      status: string,
+      schedule: IGetSurveySchedule
+    }
+  }
+}
+
+export interface IGetSurveySchedule {
```
ajturner, line 127: Similar to above, I think this should be ISurveySchedule

packages/common/src/types.ts
```diff
@@ -114,6 +114,21 @@ export interface IGetSurveyModelsResponse {
   stakeholder: IModel;
 }
 
+export interface IGetSurveyFormJson {
+  settings: {
+    [propName: string]: any,
+    openStatusInfo: {
+      status: string,
```
ajturner, line 121: can this be `status: 'scheduled' | 'closed'` or even a new enumeration like `status: SurveyStatus`

```
enum SurveyStatus {
  scheduled = "scheduled",
  closed = "closed"
}
```

packages/surveys/src/utils/status-utils.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @param {IGetSurveyFormJson} form Form.json item resource associated with survey item
+ * @param {Date} now Date object
+ * @returns {string}
+ */
+export function getSurveyStatus (
+  { settings: { openStatusInfo: { status: _status, schedule } } }: IGetSurveyFormJson,
+  now?: Date
+): string {
+  let status = _status;
+  if (status === 'scheduled') {
```
ajturner, line 14: per above suggest of a `SurveyStatus` enum this could then be `if( status === SurveyStatus.scheduled)` which is more typed and may help prevent string bugs.

packages/common/src/types.ts
ajturner: This would be fascinating to re-use [some of Survey123 types](https://github.com/ArcGIS/survey123-webform/blob/ea81d5503342bf36e06615cb0e61a13445ffd44f/libs/survey123-core/classes/form/form-settings/open-status-info.ts)

packages/common/src/types.ts
cpgruber: In opendata-ui, we make the distinction between a form item (item.type "Form") and its related form.json by typically calling them `formItem` and `formJson` respectively, so just carrying that over here

packages/common/src/types.ts
cpgruber: is ^ a private repo? cannot seem to access that link

drewctate: Can't get there either...

drewctate: @ajturner I think the enum is a reasonable idea, but I would actually vote for just a custom type in this case:

`type SurveyStatus = 'open'|'scheduled'|'closed';`

That provides the type checking without requiring the dot syntax (`SurveyStatus.something`) and I think enums are most useful when we'd like to give a more descriptive name to a non-intuitive set of values.

packages/common/src/types.ts
```diff
@@ -114,6 +114,21 @@ export interface IGetSurveyModelsResponse {
...
+      schedule: IGetSurveySchedule
+    }
+  }
+}
+
+export interface IGetSurveySchedule {
+  start: string | null,
```
drewctate, line 128: If you just want `start` to be optional, you can use `start?: string,`

tomwayson (changes requested):
Generally this looks good.

If the survey types are only used in the survey package, let's put them there, either in `src/types.ts` if they are shared amongst modules w/in that or directly in the modules that uses them if there's only one.

packages/common/src/types.ts
tomwayson: Also, using enums (or was it the _third_ way to skin this cat) [bit us in the ass in rest-js](https://github.com/Esri/arcgis-rest-js/pull/681). So until I understand how these things work I'm always going to use union types.

packages/surveys/src/utils/status-utils.ts
```diff
@@ -0,0 +1,67 @@
+import {
+  ISurveyFormJson,
+  ISurveySchedule,
+  ISurveyStatus,
+  IDetailedSurveyStatus
+} from "@esri/hub-types";
```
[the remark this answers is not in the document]
cpgruber: ok wasn't sure about where interfaces should go. @drewctate metioned `hub-types` as a landing spot for them. What's the rule of thumb for deciding where things go?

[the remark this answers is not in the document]
tomwayson: So

Rule of thumb should be as adjacent as possible to the fns that take/return the types.

Specifically:

1. I'm going to move hub-types into hub-common, so definitely don't use that package
1. If the types are shared amongst pacakges, use hub-common
1. Beyond that they should go in the module w/ the fns that take/return them, b/c likely they are only needed by those fns and/or the consumers of that fns
1. In rare cases we may have multiple modules w/in a package that need the same interface/type but one of the modules should not import from another (i.e. it would cause a circular dependency), in that case put the type in a `src/types.ts` in that package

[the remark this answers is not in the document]
cpgruber: immensely helpful, thanks @tomwayson

tomwayson (approved):

PR #382, opened 2020-09-21
survey results methods

cpgruber:
adds more methods for survey content views


packages/surveys/src/utils/results-availability/has-user-responded.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * check is provided user has any survey submissions
+ * @param {IStakeholderItem.url} _url feature service url
+ * @param {string} username the username to check for survey responses
+ * @param {IRequestOptions} requestOptions The request options
+ * @returns {Promise<boolean>}
+ */
+export const hasUserResponded = (
+  _url: IStakeholderItem["url"],
```
ajturner, line 13: I'm curious how this parameter works. It looks optional with type defined by the interface attribute type?

packages/surveys/src/utils/results-availability/migrate-form-properties-settings.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  getProp,
+  cloneObject,
+  IGetSurveyModelsResponse
+} from "@esri/hub-common";
+
+/**
+ * Populate form item properties is missing
```
ajturner, line 8: Is it valuable to provide more context for why this might occur? When would a dev use this method (or is it effectively 'internal').

packages/surveys/src/types.ts
```diff
@@ -0,0 +1,35 @@
+/* Copyright (c) 2020 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IItem } from "@esri/arcgis-rest-types";
+
+type IResultsAvailability = "after" | "always";
```
ajturner, line 7: I think we should start documenting types. This would help developers know the purpose and design intent of a particular type. Particularly since this is for `SurveyResults` - maybe a prefix would also be useful.

packages/surveys/src/utils/results-availability/has-user-responded.ts
```diff
@@ -0,0 +1,25 @@
+import { queryFeatures, IQueryResponse } from "@esri/arcgis-rest-feature-layer";
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IStakeholderItem } from "../../types";
+
+/**
+ * check is provided user has any survey submissions
```
ajturner, line 6: I think some more explanation of this method purpose would be helpful. It is likely useful for many future cases, but from the design & the tests (yay) it seems this is something like `Example: used to determine if resultsAvailability is set to 'after'`

packages/surveys/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface IFormItemProperties {
+  [propName: string]: any;
+  previewUrl?: string;
+  parentId?: string;
+  settings?: {
+    resultsAvailability: IResultsAvailability;
+  };
+}
+
+export interface IFormItem extends IItem {
```
ajturner, line 18: I'm a little concerned about our interchangeable use of `Survey` and `Form`. I understand that the AGO Item type is `Form`. However from Hub domain model we refer to `survey`. Would it make a better DevEx to use Survey consistently?

packages/surveys/src/utils/results-availability/has-user-responded.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param {string} username the username to check for survey responses
+ * @param {IRequestOptions} requestOptions The request options
+ * @returns {Promise<boolean>}
+ */
+export const hasUserResponded = (
+  _url: IStakeholderItem["url"],
+  username: string,
+  requestOptions: IRequestOptions
+): Promise<boolean> => {
+  const url = `${_url}/0`;
```
ajturner, line 17: is there a chance that the URL already includes the layer index? Should you check if it does before appending (otherwise it will get the first individual Feature, so still be a 200 OK)

packages/surveys/src/utils/results-availability/has-user-responded.ts
cpgruber: This is probably an overreach on my part learning TS. My guess is that it resolves to `_url: string`, since the definition of `IStakeholderItem > IItem.url` is `string`. I should probably just make it `string` for clarity

packages/surveys/src/utils/results-availability/has-user-responded.ts
cpgruber: Yes, there will be more values of `resultsAvailability` to come

packages/surveys/src/utils/results-availability/has-user-responded.ts
cpgruber: The url should come from a feature service item, which I believe does not include the layer index.

packages/surveys/src/types.ts
cpgruber: That's a fair point. I think (and @rweber-esri can correct me if I'm wrong) in `opendata-ui` we tend to refer to `survey` as the composite collection of related things (form item, form.json, views). Worth a discussion nevertheless.

packages/surveys/src/utils/results-availability/has-user-responded.ts
tomwayson: why is it `_url` instead of `url`?

packages/surveys/src/utils/results-availability/has-user-responded.ts
tomwayson: we do a lot of CYA in URL building in the rest of the library.

packages/surveys/src/utils/results-availability/migrate-form-properties-settings.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param {IGetSurveyModelsResponse} _models
+ * @returns {IGetSurveyModelsResponse}
+ */
+export const migrateFormPropertiesSettings = (
+  _models: IGetSurveyModelsResponse
```
tomwayson, line 13: I prefer to use `_` for the internal var, and no `_` for the external facing argument name.

packages/surveys/src/utils/results-availability/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./has-user-responded";
+export * from "./migrate-form-properties-settings";
+export * from "./should-display-results";
```
tomwayson, line 3: as I suggested in #384 - my preference would be to move these fns into this file rather than re-export them.

tomwayson (approved):
Nice work. Some minor suggestions.

packages/surveys/src/utils/results-availability/index.ts
cpgruber: that's fine, but having them individual makes them stub-able in tests

packages/surveys/src/types.ts
```diff
@@ -0,0 +1,35 @@
+/* Copyright (c) 2020 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IItem } from "@esri/arcgis-rest-types";
+
+type IResultsAvailability = "after" | "always";
+
+export interface IFormItemProperties {
+  [propName: string]: any;
```
rweber-esri, line 10: Minor, concrete interface properties are generally declared  _before_ generic catch-alls like `[propName: string]: any`, similar to how we'd use the `rest` operator at the end of object or array destructuring.

packages/surveys/src/utils/results-availability/has-user-responded.ts
cpgruber: > why is it `_url` instead of `url`?

just so i can use `url` in the function itself

packages/surveys/src/types.ts
rweber-esri: @cpgruber that's correct. There's a few places where we interact with more than just the `Form` item, and I've been using `Survey` as a higher level aggregate type comprised of the different survey item models.

packages/surveys/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  previewUrl?: string;
+  parentId?: string;
+  settings?: {
+    resultsAvailability: IResultsAvailability;
+  };
+}
+
+export interface IFormItem extends IItem {
+  id: string;
```
rweber-esri, line 19: We get this prop from [IItem](https://github.com/Esri/arcgis-rest-js/blob/v2.13.0/packages/arcgis-rest-types/src/item.ts#L57)

packages/surveys/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  type: "Form";
+  properties?: IFormItemProperties;
+}
+
+type IStakeholderItemTypeKeywords =
```
rweber-esri, line 24: Not sure there's a lot of value in this type since any `string` will do. Could just define `typeKeywords: string[]` on `IStakeholderItem`.

cpgruber: Yea, i was hoping to find a typescript-y way to enforce that `IStakeholderItem.typeKeywords` must include those listed, in addition to any others

rweber-esri (approved):

PR #384, opened 2020-09-22
feat(can-edit): add methods to check edit access to various hub entities

cpgruber:
affects: @esri/hub-common


packages/common/src/access/can-edit-dataset.ts
```diff
@@ -0,0 +1,13 @@
+import { IItem, IUser } from "@esri/arcgis-rest-types";
+import { hasBasePriv } from "./has-base-priv";
+import { canEditItem } from "./can-edit-item";
+
+/**
+ *
```
ajturner, line 6: Please add a description for this method. In particular what it means to check for editing.

packages/common/src/access/can-edit-event.ts
```diff
@@ -0,0 +1,29 @@
...
+import { findBy } from "../util";
+import { hasBasePriv } from "./has-base-priv";
+
+export interface IEventModel {
+  [propName: string]: any;
+  initiative: IInitiativeModel;
+}
+
+/**
+ *
```
ajturner, line 13: Please add description for `canEditEvent`

packages/common/src/access/can-edit-dataset.ts
```diff
@@ -0,0 +1,13 @@
...
+ * @param {IItem} model
+ * @param {IUser} currentUser
```
ajturner, line 8: should this be `username` for both consistency and also the method can be used to check if another user (not current) is able to edit a dataset. For example a table could show all members of a group and an indication if that user can edit this dataset.

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,46 @@
+import { IItem, IUser, IGroup } from "@esri/arcgis-rest-types";
+import { isUpdateGroup, includes } from "../utils";
+import { getProp } from "../objects";
+import { hasBasePriv } from "./has-base-priv";
+
+/**
+ *
```
ajturner, line 7: Please add description for `canEditItem`

packages/common/src/access/can-edit-site.ts
```diff
@@ -0,0 +1,19 @@
+import { IItem, IUser } from "@esri/arcgis-rest-types";
+import { getProp } from "../objects";
+import { hasBasePriv } from "./has-base-priv";
+import { canEditItem } from "./can-edit-item";
+
+/**
+ *
```
ajturner, line 7: Please add description. And particularly that this has special logic for "default Hub Home". In the near future we'll be adding "hub home" capability and it should be called out that this method will likely need to be updated.

ajturner (changes requested):
Nice job moving these into hub.js. 
There aren't description on methods which would be helpful (necessary) for Dev docs. They should also provide relevant details on caveats (e.g. if default hub home)

packages/common/src/access/can-edit-dataset.ts
cpgruber: that's true, it should be `user`

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param {IItem} model
+ * @param {IUser} currentUser
+ * @returns {boolean}
+ */
+export function canEditItem(model: IItem, currentUser?: IUser): boolean {
```
tomwayson, line 12:
```suggestion
export function canEditItem(item: IItem, currentUser?: IUser): boolean {
```

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,46 @@
...
+  let res = false;
+  if (currentUser) {
+    const itemControls = ["admin", "update"];
+    const { itemControl, owner, orgId: itemOrgId } = model;
+    let { isOrgItem } = model;
```
tomwayson, line 17: is `isOrgItem` a thing we stab on items sometimes? If so, and this is truly needed, should pass this in via an options argument instead of relying on consumers stabbing an undoc'd prop onto the item?

packages/common/src/access/can-edit-site-content.ts
```diff
@@ -0,0 +1,37 @@
...
+  "portal:user:viewOrgItems"
+];
+
+/**
+ *
+ * @param {IItem} model
+ * @param {IUser} currentUser
+ * @returns {boolean}
+ */
+export function canEditSiteContent(model: IItem, currentUser?: IUser): boolean {
```
tomwayson, line 21: let's use `item` here too.

tomwayson (commented):
This is great work. Other than agreeing w/ @ajturner's feedback and my one comment re: `isOrgItem` I'd say generally we should use `item` instead of `model` as the argument name to these fns. Also, I'm wondering if `canEditItem()` and `hasBasePriv()` ultimately belong in rest-js. One way to defer that decision would be to move all the fns into a single module (i.e. `access` or `can-edit`) and not export those fns. I'm also suggesting that b/c I prefer that style over one fn per file (although yes, I realize that most of the code in this repo is one fn per file, but I would like to change that and be more intentional about our modules - code style discussion to follow).

packages/common/src/access/can-edit-item.ts
cpgruber: `isOrgItem` comes from portal API, but i could see it being redundant to just checking item.orgId against user.orgId

cpgruber:
can someone spot check `package-lock.json`. looks like an aggressive diff there

packages/common/src/access/can-edit-dataset.ts
```diff
@@ -0,0 +1,13 @@
...
+import { hasBasePriv } from "./has-base-priv";
+import { canEditItem } from "./can-edit-item";
+
+/**
+ * Checks if user has access to edit dataset in Hub
+ * @param {IItem} model
+ * @param {IUser} user
+ * @returns {boolean}
+ */
+export function canEditDataset(model: IItem, user?: IUser): boolean {
```
mikeringrose, line 11: is user really optional?

cpgruber: it is in the sense that, if you don't provide user you get `false`

packages/common/src/access/can-edit-event.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param {IUser} user
+ * @returns {boolean}
+ */
+export function canEditEvent(model: IEventModel, user?: IUser): boolean {
+  let res = false;
+  if (hasBasePriv(user)) {
+    const coreTeamId = getProp(
+      model,
+      "initiative.item.properties.collaborationGroupId"
+    );
```
mikeringrose, line 24: not sure what is clearer, but i prefer optional chaining -- `model?.initiative?.item?.properties?. collaborationGroupId

packages/common/src/access/can-edit-dataset.ts
cpgruber: imo, it's better handled here as an optional arg than in the ember app. `session.currentUser` can be undefined, so by making it optional we don't have to gate these methods there

packages/common/src/access/can-edit-event.ts
```diff
@@ -0,0 +1,29 @@
...
+    const groups = user.groups || [];
```
mikeringrose, line 25: could just do `const { groups = [] } = user`.

packages/common/src/access/can-edit-dataset.ts
```diff
@@ -0,0 +1,13 @@
...
+  return hasBasePriv(user) && canEditItem(model, user);
```
mikeringrose, line 12: do you need to call `hasBasePriv` here? it's called in `canEditItem`?

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,43 @@
...
+import { getProp } from "../objects";
+import { hasBasePriv } from "./has-base-priv";
+
+/**
+ * Checks if user has access to edit an item in Hub
+ * @param {IItem} item
+ * @param {IUser} user
+ * @returns {boolean}
+ */
+export function canEditItem(item: IItem, user?: IUser): boolean {
```
mikeringrose, line 12: Why is user optional?

packages/common/src/access/can-edit-event.ts
tomwayson: me too, someone just needs to verify that our build pipeline will properly transpile and/or polyfill them for our target enviornments.

packages/common/src/access/can-edit-dataset.ts
mikeringrose: i'd argue the client shouldn't call this method if the currentUser is not available.

packages/common/src/access/can-edit-event.ts
mikeringrose: looking at our tsconfig, it should be but i haven't verified it. https://www.typescriptlang.org/play?target=2#code/MYewdgzgLgBCBGArGBeGBvGBLAJgLgxngLAFcAbcmAXxoG4AoUSEcgUwDpyQBzACgSIA-B1wj4ASkZA

cpgruber: FYI codecov considers each optional bit as a codepath that requires test coverage

packages/common/src/access/can-edit-dataset.ts
cpgruber: I'm nuking this in favor of calling `canEditItem` for datasets, until a different more specific check is needed

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param {IUser} user
+ * @returns {boolean}
+ */
+export function canEditItem(item: IItem, user: IUser): boolean {
+  let res = false;
+  const itemControls = ["admin", "update"];
+  const { itemControl, owner, orgId: itemOrgId } = item;
+  const { roleId, role, username, groups: userGroups, orgId: userOrgId } = user;
+  const hasItemControl = includes(itemControls, itemControl);
+  const isOwner = !!owner && !!username && owner === username;
```
rweber-esri, line 18: A couple of these assignments could be simplified.

```
const isOwner = !!owner && owner === username;
const isOrgItem = !!itemOrgId && itemOrgId === userOrgId;
```

packages/common/src/access/can-edit-site-content.ts
```diff
@@ -0,0 +1,39 @@
...
+export function canEditSiteContent(item: IItem, user: IUser): boolean {
+  let res = false;
+  const isDefaultHubHome = getProp(item, "properties.isDefaultHubHome");
+  const hasPriv = hasBasePriv(user);
+  if (!isDefaultHubHome && hasPriv) {
+    res = canEditItem(item, user);
+  } else if (hasPriv) {
+    const userOrgId = user.orgId;
+    const itemOrgId = item.orgId;
+    const sameOrg = !!userOrgId && !!itemOrgId && userOrgId === itemOrgId;
```
rweber-esri, line 32: Similar, could drop `&& !!ItemOrgId`

rweber-esri (approved):
Nice job. Added a couple optional comments, approving since other feedback was addressed.

packages/common/src/access/can-edit-item.ts
```diff
@@ -0,0 +1,35 @@
...
+  const isOrgItem = !!itemOrgId && !!userOrgId && itemOrgId === userOrgId;
+  const isItemOrgAdmin = !!isOrgItem && !roleId && role === "org_admin";
```
rweber-esri, line 20: `isItemOrgAdmin` might be deserving of it's own method.

packages/common/src/access/can-edit-site.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Checks if user has access to edit site in Hub
+ * Currently, Hub Home sites are not editable
+ * In initiative context, check is delegated to canEditItem for the initiative site item
+ * @param {IItem} model
+ * @param {IUser} user
+ * @returns {boolean}
+ */
+export function canEditSite(model: IItem, user: IUser): boolean {
```
drewctate, line 14: Should probably be in `hub-sites`, no?

drewctate (commented):
LGTM! Should we move `canEditSite` into the `hub-sites` package? Should be easy to do consider that hub-common is a dependency.

drewctate (approved):

Issue #385, opened 2020-09-28
parse `item.categories` for private data to match indexer


drewctate:
When we get an item from the Portal API, we need to follow similar logic to parse `item.categories` entries. For example
```
["/Categories/Boundaries", "/Categories/Planning and cadastre/Property records", "/Categories/Structure"]
```
Should end up being
```
["Boundaries", "Planning and cadastre", "Property records", "Structure"]
```

https://github.com/ArcGIS/hub-indexer/blob/438ed91901fa2772ed1c8a03748ed9054341bc25/packages/duke/enrich/categories.js#L1-L20

PR #387, opened 2020-09-28
parsing item categories to match indexer logic

drewctate:
ISSUES CLOSED: #385

### DIFF
--- a/packages/content/src/portal.ts
+++ b/packages/content/src/portal.ts
@@ -82,6 +82,7 @@ export function itemToContent(item: IItem): IHubContent {
     hubType: getItemHubType(item),
     normalizedType: normalizeItemType(item),
     categories: parseItemCategories(item.categories),
+    itemCategories: item.categories,
     // can we strip HTML from description, and do we need to trim it to a X chars?
     summary: item.snippet || item.description,
     publisher: {
--- a/packages/content/src/portal.ts
+++ b/packages/content/src/portal.ts
@@ -15,7 +15,8 @@ import {
   getItemApiUrl,
   getItemDataUrl,
   getItemThumbnailUrl,
-  cloneObject
+  cloneObject,
+  includes
 } from "@esri/hub-common";
 import { getContentMetadata } from "./metadata";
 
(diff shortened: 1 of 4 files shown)


packages/content/src/portal.ts
```diff
@@ -80,6 +81,7 @@ export function itemToContent(item: IItem): IHubContent {
     hubId: item.id,
     hubType: getItemHubType(item),
     normalizedType: normalizeItemType(item),
+    categories: parseItemCategories(item.categories),
```
tomwayson, line 84: Let's not overwrite the item's own categories, but instead add a new property like 

```suggestion
    flattenedCategories: parseItemCategories(item.categories),
```

Open to suggestions on the name. Whatever we decide, let's add it as an optional (w/ `?`) property on `IHubCategory`.

tomwayson (commented):
Nice work, just need to put it under a property w/ a different name.

packages/content/src/portal.ts
drewctate: I thought about this, too. The reason I overwrote categories is that that's what the indexer does. Will assigning it to a new property name maintain parity between content coming from the Hub and Portal APIs?

tomwayson: Does the indexer preserve the item's categories under any other field (like `itemCategories`)?

tomwayson: If so, let's duplicate that behavior here.

If not, let's do that here on the client and then open an issue in the indexer repo to talk about doing the same there.

drewctate: Okay, it looks like it doesn't. `flattenedCategories` sounds good. I'll open an issue on the indexer.

tomwayson: well in this case we would not be adding `flattenedCategories`, we'd be adding `itemCategories = item.categories` in order to not create a breaking change.

drewctate: I see. Okay. Worth noting that unfortunately, `itemCategories` will only be genuine when we're using the Portal API. For content from the Hub API, the two attributes will be identical.

tomwayson (changes requested):
One more small fix.

tomwayson (approved):
Well done.

PR #389, opened 2020-09-30
fix(site and page drafts): deleting all draft resources when a new on…

drewctate:
…e is saved and using the most r

affects: @esri/hub-sites

deleting all draft resources when a new one is saved and using the most recent one if there are
multiple


dbouwman (approved):
looks good!

PR #390, opened 2020-10-01
added collections map

drspacemanphd:
affects: @esri/hub-common

Adds a collections map of categories currently listed in Hub.js. This is helpful for checking if a category in opendata-ui is associated with **any** collection and helps reduce the possibility of a new category causing translation issues.

### DIFF
--- a/packages/common/src/collections.ts
+++ b/packages/common/src/collections.ts
@@ -131,3 +131,14 @@ export const collections: { [key: string]: string[] } = {
   other,
   site
 };
+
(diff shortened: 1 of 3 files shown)


drewctate (approved):
Nice work. But, remember to hold off on merging until we're able to release `6.12.0` (@tomwayson is waiting on NPM support).

tomwayson:
As we discussed, this will not be needed, the app can just call `getCollection()`

Issue #355, opened 2020-09-02
Search collections impacting content categorization and UX


brollison:
It has come to my attention this search [collections](https://github.com/Esri/hub.js/blob/master/packages/common/src/collections.ts) file, which feeds into this [categories](https://github.com/Esri/hub.js/blob/master/packages/common/src/categories.ts) file, is now impacting...
1. What content is supported on what content views
1. Custom dimensions in telemetry insights
1. How content may be referred to within the application

I'll break these out into separate points to illustrate the disconnect.

#### 1. Supported Content on Views
It was recently discovered that `Microsoft Excel` documents are now being classified as a `dataset`, thus when attempting to visit `/documents/:id` ([example](https://brollywood.hikes.dev/documents/3b6e18c65cc44de7afb11d03cd2e3631)) for a `type: Microsoft Excel` item the user is redirected to `/data/:id/explore` ([result](https://brollywood.hikes.dev/data/qa-pre-a-hub::microsoft-excel/explore)).  While on the surface an Excel document may seem similar to a `type: CSV` item, the data cannot be extracted (at least to my knowledge) and visualized on a table; thus, as a team we [defined](https://esriis.sharepoint.com/:x:/s/ArcGISHub/EdzJoa1sT-RNksZpPvlXHPoBQYRnWRAtL_bEK02AEgYzDw?e=jdAOHN) Excel items as `documents`.

In fact, quite a few of the items listed in the [collection](https://github.com/Esri/hub.js/blob/master/packages/common/src/collections.ts) file do not match the collaborative classification we defined.  A few other examples...
* `type: CityEngine Web Scene` being classified as a `map` when in reality all parties chose `apps` (if embeddable, which it is)
* `type: Document Link` as a `document` where we chose `content` (generic fallback) after quite a bit back-and-forth

I'm generally concerned about forcing a 1-1 match between search collections an in-app categorization of content.  Should/could categorization feed into collections?  Sure, but it likely isn't always desired.

#### 2. Impact on Telemetry
We have a product priority to better understand how and what content is being interacted with within the application.  The purposes of these insights range from functionality efficacy / roadmap to customer-facing value of "how are xyz doing on my site?".  The telemetry.js model allow us to specify 3 global fields...
```json
category: Engagement
action: Favorite
label: Content-Hero
```
...however, we can only get more granular details via custom dimensions.  We are targeting the following...
```json
id: <item.id>
type: StoryMap
content: Apps
```
[...]

PR #392, opened 2020-10-01
F/collection updates

tomwayson:
Addresses https://github.com/Esri/hub.js/issues/355#issuecomment-697946638 and https://github.com/Esri/hub.js/issues/355#issuecomment-700923580.

@drspacemanphd note that f06bff0 will change the behavior of any search code that calls `encodeAgoQuery()`.


drewctate (commented):
So... these types just won't come back under the map facet on `/search`? If that's the case, I don't really have an opinion.

tomwayson:
Well I would guess that search results in the app would not be affected b/c we group maps and apps together: 

https://hub.arcgis.com/search?collection=App%2CMap

But that's why I want @drspacemanphd to vet this.

drewctate:
Good point @tomwayson .

drspacemanphd:
This makes sense to me, given that App & Map are queried together as part of both Search and Content Library. I've approved

drspacemanphd (approved):

PR #396, opened 2020-10-08
Domains

robertsteilberg:
* changes domain service urls to `api/v3/domains`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ packages/*/debug/
 
 ## for john to poke at
 test.html
+
+.idea/
+.vscode/
--- a/packages/sites/src/domains/add-domain.ts
+++ b/packages/sites/src/domains/add-domain.ts
@@ -20,7 +20,7 @@ export function addDomain(
   }
   const headers = _getAuthHeader(hubRequestOptions);
   headers["Content-Type"] = "application/json";
-  const url = `${_getDomainServiceUrl(hubRequestOptions)}`;
+  const url = `${_getDomainServiceUrl(hubRequestOptions.hubApiUrl)}`;
   return fetch(url, {
     method: "POST",
     headers,
--- a/packages/sites/src/domains/add-domain.ts
+++ b/packages/sites/src/domains/add-domain.ts
@@ -20,7 +20,7 @@ export function addDomain(
   }
   const headers = _getAuthHeader(hubRequestOptions);
   headers["Content-Type"] = "application/json";
-  const url = `${_getDomainServiceUrl(hubRequestOptions)}?force=true`; // Use the force...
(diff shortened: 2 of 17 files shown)


drewctate:
Has this been tested on Portal @robertsteilberg?

robertsteilberg:
No, but Portal does not use the domain service or ever encounter the domain service, so I do not think it will make a difference

drewctate (approved):
I think you're probably safe. As you say, these utils shouldn't even be fired on Portal. Nice work.

packages/sites/src/get-domain.ts
```diff
@@ -15,7 +15,7 @@ export function getDomains(
   requestOptions?: IRequestOptions
 ): Promise<any> {
   const apiUrl = getHubApiUrl(requestOptions);
-  const url = `${apiUrl}/utilities/domains`;
+  const url = `${apiUrl}/api/v3/domains`;
```
tomwayson, line 18: why not use `_getDomainServiceUrl()` here/

packages/sites/src/domains/add-domain.ts
```diff
@@ -20,7 +20,7 @@ export function addDomain(
   }
   const headers = _getAuthHeader(hubRequestOptions);
   headers["Content-Type"] = "application/json";
-  const url = `${_getDomainServiceUrl(hubRequestOptions)}?force=true`; // Use the force...
+  const url = `${_getDomainServiceUrl(hubRequestOptions)}`;
```
drewctate, line 23: Just curious. Why no `force=true`?

.gitignore
```diff
@@ -84,3 +84,6 @@ packages/*/debug/
 
 ## for john to poke at
 test.html
+
+.idea/
+.vscode/
```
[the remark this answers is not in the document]
robertsteilberg: ;)

packages/sites/src/get-domain.ts
robertsteilberg: good point i'll refactor that

packages/sites/src/domains/add-domain.ts
robertsteilberg: this flag never worked in the original domain service, so we didn't implement it in the new one

packages/sites/src/get-domain.ts
robertsteilberg: looks like a type mismatch--`_getDomainServiceUrl` expects `IHubRequestOptions` which we don't have in `getDomains`

tomwayson: you can change that to accept `hubRequestOptions: IHubRequestOptions | hubApiUrl: string`

tomwayson (approved):
thanks @robertsteilberg

Issue #395, opened 2020-10-08
Unable to comment, commit, or push from Bash shell on Windows


MikeTschudi:
Using MINGW64, commands such as `npm run c` and `"C:\Program Files\Git\bin\git.exe" commit -F "C:\...\hub.js\.git\COMMITMESSAGE"` fail with

```
$ npm run c

> @esri/hub.js@6.11.2 c C:\...\hub.js
> npm run precommit && git-cz


> @esri/hub.js@6.11.2 precommit C:\...\hub.js
> ./exit_if_no_staged.sh && lint-staged

'.' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @esri/hub.js@6.11.2 precommit: `./exit_if_no_staged.sh && lint-staged`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @esri/hub.js@6.11.2 precommit script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\...\AppData\Roaming\npm-cache\_logs\2020-10-08T17_16_28_173Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @esri/hub.js@6.11.2 c: `npm run precommit && git-cz`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @esri/hub.js@6.11.2 c script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\...\AppData\Roaming\npm-cache\_logs\2020-10-08T17_16_28_207Z-debug.log
```

drewctate:
Sure. Or maybe I'll just convert it to be a node script.

tomwayson:
@MikeTschudi next time you pull master could you please confirm that this has been fixed?

andrewctate:
@tomwayson I think we can safely close this.

MikeTschudi:
Haven't had a commit to test with, but `npm run c` doesn't fail as shown above; it now fails by not allowing the arrow key to select the type of commit.

PR #406, opened 2020-10-18
fix(Precommit hook): uses node instead of bash to check if staged changes

drewctate:
addresses #395


tomwayson (approved):
thanks @drewctate!

PR #407, opened 2020-10-18
Removing reference "Hub Solution Template" item type

drewctate:
Little did I know... it doesn't actually exist.


tomwayson (approved):

packages/common/src/collections.ts
```diff
@@ -54,7 +54,9 @@ const feedback: string[] = ["Form"];
 
 const initiative: string[] = ["Hub Initiative"];
 
-const template: string[] = ["Hub Initiative Template", "Hub Solution Template"];
+const solution: string[] = ["Solution"];
```
drewctate, line 57: Based on conversation with @ghudgins and @jdev42092 , we'll be assuming that `Solutions` are one type and handling `Solution` templates on the app side.

tomwayson (approved):

PR #411, opened 2020-10-22
fix(getItemThumbnailUrl): pass token when calling from withPortalUrls…

tomwayson:
…(); drop support for options.wi


packages/common/src/urls/get-item-api-url.ts
```diff
@@ -20,6 +20,7 @@ export const getItemApiUrl = (
   const { id, access } = item;
   const url = `${getPortalApiUrl(portalUrlOrObject)}/content/items/${id}`;
   const params = new URLSearchParams({ f: "json" });
+  // TODO: derive token from from requestOptions if passed in?
```
drewctate, line 23: Good idea.

drewctate (commented):
Nice

PR #412, opened 2020-10-22
feat(hub-sites): adds socialSharing as default capability for non-ent…

cpgruber:
…erprise sites

affects: @esri/hub-sites

sites created from embedded templates in non-portal environments will have "socialSharing" included
by default

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -554,71 +554,6 @@
         }
       }
     },
-    "@esri/arcgis-rest-auth": {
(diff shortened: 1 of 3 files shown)


cpgruber:
pretty gnarly diff in package-lock, so unless that raises a flag to anyone i'll blame that on computers doing computer things

drewctate (approved):
Nice work.

Issue #397, opened 2020-10-08
add searchUsers() function


juliannemarik:
## Background/Motivation
We recently released a new People Card which allows our customers to prominently feature users on their site. This feature is available on Hub, Hub Premium & Enterprise licenses. In the process of completing this feature, we discovered some of the strengths and limitations of the AGO Rest API with respect to the [Users Search](https://developers.arcgis.com/rest/users-groups-and-items/user-search.htm) and [Community User](https://developers.arcgis.com/rest/users-groups-and-items/user.htm) endpoints, namely: 

> NOTE: user profiles can have 1 of 3 types of visibility - (1) Private (2) Organization (3) Public

### Users Search
- You can pass a query string with the usernames of the people you want returned. This means you only need to hit the endpoint _one_ time with a query string like: 
```
"q=username:username1 OR username:username2 OR username:username3..."
```
- Hitting the endpoint as described above handles profile visibility, meaning that AGO will only return the subset of people which the current user has access to. 
- The endpoint is only available to _authenticated_ users. Since unauthenticated users should still be able to see public profiles, hitting this endpoint is insufficient by itself

### Community Users
- This endpoint is used to individually fetch users by their username. This means that a call needs to be made for _each_ user.
- This endpoint, unlike the Users Search endpoint which only returns the subset of people that the current user has access to, will only _limit_ the information returned for each user (i.e. `firstname`, `lastname`, and `fullname` will be empty strings if an unauthenticated users tries to access a non-public profile). 
- This endpoint is available to unauthenticated users

Because of these strengths/limitations, we ended up conditionally hitting the appropriate endpoint depending on whether the user is authenticated. 

---
In addition to the implementation above, there's an ongoing effort to add a hub-specific User Search to the Hub API. Ideally once complete, we'd just switch the above implementation over to using this hub-specific implementation; however, because the Hub API is not available in Enterprise, we'd still need to use the above implementation as a fallback. 

## Proposal
Encapsulate the above logic in a function, `getUsers()`, within `Hub.js`.

### Pseudo-Code
```
getUsers(usernames) {
  let users;

  if(online) {
    users = use Hub API User Search (whenever it becomes available)
  } else if (enterprise) {
    if(user is authenticated) {
      users = use AGO User Search endpoint and pass in a query string with the usernames 
    } else {
      users = use AGO Community User endpoint to fetch one user at a time by username and filter users by profile visibility
    }
  }

  return users
}
```

brollison:
[...]

PR #413, opened 2020-10-23
feat(hub-sites): add getMembers function

juliannemarik:
affects: @esri/hub-sites

add new getMembers function to allow authenticated and unauthenticated users to fetch a list of
members by username

[Related Issue](https://github.com/Esri/hub.js/issues/397)

future work:
- once the Hub API User Search is complete, we should integrate it into this function for AGO users 
and use the current implementation as a fallback for enterprise users
- extend or create a new function that supports full search of members, not just fetching by username

### DIFF
--- a/packages/sites/src/get-members.ts
+++ b/packages/sites/src/get-members.ts
@@ -64,7 +64,7 @@ function authenticatedGetMembers(
     };
   });
 
-  return batch(chunkedOptions, batchRequest).then(batchedMembers => {
+  return batch(chunkedOptions, batchMemberRequest).then(batchedMembers => {
     return batchedMembers.reduce((flat: IUser[], toFlatten: IUser[]) => {
       return flat.concat(toFlatten);
     }, []);
@@ -107,7 +107,7 @@ function unauthenticatedGetMembers(
   ).then(members => members.filter(Boolean));
 }
 
-interface IBatchRequestOptions {
+interface IBatchMemberRequestOptions {
   urlPath: "string";
   requestOptions: IHubRequestOptions;
 }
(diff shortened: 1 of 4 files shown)


packages/sites/src/get-members.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * callback function to batch the requests to the user search endpoint
+ * if >100 usernames are supplied. This is necessary because the
+ * endpoint sets the maximum number of results to be included in the
+ * result set response to 100
+ *
+ * @param options IBatchRequestOptions
+ */
+function batchRequest(options: IBatchRequestOptions): Promise<IUser[][]> {
```
rweber-esri, line 123: `batchRequest` seems generic, but it's return type is specific. I'd consider either renaming to something more specific like `batchMemberRequest` and `IBatchMemberRequestOptions`, or consider making the return signature more generic. If you take the latter route, it'll require you to do some type casting where you consume this method.

packages/sites/src/get-members.ts
```diff
@@ -0,0 +1,134 @@
...
+ * an unauthenticated user tries to access a non-public profile).
+ *
+ * @param {Array} usernames List of usernames to search for
+ * @param {object} requestOptions IHubRequestOptions
+ */
+function unauthenticatedGetMembers(
+  usernames: string[],
+  requestOptions: IHubRequestOptions
+): Promise<IUser[]> {
+  return Promise.all(
```
tomwayson, line 89: I suppose there's some upper limit to the number of requests we can throw at `Promise.all()`, but I guess we can cross that bridge when we come to it.

tomwayson (approved):
Great work @juliannemarik!

PR #418, opened 2020-11-03
f/content-is-downloadable

mjuniper:


packages/content/test/portal.test.ts
```diff
@@ -65,6 +65,7 @@ function validateContentFromPortal(content: IHubContent, item: IItem) {
   expect(typeof content.portalApiUrl).toBe("string");
   expect(typeof content.portalDataUrl).toBe("string");
   expect(typeof content.thumbnailUrl).toBe("string");
+  expect(content.isDownloadable).toBe(true);
```
ajturner, line 68: Post-facto, but should there be more tests, at least for the two expected conditions?

PR #419, opened 2020-11-06
feat(fetchAllPages): adds a function to fetch all pages of a search i…

drewctate:
…n batches

affects: @esri/hub-common

### DIFF
--- a/packages/common/src/items/fetch-all-pages.ts
+++ b/packages/common/src/items/fetch-all-pages.ts
@@ -5,8 +5,11 @@ import { batch } from "../utils/batch";
 const MAX_NUM = 100;
 
 /**
- * Fetches all the pages
- * @param opts search options
+ * Fetches all the pages in a search request
+ * @param {SearchFunction} searchFunc
+ * @param {ISearchOptions} opts
+ * @param {number} limit
+ * @returns {Promise<SearchableType[]>}
  */
 export function fetchAllPages(
   searchFunc: SearchFunction,
--- a/packages/common/src/items/fetch-all-pages.ts
+++ b/packages/common/src/items/fetch-all-pages.ts
@@ -1,0 +1,33 @@
+import { ISearchOptions, ISearchResult } from "@esri/arcgis-rest-portal";
+import { SearchableType, SearchFunction } from "../types";
+import { batch } from "../utils/batch";
+
+const MAX_NUM = 100;
+
+/**
+ * Fetches all the pages
+ * @param opts search options
+ */
+export function fetchAllPages(
+  searchFunc: SearchFunction,
+  opts: ISearchOptions,
+  pageSize = MAX_NUM
+): Promise<SearchableType[]> {
+  return searchFunc({ ...opts, num: 1 })
+    .then(({ total }) => {
+      const starts = [];
+      for (let i = 1; i <= total; i += pageSize) starts.push(i);
+      const batchSearchFunc = (start: number) =>
+        searchFunc({ ...opts, start, num: pageSize });
+      return batch(starts, batchSearchFunc);
+    })
+    .then(responses => {
+      return responses.reduce(
+        (acc: SearchableType[], response: ISearchResult<SearchableType>) => [
+          ...acc,
+          ...response.results
+        ],
+        []
+      );
+    });
+}
--- a/packages/common/src/items/fetch-all-pages.ts
+++ b/packages/common/src/items/fetch-all-pages.ts
@@ -28,6 +28,7 @@ export function fetchAllPages(
 
   return promise
     .then(firstResponse => {
+      // no more requests needed, return the first response
       if (firstResponse.nextStart === -1) return [firstResponse];
       const starts = [];
       for (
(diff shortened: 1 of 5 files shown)


packages/common/src/items/fetch-all-pages.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Fetches all the pages
+ * @param opts search options
+ */
+export function fetchAllPages(
+  searchFunc: SearchFunction,
+  opts: ISearchOptions,
+  pageSize = MAX_NUM
+): Promise<SearchableType[]> {
+  return searchFunc({ ...opts, num: 1 })
```
drewctate, line 16: This call is just to get the total and should be pretty quick. However, I toyed with other ideas:
- Just use the first slice to get the `total` and do things batch-wise after that. Only drawback is that the batch routine is delayed for the duration of the first request/response (which could be quite sizable for a large query).
- Allow a `limit` parameter to be passed to short-circuit this. But, that brings in the question of what should the limit be, etc.

So, I'm open to input.

tomwayson: This is fine as is, but I also I like the idea of using "the first slice to get the total and do things batch-wise after that."

packages/common/src/types.ts
```diff
@@ -2,7 +2,13 @@
  * Apache-2.0 */
 
 import { IItem, IUser, IGroup, IGeometry } from "@esri/arcgis-rest-types";
-import { IPortal } from "@esri/arcgis-rest-portal";
+import {
+  IPortal,
+  ISearchResult,
+  searchGroups,
+  searchItems,
+  searchUsers
```
tomwayson, line 10: did you mean to import these fns here?

tomwayson (commented):
This is fantastic work @drewctate, just a couple of questions, but otherwise this LGTM. Could be good to have @rweber-esri take a gander too.

packages/common/src/types.ts
drewctate: Good catch

packages/common/src/items/fetch-all-pages.ts
drewctate: Done.

packages/common/src/items/fetch-all-pages.ts
```diff
@@ -11,23 +11,50 @@ const MAX_NUM = 100;
...
+      : Promise.resolve({
+          nextStart: firstStart,
+          total: limit,
+          results: [],
+          num: pageSize
+        });
+
+  return promise
+    .then(firstResponse => {
+      if (firstResponse.nextStart === -1) return [firstResponse];
```
tomwayson, line 31:
```suggestion
      // no more requests needed, return the first response
      if (firstResponse.nextStart === -1) return [firstResponse];
```

packages/common/src/items/fetch-all-pages.ts
```diff
@@ -11,23 +11,50 @@ const MAX_NUM = 100;
...
       const starts = [];
-      for (let i = 1; i <= total; i += pageSize) starts.push(i);
+      for (
```
tomwayson, line 33:
```suggestion
      // generate batch requests for the remaining pages to fetch
      for (
```

packages/common/src/items/fetch-all-pages.ts
```diff
@@ -11,23 +11,50 @@ const MAX_NUM = 100;
...
         searchFunc({ ...opts, start, num: pageSize });
-      return batch(starts, batchSearchFunc);
+      return batch(starts, batchSearchFunc).then(responses => [
+        firstResponse,
+        ...responses
+      ]);
     })
     .then(responses => {
-      return responses.reduce(
+      const results = responses.reduce(
```
tomwayson, line 48: would be add a 
```suggestion
      // merge all the search results into a single array
      const results = responses.reduce(
```

tomwayson (approved):
This is really great, though I'd like to see a few comments. I made some suggestions where.

packages/common/src/items/fetch-all-pages.ts
```diff
@@ -0,0 +1,60 @@
+import { ISearchOptions, ISearchResult } from "@esri/arcgis-rest-portal";
+import { SearchableType, SearchFunction } from "../types";
+import { batch } from "../utils/batch";
+
+const MAX_NUM = 100;
+
+/**
+ * Fetches all the pages
+ * @param opts search options
```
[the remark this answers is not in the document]
drewctate: Good point.

packages/common/src/types.ts
```diff
@@ -173,6 +173,11 @@ export interface IHubGeography {
   geometry?: IGeometry;
 }
 
+export type SearchableType = IItem | IGroup | IUser;
+export type SearchFunction = (
+  ...args: any
```
rweber-esri, line 178: Since we're using the `rest` operator, I'd expect this to be

```
export type SearchFunction = (
  ...args: any[]
) => Promise<ISearchResult<SearchableType>>;
```

rweber-esri (approved):
Overall looks good, nice job. Added a couple minor comments, one signature-related, another comment-related.

PR #422, opened 2020-11-09
fix(hub-teams): respect org-level public sharing flag

dbouwman:
affects: @esri/hub-teams

When determining what access level to set for a Team, respect the org level public sharing flag
canSharePublic

### DIFF
--- a/packages/teams/src/utils/get-allowed-group-access.ts
+++ b/packages/teams/src/utils/get-allowed-group-access.ts
@@ -3,19 +3,22 @@ import { hasAllPrivileges } from "./has-all-privileges";
 import { GROUP_ACCESS_PRIVS } from "./group-access-privs";
 import { AGOAccess } from "../types";
 import { IPortal } from "@esri/arcgis-rest-portal";
+import { getProp } from "@esri/hub-common";
 
 /**
  * Returns the allowed group access based on a user's privileges
- * @param {String} requestedAccess public || org || private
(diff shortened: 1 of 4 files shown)


packages/teams/src/utils/get-allowed-group-access.ts
```diff
@@ -10,10 +11,15 @@ import { AGOAccess } from "../types";
  */
 export function getAllowedGroupAccess(
   requestedAccess: AGOAccess,
-  user: IUser
+  user: IUser,
+  portal: IPortal
```
tomwayson, line 15: This should be optional so that it's not a breaking change:

```suggestion
  portal?: IPortal
```

packages/teams/src/utils/get-allowed-group-access.ts
```diff
@@ -10,10 +11,15 @@ import { AGOAccess } from "../types";
...
 ): AGOAccess {
-  // figure out what they can create...
-  const canCreatePublic = hasAllPrivileges(user, GROUP_ACCESS_PRIVS.public);
+  // portal-wide flag takes presidence, and is not sync'd with privs
+  const portalWideCanSharePublic = portal.canSharePublic;
```
tomwayson, line 18: in order to prevent a breaking change, if portal is not passed in we should assume it's allowed, right?

tomwayson (commented):
In addition the changes I suggest you'll probably need a test for the case where portal is not passed in.

PR #424, opened 2020-11-13
Hoist OperationStack, OperationError and utils from Solution.js

dbouwman:
Adds:
- OperationStack & OperationError
- Operation Stack Doc
- last, uniqueBy, filterBy

Pulled from new Solution.js, but need here for use in sites package

### DIFF
--- a/packages/common/test/util.test.ts
+++ b/packages/common/test/util.test.ts
@@ -12,7 +12,10 @@ import {
   unique,
   extend,
   addDays,
-  chunkArray
+  chunkArray,
+  uniqueBy,
+  last,
+  filterBy
 } from "../src/util";
 
 import { getProp } from "../src/objects";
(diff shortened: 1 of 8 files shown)


packages/common/test/util.test.ts
```diff
@@ -675,4 +678,40 @@ describe("util functions", () => {
...
+      const apps = filterBy(arr, "item.type", "App");
+      expect(apps.length).toBe(2, "should have two apps");
+      apps.forEach(app => {
+        expect(app.item.type).toBe("App");
+      });
+    });
+  });
+
+  describe("uniqueBy: ", () => {
+    it("returns uniqu entries by deep prop", () => {
```
tomwayson, line 706:
```suggestion
    it("returns unique entries by deep prop", () => {
```

packages/common/test/util.test.ts
```diff
@@ -675,4 +678,40 @@ describe("util functions", () => {
...
+      const arr = [
+        { item: { type: "Map", id: "map1" } },
+        { item: { type: "App", id: "app1" } },
+        { item: { type: "Map", id: "map1" } },
+        { item: { type: "App", id: "app2" } }
+      ];
+      const chk = uniqueBy(arr, "item.id");
```
tomwayson, line 713: seems like `uniqueBy(arr, "item.type")` would be a better example just b/c we expect id to be unique across all items.

tomwayson (approved):
Good stuff. Stack docs are a help. Left minor comments that may not be worth addressing in this PR.

packages/common/test/util.test.ts
dbouwman: I'll mull this - how I use it, and why I wrote it, was actually to unique a set of `IDependency` objects, by Item id, so we don't try to convert the same item multiple times... but yeah - in general we'd not expect to have dupes of item-ish things

PR #425, opened 2020-11-13
Add errors array to hub api metadata and status to polling event metadata

rgwozdz:
First off, apologies.  This PR includes a lot of changes due to auto-lint fixing.

The actual purpose of the PR is to standardize the error events that get emitted if there was an error in the download export process.  All error events look like:
```
{
 detail: { error, metadata }
}
```

To facilitate that, I have also changed the response of the hubDownloadMetadata method to include the `errors` array from the Hub API.

All other changes are tests and linting.

### DIFF
--- a/packages/downloads/src/portal/portal-poll-export-job-status.ts
+++ b/packages/downloads/src/portal/portal-poll-export-job-status.ts
@@ -87,9 +87,8 @@ class PortalPoller implements IPoller {
           if (metadata.status === "failed") {
             eventEmitter.emit(`${downloadId}ExportError`, {
               detail: {
-                metadata: {
-                  errors: [new Error(metadata.statusMessage)]
-                }
+                error: new Error(metadata.statusMessage),
+                metadata: { status: "error" }
               }
             });
             return this.disablePoll();
@@ -98,11 +97,11 @@ class PortalPoller implements IPoller {
         .catch((error: any) => {
           if (error instanceof ExportCompletionError) {
             eventEmitter.emit(`${downloadId}ExportError`, {
-              detail: { metadata: { errors: [error] } }
+              detail: { error, metadata: { status: "error" } }
             });
           } else {
             eventEmitter.emit(`${downloadId}PollingError`, {
-              detail: { error }
+              detail: { error, metadata: { status: "error" } }
             });
           }
           return this.disablePoll();
--- a/packages/downloads/src/portal/portal-poll-export-job-status.ts
+++ b/packages/downloads/src/portal/portal-poll-export-job-status.ts
@@ -15,6 +15,11 @@ import { urlBuilder } from "../utils";
 import { UserSession } from "@esri/arcgis-rest-auth";
 import { IPoller } from "../poller";
 
+enum DownloadStatus {
+  READY = "ready",
+  ERROR = "error"
+}
+
 class ExportCompletionError extends Error {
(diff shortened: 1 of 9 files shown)


robertsteilberg (approved):
LGTM

packages/downloads/src/request-download-metadata.ts
```diff
@@ -24,28 +24,31 @@ export interface IDownloadMetadataRequestParams {
 
 export interface IDownloadMetadataResults {
   /* Identifier for the download */
-  downloadId: string,
+  downloadId: string;
 
-  /* ready, not_ready, creating, updating, failed*/
-  status: string,
+  /* ready, not_ready, creating, updating, failed */
+  status: string;
```
mikeringrose, line 30: why isn't this an enum?

mikeringrose (approved):

packages/downloads/src/request-download-metadata.ts
tomwayson: or a `type` like `'ready' | 'not_ready' | 'creating' | 'updating' | 'failed'`?

That's what we use in other places for this sort of thing.

@mikeringrose I still don't have good instincts on when to use one of the many ways (`type`, `enum`, `const enum`, etc) TS offers for this kind of situation. Worse, I'm sort of mentally blocked on it after having been [burned by making the wrong choice over in rest-js](https://github.com/Esri/arcgis-rest-js/pull/681). If you have experience/insight, I think it would be great to document some guidelines here.

packages/downloads/src/portal/portal-poll-export-job-status.ts
```diff
@@ -98,11 +97,11 @@ class PortalPoller implements IPoller {
         .catch((error: any) => {
           if (error instanceof ExportCompletionError) {
             eventEmitter.emit(`${downloadId}ExportError`, {
-              detail: { metadata: { errors: [error] } }
```
tomwayson, line 101: this makes it look like a breaking change. in order to avoid that we should probably keep `metadata.errors` (in addition to `detail.error`) until the next major release. You can add it to this list: https://github.com/Esri/hub.js/issues/305

tomwayson (commented):
I'm worried by the linting changes. I fixed the broken pre-commit hook back in August, and cleaned up the exisitng linting errors: https://github.com/Esri/hub.js/pull/314 That implies that these errors have slipped in since then?

packages/downloads/src/request-download-metadata.ts
rgwozdz: @mikeringrose - I'm using a `DownloadStatus` type now.

packages/downloads/src/portal/portal-poll-export-job-status.ts
rgwozdz: I've made the change to keep `metadata.errors`.  As of this writing, the code snip above isn't updating to show it, but you can view it [here](https://github.com/Esri/hub.js/blob/p/download-errors/packages/downloads/src/portal/portal-poll-export-job-status.ts#L93-L111).

rgwozdz:
@tomwayson -  I think the linting-issues slipped in quite a while ago.  I believe they occurred on a commit in which I did not use `npm run c`, but rather did `git commit`.

tomwayson (approved):
LGTM

PR #426, opened 2020-11-13
Specify "Collection" when searching for multilayer KML or CSV exports

rgwozdz:
If an item has multiple layers, and a KML or CSV item is requested, the /export endpoint actually tags the new export item as `CSV Collection` and `KML Collection`.  As a result, we need to search for that type (rather than KML or CSV) when checking if such an export is already cached.

Further, for multi-layer exports, we need fetch the last-edit-date of each layer and see if any are more recent then the timestamp of the cached multilayer export.

I moved the rest-js dependencies in hub-downloads to peers, but had to update the versions to leverage recent rest-js changes.

Also, there are some changes due to auto-linting that didn't occur earlier.


packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -67,30 +81,61 @@ export function portalRequestDownloadMetadata(
...
+      return timestamp !== undefined;
+    })
+    .sort((a, b) => {
+      return b - a;
+    });
+  return result[0];
+}
+
+function getSearchFormat(format: DownloadFormat, multilayer: boolean) {
+  if (multilayer && (format === "CSV" || format === "KML")) {
```
mikeringrose, line 126: Can't the download types be an enum?

packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -67,30 +81,61 @@ export function portalRequestDownloadMetadata(
...
+      const { editingInfo: { lastEditDate } = {} } = layer;
+      return lastEditDate;
+    })
+    .filter(timestamp => {
+      return timestamp !== undefined;
+    })
+    .sort((a, b) => {
+      return b - a;
+    });
+  return result[0];
```
mikeringrose, line 122: Aren't these already sorted from the API?

packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -67,30 +81,61 @@ export function portalRequestDownloadMetadata(
...
-    const editingInfo: any | null = layer.editingInfo;
-    return editingInfo ? editingInfo.lastEditDate : undefined;
-  });
+  return getService({ url, authentication }).then(
+    (response: IFeatureServiceDefinition) => {
+      const layers: ILayerDefinition[] = response.layers;
+      const multilayer = isMultilayerRequest(layerId, layers);
+      return {
+        format: getSearchFormat(format, multilayer),
+        modified: extractLastEditDate(layers)
```
mikeringrose, line 100: Can layers be null or empty?

packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -67,30 +81,61 @@ export function portalRequestDownloadMetadata(
...
+    .filter(timestamp => {
+      return timestamp !== undefined;
+    })
+    .sort((a, b) => {
+      return b - a;
+    });
+  return result[0];
+}
+
+function getSearchFormat(format: DownloadFormat, multilayer: boolean) {
```
mikeringrose, line 125: Generally considered an anti-pattern (https://martinfowler.com/bliki/FlagArgument.html), why not just be explicit?

packages/downloads/src/portal/portal-request-download-metadata.ts
rgwozdz: No, I don't think so.  Here I want to sort the layers on lastEditDate (DESC).  It's a nested property (layer.editingInfo.lastEditDate), and not all layers will have it.

packages/downloads/src/portal/portal-request-download-metadata.ts
rgwozdz: ArcGIS rest-js indicate `layers` is non-optional, but I have now added handling in the event is comes back null or undefined.  Also added a test for that situation.

packages/downloads/src/portal/portal-request-download-metadata.ts
rgwozdz: Eliminated the flag param.

packages/downloads/src/portal/portal-request-download-metadata.ts
rgwozdz: Refactored as enums, also refactored the item type checks as enums.

packages/downloads/package.json
```diff
@@ -10,14 +10,19 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
-    "@esri/arcgis-rest-feature-layer": "^2.14.1",
```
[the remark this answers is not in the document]
rgwozdz: Rebase took care of this.

packages/downloads/package.json
```diff
@@ -10,14 +10,19 @@
...
+    "@esri/arcgis-rest-feature-layer": "^2.21.0",
+    "@esri/arcgis-rest-portal": "^2.21.0",
+    "@esri/arcgis-rest-request": "^2.21.0"
+  },
   "devDependencies": {
-    "@esri/arcgis-rest-auth": "^2.14.1",
+    "@esri/arcgis-rest-auth": "^2.21.0",
+    "@esri/arcgis-rest-feature-layer": "^2.21.0",
+    "@esri/arcgis-rest-portal": "^2.21.0",
+    "@esri/arcgis-rest-request": "^2.21.0",
```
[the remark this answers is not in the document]
rgwozdz: Rebase took care of this.

packages/downloads/src/portal/portal-request-download-metadata.ts
mikeringrose: i was thinking more of having a download-formats.ts:
```
enum DownloadFormats {
  CSV: "CSV",
  KML: "KML",
  Shapefile: "Shapefile",
  ...
}

isCollectionType(format: DownloadFormat) {
  return [DownloadFormats.CSV, DownloadFormats.KML].includes(format);
}
```

rgwozdz: Made the change in the contents of download-format.ts:

```
export enum DownloadFormats {
  Shapefile = "Shapefile",
  CSV = "CSV",
  KML = "KML",
  GeoJson = "GeoJson",
  Excel = "Excel",
  "File Geodatabase" = "File Geodatabase",
  "Feature Collection" = "Feature Collection",
  "Scene Package" = "Scene Package"
}

export type DownloadFormat = keyof typeof DownloadFormats
```

I did the isCollectionType function like

```
const isCollectionType = (format: DownloadFormat) => format === DownloadFormats.CSV || format === DownloadFormats.KML;
```

Because the typescript compile didn't recognize `.includes`.

packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -1,9 +1,21 @@
...
+  IFeatureServiceDefinition,
+  ILayerDefinition
+} from "@esri/arcgis-rest-feature-layer";
+import { DownloadFormat, DownloadFormats } from "../download-format";
 import { urlBuilder, composeDownloadId } from "../utils";
 
+enum ItemTypes {
+  FeatureService = "Feature Service",
+  MapService = "Map Service"
+}
```
mikeringrose, line 14: These could probably just be constants. An enum here might be overkill.

mikeringrose (approved):

Issue #306, opened 2020-08-03
consolidate packages


tomwayson:
Seems to me that there is unnecessary friction in the development and consumption of this library due the large number of packages.

The strategy of micro-packages that are peer depenedencies came from rest-js. When we started work on that, tree shaking and code splitting were less reliable, and not available to the Ember apps that were going to use the packages. Neither of those are true any longer. We can assume that most consumers will use the ESM builds and be able to tree-shake out just the code that their apps need.

I think we could:
- merge search and downloads in to content (at the very least)
- merge types into common
- merge teams into members (or vice versa)?
- merge sites into initiatives (or vice versa)?
- merge auth into common?

I think things like events and surveys should stay on their own.

tomwayson:
@cpgruber what would you think about consolidating `hub-surveys` into `hub-content`?

@ajturner ultimately I could see a `hub-feedback` package that replaces `hub-annotations` and includes all that code and it would depend on the @hub-content package for survey related functions.

cpgruber:
> @cpgruber what would you think about consolidating `hub-surveys` into `hub-content`?
> 
> @ajturner ultimately I could see a `hub-feedback` package that replaces `hub-annotations` and includes all that code and it would depend on the @hub-content package for survey related functions.

I would defer to @rweber-esri and @brollison

brollison:
@cpgruber I don't believe I have enough expertise here to comment in a helpful way; however, if a package called `hub-feedback` is meant to exist, then `hub-surveys` may be better suited there over `hub-content`.  There is a lot we do or plan to do with surveys which treats them differently from most other content, but I'm not sure if that means they need to be separated from more generic content items.

// @rweber-esri please be more helpful / constructive than me ;-)

rweber-esri:
@tomwayson 

It may help to have a little more context into what the friction is specifically. Is the friction that people are having difficulty finding what they need or more along the lines of build/version bump pains (suspect the latter)? My main concern about consolidating the different type-specific packages into a more generic `content` package is centered around exporting & collisions. We do a lot of `export * from "../wherever";` and could have some collisions that would force us to rename functions and/or add an abstraction layer like `solutions-service` in `opendata-ui` that delegates.

tomwayson:
> more along the lines of build/version bump pains (suspect the latter)

correct

re: exporting and collisions, that's already an issue since our UMD builds put all the fns into a single namespace (`arcgisHub`).

tomwayson:
[...]

PR #427, opened 2020-11-13
Downloads: if item is a portal item and it is multilayer, export all layers

rgwozdz:
Download export requests for multilayer, portal items should export ALL layers.  This PR ensures that occurs by remove the "layers" property from the `exportItem` call.

### DIFF
--- a/packages/downloads/package.json
+++ b/packages/downloads/package.json
@@ -10,14 +10,14 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
-    "@esri/arcgis-rest-feature-layer": "^2.14.1",
-    "@esri/arcgis-rest-portal": "^2.14.1",
-    "@esri/arcgis-rest-request": "^2.14.1",
+    "@esri/arcgis-rest-feature-layer": "^2.21.0",
+    "@esri/arcgis-rest-portal": "^2.21.0",
+    "@esri/arcgis-rest-request": "^2.21.0",
     "eventemitter3": "^4.0.4",
     "tslib": "^1.13.0"
   },
   "devDependencies": {
-    "@esri/arcgis-rest-auth": "^2.14.1",
+    "@esri/arcgis-rest-auth": "^2.21.0",
     "@rollup/plugin-commonjs": "^15.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
--- a/packages/downloads/src/portal/portal-request-dataset-export.ts
+++ b/packages/downloads/src/portal/portal-request-dataset-export.ts
@@ -42,7 +42,12 @@ export function portalRequestDatasetExport(
 
 function composeExportParameters(params: any) {
   const { datasetId, spatialRefId, where } = params;
-  const layerId = datasetId.split("_")[1] || 0;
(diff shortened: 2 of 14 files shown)


robertsteilberg (approved):
LGTM but someone more experienced that I should review the dep changes

mikeringrose (approved):

packages/downloads/package.json
```diff
@@ -10,14 +10,19 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
-    "@esri/arcgis-rest-feature-layer": "^2.14.1",
-    "@esri/arcgis-rest-portal": "^2.14.1",
-    "@esri/arcgis-rest-request": "^2.14.1",
```
tomwayson, line 15: I'm glad to see these become `peerDepenencies`, but Unfortunately, this will manifest as breaking change b/c people will now need to install these.

Also, I'm considering consolidating this package w/ `hub-content` (#306), which would solve this issue.

I think we should update the versions, but leave them as `dependencies` for now.

packages/downloads/src/portal/portal-request-dataset-export.ts
```diff
@@ -42,7 +42,12 @@ export function portalRequestDatasetExport(
 
 function composeExportParameters(params: any) {
   const { datasetId, spatialRefId, where } = params;
-  const layerId = datasetId.split("_")[1] || 0;
+  const layerId = datasetId.split("_")[1];
```
tomwayson, line 45: FYI - there's a util for this:

https://esri.github.io/hub.js/api/content/parseDatasetId/

Unfortunately it's not available to this package, let's leave a TODO comment like this

```suggestion
  // TODO: move parseDatasetId() to hub-common and use that here
  const layerId = datasetId.split("_")[1];
```

tomwayson (changes requested):

packages/downloads/package.json
```diff
@@ -10,14 +10,19 @@
...
     "tslib": "^1.13.0"
   },
+  "peerDependencies": {
+    "@esri/arcgis-rest-feature-layer": "^2.21.0",
+    "@esri/arcgis-rest-portal": "^2.21.0",
+    "@esri/arcgis-rest-request": "^2.21.0"
+  },
   "devDependencies": {
-    "@esri/arcgis-rest-auth": "^2.14.1",
+    "@esri/arcgis-rest-auth": "^2.21.0",
```
tomwayson, line 22: Also, we'll need to update these `devDependencies` in all other packages to match these (even if they're `dependencies`), as per https://github.com/Esri/hub.js/wiki/Dependencies#outside-the-monorepo:

> Then you would bump the version of the dependency in all existing devDependencies within the monorepo to the new minor version.

packages/downloads/package.json
rgwozdz: Done.

packages/downloads/src/portal/portal-request-dataset-export.ts
rgwozdz: Done.

packages/downloads/package.json
rgwozdz: Done.

tomwayson (approved):
thanks @rgwozdz!

PR #434, opened 2020-12-15
now creating the initiative item for basic orgs

drewctate:


packages/sites/src/create-site-model-from-template.ts
```diff
@@ -108,7 +108,7 @@ export function createSiteModelFromTemplate(
       // create the initiative
       let handleInitiative = Promise.resolve(null);
 
-      if (product === "premium") {
+      if (product !== "portal") {
```
dbouwman, line 111: we double checked that `portal` is the product name ?

dbouwman (approved):
Thanks!

PR #436, opened 2020-12-18
F/161849 move unified search merge sort

drspacemanphd:
[161489](https://esriarlington.tpondemand.com/entity/161849-move-merging-and-sorting-functions-into)

As an initial step towards a more unified search interface, several methods currently in Hub-Indexer are being refactored and moved to Hub.js. This PR moves the merge-sort functionality of Hub-Indexer Unified Content Search into Hub.js. It provides a more generic method for merging (and potentially sorting) several individually sorted lists of data, as is necessary for unified search when merging items from Portal and datasets from the Elasticsearch Index.


packages/search/test/util/merge-sort/merge.test.ts
```diff
@@ -0,0 +1,946 @@
...
+    const lists = [
+      first,
+      second,
+      third,
+      fourth,
+      fifth,
+      sixth,
+      seventh,
+      eighth,
+      nineth,
```
robertsteilberg, line 156: typo

packages/search/src/util/merge-sort/merge.ts
```diff
@@ -0,0 +1,300 @@
...
+ *
+ * The returned object contains merged (and potentially sorted) data, as well as key-value
+ * pairs that represent the index of a result set in the "data" parameter (the key) tied to how
+ * many results (the value) of that result set were added to the final merged data array.
+ * For example, if "data" represents an array of two result sets, with 3 and 5 results added to the final
+ * result set, respectively, the returned object would be { data: [...], "0": 3, "1": 5 }
+ *
+ * @param data An array of of result sets, each an array of type T
+ * @param resultLimit the maximum number of merged results to return, defaults to 10
+ * @param cmptr comparator function that takes in two instances of type T and returns a number to determine sorting order
```
robertsteilberg, line 30: note what the rules are here regarding the number--i.e. +1 if a comes before b, -1 if b comes before a, 0 if equivalent, etc

packages/search/src/util/merge-sort/merge.ts
```diff
@@ -0,0 +1,300 @@
...
+  resultLimit: number = 10,
+  cmptr?: comparator<T[]>,
+  direction?: HeapDirection
+): { data: T[]; [key: string]: any } {
+  const results = cmptr
+    ? kMergeSort(data, cmptr, direction, resultLimit)
+    : kMergeDefault(data, resultLimit);
+
+  // Zero-fill any result set that did not have any results added to final
+  for (let i = 0; i < data.length; i++) {
```
robertsteilberg, line 45: why do we do this?

packages/search/test/util/merge-sort/merge.test.ts
drspacemanphd: Good call, this has been fixed.

packages/search/src/util/merge-sort/merge.ts
drspacemanphd: Good call, the name `comparator` is inspired by the corresponding [Java interface](https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html), but not everyone would know

packages/search/src/util/merge-sort/merge.ts
drspacemanphd: This is completely optional, but if a result set that is provided as input had 0 results added to the final set, the corresponding index key in the return object would be undefined. This sets it to 0 so client doesn't have to check.

robertsteilberg (approved):
this LGTM, might benefit to have someone with seasoned eyes look over it as well. i read thru the algo but did not take the time to fully reason thru it.

packages/search/src/util/merge-sort/merge.ts
```diff
@@ -0,0 +1,300 @@
...
+ * If a comparator function used to compare data is provided, a k-way merge sort is performed using
+ * a Binary Heap implementation. Otherwise data from each input array is merged incrementally
+ *
+ * The returned object contains merged (and potentially sorted) data, as well as key-value
+ * pairs that represent the index of a result set in the "data" parameter (the key) tied to how
+ * many results (the value) of that result set were added to the final merged data array.
+ * For example, if "data" represents an array of two result sets, with 3 and 5 results added to the final
+ * result set, respectively, the returned object would be { data: [...], "0": 3, "1": 5 }
+ *
+ * @param data An array of of result sets, each an array of type T
```
trescube, line 28: duplicate "of"

packages/search/src/util/merge-sort/merge.ts
```diff
@@ -0,0 +1,300 @@
...
+    !_hasReachedLimit(results.data.length, resultLimit) &&
+    totalDataRemaining > 0
+  ) {
+    const dataToAdd: T[] = [];
+    data.forEach((dataArr, index) => {
+      if (
+        dataArr.length > 0 &&
+        _canDataBeAdded(dataToAdd, results.data, resultLimit)
+      ) {
+        results[index.toString()] = results[index.toString()]
```
trescube, line 74: this is the bookkeeping for how many from each dataset were added, right?

packages/search/test/util/merge-sort/merge.test.ts
```diff
@@ -0,0 +1,946 @@
...
+
+      // Assertions
+      expect(result).toBeDefined();
+      expect(result).toEqual({ data: expected, "0": 0, "1": 4, "2": 3 });
+    });
+
+    it("should be able to handle 3 results set with data", () => {
+      // Setup
+      const sortFunc = (a: number, b: number) => a - b;
+      const arrOne: number[] = Array.apply(null, Array(4))
```
trescube, line 861: A more clear way to write this is with lodash, a la: `_.range(0, 4).map(...).sort(...)`

trescube (approved):

packages/search/src/util/merge-sort/merge.ts
drspacemanphd: Thanks, this has been removed.

packages/search/src/util/merge-sort/merge.ts
drspacemanphd: Yes that's correct.

packages/search/test/util/merge-sort/merge.test.ts
drspacemanphd: Lodash currently isn't a dependency. If we decide to use it, I can certainly change this, as well as use other lodash functions in the actual implementation

trescube: Oh right, I forgot that this is hub.js and lodash isn't included

PR #438, opened 2020-12-18
fix(content): add structuredLicense to content from dataset

jPurush:
…dLicense to content from dataset

affects: @esri/hub-content

### DIFF
--- a/packages/content/src/hub.ts
+++ b/packages/content/src/hub.ts
@@ -83,7 +83,8 @@ export function datasetToContent(dataset: DatasetResource): IHubContent {
     modifiedProvenance,
     slug,
     searchDescription,
-    groupIds
+    groupIds,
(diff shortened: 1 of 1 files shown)


tomwayson (approved):
this is good as long you add an issue referencing this PR about whether or not we want to do this when from fetching from portal and linking out the hub-indexer implementation as well as the duplicate code you mentioned in -ui

jPurush:
> this is good as long you add an issue referencing this PR about whether or not we want to do this when from fetching from portal and linking out the hub-indexer implementation as well as the duplicate code you mentioned in -ui

done! https://github.com/Esri/hub.js/pull/439

jPurush:
Screen Shot 2020-12-18 at 2 50 22 PM

PR #441, opened 2020-12-20
Adding Github Actions

mikeringrose:
Enables Github Actions. Two workflow have been added:

1. `build-test` - runs on pull requests and pushes to master
2. `gh-pages` - only runs on master, after the build-test workflow completes

For the `gh-pages` workflow to run correctly, a secret needs to be created named `HUBJS_TOKEN` with the scope of repo:public.


drspacemanphd (approved):

tomwayson:
This is awesome @mikeringrose. Can we see the output of the `build-test` on this PR?

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    strategy:
+      matrix:
+        node-version: [10.x]
```
tomwayson, line 17: can we specify something like `node, lts/*` to test on the current stable, and latest LTS?

tomwayson: and should we #443

mikeringrose:
> This is awesome @mikeringrose. Can we configure it so that we see the [output of the `build-test` action](https://github.com/Esri/hub.js/actions) inline in this PR?

The codecov output in the comments from the github action.

tomwayson (approved):
Now I see the link to the action output under the checks section.

I whole heartedly approve this PR. Great work @mikeringrose!

PR #442, opened 2020-12-21
f/169296 added content paginator

drspacemanphd:
[169296](https://esriarlington.tpondemand.com/entity/169296-move-pagination-and-sorting-functions-into)

This is the second of a series of PRs that move business logic associated with unified search to Hub.js. In this PR, pagination is moved to hub.js. The general pattern was meant to be compatible with a future generic pagination interface, such that content search and [user search](https://devtopia.esri.com/dc/hub-profiles-api/blob/master/apps/main-api/src/graphql.schema.ts#L125) paginations return a similar structure.

### DIFF
--- a/packages/search/package.json
+++ b/packages/search/package.json
@@ -10,6 +10,7 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
+    "base-64": "^1.0.0",
     "tslib": "^1.13.0"
   },
   "peerDependencies": {
@@ -30,6 +31,7 @@
     "@rollup/plugin-commonjs": "^15.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
+    "@types/base-64": "^0.1.3",
     "@types/faker": "^5.1.5",
     "faker": "^5.1.0",
     "rollup": "^2.26.5",
--- a/packages/search/src/util/pagination/content-paginator.ts
+++ b/packages/search/src/util/pagination/content-paginator.ts
@@ -1,0 +1,48 @@
+import { encode } from "base-64";
+import { IPageResponse } from "./paginator";
+
+interface IContentPageStart {
+  hub: number;
+  ago: number;
+}
+
+interface IContentPage {
+  hubAdded: number;
+  agoAdded: number;
+  hubTotal: number;
+  agoTotal: number;
+}
+
+/**
+ * @param start - starting cursor when perfoming current search
+ * @param page - data associated with the current page of results
+ * @returns an object with a stringified cursor for use in searching next page of results,
+ * the total number of results, and if there are no more pages left
+ *
+ * Note: The signature for this could be converted to a more generic interface with two parameters:
+ * start - indicates the starting point for the current search
+ * page - indicates the page of the current set of reseults to use to determine the "next" starting point
+ */
+export function pageContent(
+  start: IContentPageStart,
+  page: IContentPage
+): IPageResponse {
+  const hubNextStart: number = (start.hub || 0) + (page.hubAdded || 0);
+  const agoNextStart: number = (start.ago || 0) + (page.agoAdded || 0);
+
+  const cursor = encode(
+    JSON.stringify({
+      hub: hubNextStart,
+      ago: agoNextStart
+    })
+  );
+  const total = (page.hubTotal || 0) + (page.agoTotal || 0);
+  const hasNextPage =
+    hubNextStart < page.hubTotal || agoNextStart < page.agoTotal;
+
+  return {
+    cursor,
+    total,
+    hasNextPage
+  };
+}
--- a/packages/search/src/util/pagination/content-paginator.ts
+++ b/packages/search/src/util/pagination/content-paginator.ts
@@ -1,16 +1,33 @@
 import { encode } from "base-64";
 import { IPageResponse } from "./paginator";
 
+/**
+ * Interface for specifying the starting page for a particular content search
+ * in an AGO (non-Portal) environment, where AGO and Hub Indexer content results are combined
+ * as part of returning search results
+ *
+ * @param hub the starting point for Hub Index content
+ * @param ago the starting point for AGO (private) content
+ */
 interface IContentPageStart {
   hub: number;
   ago: number;
 }
 
+/**
+ * Interface for specifying the number of Hub Index (hub) and AGO (ago) items
+ * returned as part of the current page of search results.
+ *
+ * @param hubRecordsAdded the number of Hub Index records returned as part of results page
+ * @param agoRecordsAdded the number of AGO records returned as part of results page
+ * @param hubRecordsTotal the number of total Hub Index records to be returned as part of search
+ * @param agoRecordsTotal the number of total AGO records to be returned as part of search
+ */
 interface IContentPage {
-  hubAdded: number;
-  agoAdded: number;
-  hubTotal: number;
-  agoTotal: number;
+  hubRecordsAdded: number;
+  agoRecordsAdded: number;
+  hubRecordsTotal: number;
+  agoRecordsTotal: number;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


packages/search/package.json
```diff
@@ -10,6 +10,7 @@
   "types": "dist/esm/index.d.ts",
   "license": "Apache-2.0",
   "dependencies": {
+    "base-64": "^1.0.0",
```
tomwayson, line 13: Browsers have built-in fns lik [btoa()](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) for this, I assume this is needed for node? It's not a big deal to add this I suppose at a whopping [1kb gzipped](https://bundlephobia.com/result?p=base-64@1.0.0). I tried looking into shimming the browser built-ins in a node env, but came across [this comment](https://stackoverflow.com/questions/23097928/node-js-throws-btoa-is-not-defined-error#comment111681927_46524254) which leads me to believe that ultimately it's best to have a consistent implementation on both sides of the wire.

packages/search/src/util/pagination/content-paginator.ts
```diff
@@ -0,0 +1,48 @@
+import { encode } from "base-64";
+import { IPageResponse } from "./paginator";
+
+interface IContentPageStart {
+  hub: number;
+  ago: number;
```
tomwayson, line 6: If this applies to enterprise, this should be `portal`, not `ago`.

packages/search/src/util/pagination/content-paginator.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+interface IContentPage {
+  hubAdded: number;
+  agoAdded: number;
```
tomwayson, line 11: It's not clear to me what these represent. Number of pages to add to the current position for each? Or are they records? Generally it would be good to document w/ code comments (which will show up in docs) all the members of each interface.

packages/search/src/util/pagination/paginator.ts
```diff
@@ -0,0 +1,5 @@
+export interface IPageResponse {
+  cursor: string;
+  total: number;
```
tomwayson, line 3: is this total records or pages?

tomwayson (commented):
Nice work. A few questions.

packages/search/package.json
drspacemanphd: Yes your assumption is correct. These are functions used by the Hub Indexer API so it will be in Node env. Also came across the same comment.

packages/search/src/util/pagination/content-paginator.ts
drspacemanphd: This is directly invoked from the indexer, so AGO only.

tomwayson: So, in the scenario where the app running in enterprise (w/ no access to the the Hub API) is paging through search results, what fn would it call?

drspacemanphd: This is a util function to be used by the Hub Indexer only, as it's only relevant when we have to create a paging cursor that points to remaining Portal (AGO) and Hub Index content remaining to be served. In a Portal environment, only portal content is served with pagination already supported by the API.

The scope of this is currently limited to moving over helper functions from the indexer (NodeJS) to Hub.js (see TP [here](https://esriarlington.tpondemand.com/entity/169296-move-pagination-and-sorting-functions-into), but I can certainly revisit this as implementation begins for a more unified content search API.

packages/search/src/util/pagination/content-paginator.ts
drspacemanphd: Good call. I changed these to `hubRecordsAdded` and `agoRecordsAdded`, respectively. I also changed the other two props to `hubRecordsTotal` and `agoRecordsTotal` and added documentation to the interfaces created as part of this PR.

packages/search/src/util/pagination/paginator.ts
drspacemanphd: This is total records and I've added documentation to clarify that.

packages/search/src/util/pagination/content-paginator.ts
tomwayson: I guess I would argue that if it's to be used by the Hub indexer only, then why move it out of the indexer and into Hub.js? especially considering that these bits will end up in the client Ember app and never be used.

drspacemanphd: It's a valid question. We are currently performing the pre-requisite work for a more unified content search, with Hub.js as the main entryway. Although this particular function may be a bit more specific to Hub Indexer, it should follow similar patterns for pagination that we are developing for other search work that will also be in Hub.js, e.g. user searching.

tomwayson: It's good to have the same patterns, but anything that will not be used on the client should not end up in Hub.js. Tree-shaking doesn't appear to be working in our app (and really it's the exception more than the rule for it to work anywhere), so _every_ line of code in Hub.js goes into the Hub application. Worse, because we have very "chunky" lazy loading, it's safe to assume most if not all Hub.js code lands in the initial page load. Specifically, b/c the search package is included by the app itself (and not in a lazy-loaded engine) it absolutely is included in that initial payload.

tomwayson (changes requested):
Please limit this PR to only code that will be used on the client.

mikeringrose:
> Please limit this PR to only code that will be used on the client.

@tomwayson we'll come up with a solution, but as was mentioned before the code should be able to run in node.js

packages/search/src/util/pagination/content-paginator.ts
```diff
@@ -0,0 +1,65 @@
...
+ */
+interface IContentPage {
+  hubRecordsAdded: number;
+  agoRecordsAdded: number;
+  hubRecordsTotal: number;
+  agoRecordsTotal: number;
+}
+
+/**
+ * @param start - starting cursor when perfoming current search
```
robertsteilberg, line 34: typo

packages/search/package.json
mikeringrose: It could implemented with an interface in which you inject the function to do the base64 encoding, but doesn't now have build in base64 support any ways:
```
Buffer.from("gmike is amazing").toString('base64')
```

robertsteilberg:
great work!

mikeringrose:
couldn't the paginator be written a bit more generically without losing functionality. if we think of our search "sources" in terms of an interface, portal, hub, etc are just sources. couldn't a source be represented as a Source type like so,
```
interface Source {
    name: string;
    next: () => Promise<IPageResponse>;
}
```
Then the paginator code would just have to concatenate all of the responses from the various sources and do its magic. then we wouldn't have to hardcode AGO and Hub anywhere. This would also allow us the ability to "register" different sources on the server to do the searching and give us a hook for optimizations.

It's late so maybe I'm missing the point.

packages/search/package.json
drspacemanphd: This is a better approach. I will update to use this.

drspacemanphd:
> Couldn't the paginator be written a bit more generically without losing functionality. if we think of our search "sources" in terms of an interface, portal, hub, etc are just sources. Couldn't a source be represented as a Source type like so,
> 
> ```
> interface Source {
>     name: string;
>     next: () => Promise<IPageResponse>;
> }
> ```
> 
> Then the paginator code would just have to concatenate all of the responses from the various sources and do its magic. then we wouldn't have to hardcode AGO and Hub anywhere. This would also allow us the ability to "register" different sources on the server to do the searching and give us a hook for optimizations.
> 
> It's late so maybe I'm missing the point.

I don't think you're missing the point at all. It's basically what I did with [refactoring the merge-sort](https://github.com/Esri/hub.js/pull/436). Rather than hardcoding in Hub and Ago as sources, the interface only specified sources of data as an array of arrays. Given that the input for a generic pagination could be many sources, and the output would be an encoded string, care would have to be taken to ensure that the decoded output makes sense from a client perspective. I think having an input interface that specifies a `string` label, similar to what you have above, would be helpful.

drspacemanphd:
Given the conversation we had yesterday, and the goal to make this a bit more generic, I'm closing this without merging. I will ping the reviewers here when an updated PR is created.

PR #444, opened 2020-12-22
fix(downloads): portalRequestDownloadMetadata() function did not recongize layer id

haoliangyu:
affects: @esri/hub-downloads

The `fetchCacheSearchMetadata()` in the `downloads` package was not passing the layer id to its internal util function `fetchCacheSearchMetadata`. As a result, it created incorrect item types to query cache for sublayers of a feature service item. For example, it should be `CSV`, not `CSV Collection` for an individual sublayer.

### DIFF
--- a/packages/downloads/test/portal/portal-request-download-metadata.test.ts
+++ b/packages/downloads/test/portal/portal-request-download-metadata.test.ts
@@ -554,6 +554,77 @@ describe("portalRequestDownloadMetadata", () => {
       }
     });
 
+    it("layer id, multi-layer, no cachehd", async done => {
+      try {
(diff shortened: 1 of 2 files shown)


packages/downloads/test/portal/portal-request-download-metadata.test.ts
```diff
@@ -554,6 +554,77 @@ describe("portalRequestDownloadMetadata", () => {
       }
     });
 
+    it("layer id, multi-layer, no cachehd", async done => {
```
rgwozdz, line 557: Typo in test name.

rgwozdz (approved):
Just the one typo in the test label.  

Thanks for fixing this.

PR #445, opened 2021-01-06
 F/user search proto

r00b:
Primer for a client of the new GraphQL-based User Indexing API. The feature is described in TP [here](https://esriarlington.tpondemand.com/entity/155501-user-indexing).


packages/search/src/users/errors.ts
```diff
@@ -0,0 +1,11 @@
+export class GraphQLError extends Error {
```
robertsteilberg, line 1: I am conflicted over whether this class is necessary. On one hand, we could just throw what the GraphQL API gives us. On the other hand, it would be nice if we abstracted that implementation a bit to the client and gave the client an easier-to-use error message

packages/search/src/users/errors.ts
```diff
@@ -0,0 +1,15 @@
+import { getProp } from "@esri/hub-common";
+
+export class GraphQLError extends Error {
+  constructor(private errors: Error[]) {
+    /* istanbul ignore next */
+    super();
+    Object.setPrototypeOf(this, GraphQLError.prototype);
```
robertsteilberg, line 7: TypeScript cannoto extend certain built-ins, see [this breaking change](https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work). Setting the prototype lets us override `Error` with a custom-formatted error message.

packages/search/src/users/queries.ts
```diff
@@ -0,0 +1,77 @@
+/* Copyright (c) 2021 Environmental Systems Research Institute, Inc.
+ * Apache-2.0 */
+
+import { gql } from "graphql-request";
```
robertsteilberg, line 4: Is this the proper way to definie and export graphql queries or should this be done in a `.graphql` file?

packages/search/src/users/users.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export class UserService {
+  constructor(private portalUrl: string, private userIndex: GraphQLClient) {}
+
+  static create(
+    portalUrl: string,
+    userIndexApi: string,
+    authentication: UserSession
+  ): UserService {
+    const api: GraphQLClient = new GraphQLClient(userIndexApi, {
```
robertsteilberg, line 74: I am creating the client in the static function and passing it directly to the constructor so that we can stub this API object easily in tests

packages/search/test/users/users.test.ts
```diff
@@ -0,0 +1,397 @@
...
+  it("static constructor", () => {
+    const session: UserSession = new UserSession({
+      token: "token"
+    });
+    const service = UserService.create(
+      portalUrl,
+      "https://foo.com/graphql",
+      session
+    );
+    expect(service).toBeTruthy();
```
robertsteilberg, line 31: Not sure how else to test this. We can't access the created GraphQLClient object to check that the token was properly set or anything because it is private.

dbouwman (commented):
The Hub.js docs auto generated directly from the tsdoc/jsdoc style comments in the code.  Also consider adding a guide to with example code

robertsteilberg:
Where should such a guide go? Should it be hardcoded into the `docs` directory somewhere?

robertsteilberg:
verified that this will build + run in opendata-ui

robertsteilberg:
verified that the umd build works as expected

```
const { UserService } = require('./search.umd');
const service = UserService.create(
  "https://devext.arcgis.com",
  'https://ingress.eks.dev.hub.geocloud.com/graphql',
  {
    token: "REDACTED"
  }
);
const resp = service.getSelf().then(resp => console.log(resp));
```
```
$ node test.js
{
  self: {
    username: 'robadmin',
    lastHubSession: 1610981423321,
    firstName: 'Robert',
    lastName: 'Steilberg',
    fullName: 'Robert Steilberg',
    visitsLast30Days: 63,
    visitsLast60Days: 63,
    groups: [],
    followedInitiatives: [],
    registeredEvents: [],
    teams: []
  }
}
```

packages/search/src/users/users.ts
```diff
@@ -0,0 +1,171 @@
...
+      portalUrl: this.portalUrl
+    });
+  }
+
+  /**
+   * Get the current user tied to the token used to instantiate the UserService
+   */
+  getSelf(): Promise<any> {
+    return this.userIndex.request(userSelfQuery);
+  }
```
mjuniper, line 117: Should this be marked `async` like `searchUsers`?

ajturner:
Can you clarify if this version supports polymorphism against either Hub index or Portal API depending on the environment (Hub in Online vs. Enterprise Sites).

robertsteilberg:
This version does not support polymorphism against the Portal API, it is only configured for use against a GraphQL endpoint, which does not exist in Portal. Next steps were to set it up for that but did not have the time to do so.

robertsteilberg:
@mikeringrose and I confirmed that the node libs arer not pulled into the Ember app cc @tomwayson

packages/search/src/users/users.ts
```diff
@@ -0,0 +1,171 @@
...
+export interface IUserSearchOptions {
+  pagingOptions?: IUserPagingOptions;
+  sortingOptions?: ISortingOption[];
+}
+
+/**
+ * Single service that, upon instantiation with proper authentication, exposes three endpoints
+ * for interfacing with the GraphQL API exposed by Hub Profiles API
+ */
+export class UserService {
```
tomwayson, line 79: The `HubService` class is not really useful in the Hub app. It essentially duplicates the Ember service w/o the benefit of DI. Sure you could have the ember service wrap an instance of this class, but that's just means additional code to wire up in `init()`. Furthermore, b/c a user can sign in/out at any point during a session, the Ember service would either still have to pass in `authentication`  each time, or update the instance whenever a user does so. All in this class doesn't benefit Ember.

I realize that the Ember app is not the only client, and that being able to use an instance might benefit other clients, but I think some more thought needs to go into the design of that. Should there be only one instance `const hub = Hub.create(portalUrl, hubApiUrl, isPortal)`. Should it be implemented at a higher level package?

Until we hone that design I propose, at the very least, that Hub.js continue to expose the stateless fns not attached to a class.

tomwayson (changes requested):
I think it's great that you've gotten a prototype of a graphQL client working, but I think that there are a few changes we would need to make in order to make this PR something we could merge and iterate on (if that is the intent).

tomwayson:
A few more questions.

Is `searchUsers()` meant to be used to fetch a single user, or would we need to also add a `getUser()` (or `fetchUser()`) fn?

Is "users" the right term conceptually, or should it be "members"?

packages/search/src/users/users.ts
```diff
@@ -0,0 +1,171 @@
...
+ */
+export interface IHubUser extends IUser {
+  followedInitiatives?: any[];
+  groups?: any[];
+  teams?: any[];
+  registeredEvents?: any[];
+  lastHubSession?: number;
+  visitsLast30Days?: number;
+  visitsLast60Days?: number;
+}
```
mjuniper, line 51: I would love to have a `thumbnailUrl` that is a fully qualified url to the user's thumbnail. We do this on `IHubContent` via `IHubResource`.

ajturner:
> A few more questions.
> 
> Is `searchUsers()` meant to be used to fetch a single user, or would we need to also add a `getUser()` (or `fetchUser()`) fn?
> 
> Is "users" the right term conceptually, or should it be "members"?

For terms, we have used several variations: `User`, `Member`, `Person`. While I'm a strong advocate for consistent domain terminology, we also know that unclear variation from the Portal API can be confusing as well. I also see `addUsersToTeam` and `addUsersToEvent`. So I would be OK if we stay with `Users` in our API while our UI shows Members/People as appropriate in the context. 

For the methods, I agree that we should likely have both `searchUsers()` and `getUser()`. These should be consistent with other Hub.js methods as I started documenting in our internal wiki like this:


Resource | Method | Function | Class
-- | -- | -- | --
Site | Search |   |  
  | Get | getSiteById |  
  | Create | createSite |  
  | Update | updateSite |  
  | Delete |   |  
Page | Search |   |  
  | Get |   |  
  | Create |   |  
  | Update |   |  
  | Delete |   |  
Team | Search |   |  
  | Get |   |  
  | Create | createHubTeam |  
  | Update |   |  
  | Delete |   |  
  | Members | addUsersToTeam |  
Person | Search |   |  
  | Get |   |  
  | Create |   |  
  | Update |   |  
  | Delete |   |  
Content | Search |   |  
  | Get | getContent |  
  | Create |   |  
  | Update |   |  
  | Delete |   |  
  | Download | requestDatasetExport |  
Event | Search | searchEvents |  
  | Get |   |  
  | Create |   |  
  | Update |   |  
  | Delete |   |  
  | Register | registerForEvent, unregisterForEvent, addUsersToEvent |  
Discussion | Search |   |  
  | Get |   |  
  | Create |   |  
  | Update |   |  
  | Delete |   |

packages/search/src/users/users.ts
r00b: This should be implemented in the hub-profiles I think cc @trescube

packages/search/src/users/users.ts
```diff
@@ -0,0 +1,171 @@
...
+   * @param filter {ISearchUsersFilter} - filter object describing which users should be returned
+   * @param options {IUserSearchOptions} - options hash optionally containing IUserPagingOptions for paging
+   *                                       or an array of ISortingOptions for sorting
+   * @returns ICursorSearchResults<IHubUser> that contains the specified page of results along with a next function
+   *          that returns the next page of results or null if there are no more results
+   */
+  async searchUsers(
+    filter: ISearchUsersFilter,
+    options: IUserSearchOptions = {}
+  ): Promise<ICursorSearchResults<IHubUser>> {
```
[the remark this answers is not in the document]
mikeringrose: We just need to change the return type. There won't be a breaking change.

packages/search/src/users/users.ts
mikeringrose: Trailing boolean parameters are an anti-pattern that should be avoided (https://martinfowler.com/bliki/FlagArgument.html). The service class is beneficial and allows for easy mocking and polymorphism. A little more thought just needs to be given to authentication, but it's really solved problem.

dbouwman: Tom's point about wrangling the auth as users sign in / out is something we should work out sooner vs later. Here's the 20,000 ft summary, but really this is the stuff we should be discussing in a wholistic design document.

### Ember Services and Class Instances

Ember services have access to the current session from Torii, which is DI'ed in. When a function on an Ember service is called, it typically delegates to Hub.js/Rest.js function calls, and (ideally) the only logic is to pass in the current session information from Torii. This leaves the Ember service as a lean facade who's main purpose is to inject auth, and makes it very easy to mock the service from w/in Ember tests (i.e. for routes & controllers)

Adding an instance of a class (i.e. `UserService`, as written) into the Ember Service, raises the issue of when in the Ember lifecycle this get's created. If we do it in the `init()` hook, we will never have access to the session, as service initiation happens very early in the app lifecycle and we can't be assured we have a session yet. Also we'd only ever have a session if the user had previously auth'd and had token info in localStorage, or has a platform cookie.

If we do it later... well there really is no "hook" for that... so that really means doing it when a function is called... and that could be done... check if an instance exists, if not, create it. Except in Hub, we support auth'd and un-authed routes... and so if the `UserService` exposes any fn's that can be called w/o a `UserSession` then we're up against the problem of sync'ing the auth state of the app, with the auth state of the instances of these Hub.js sessions. While we could use Ember's "observer" infrastructure, but in general that's not recommended.

The alternative is to drop the `UserSession` from the constructor of the `UserService`, and expose it on the function calls... which then really calls into question the value of using a class in the first place.

There are likely some other solutions to this, but let's hash them out now, vs assuming we'll have an epiphany down the line.

mikeringrose: Cool. Something for us to solve and make better.

tomwayson: Just to clarify, if we were to merge this PR as is we would be _introducing_ the problem that would then need solve and make better.

I've proposed a few small changes to this PR that I think I would be comfortable merging until we've had a chance to do a wholistic design to address the concerns that @ajturner, @dbouwman , and I have raised.

- drop the `UserService` and export it's functions & have them get the context they need from their arguments (something that extends `IHubRequestOptions`
- rename `searchUsers` to `searchUsersFromHub`

Then once we've figured out how "service" classes fit in to this API and how we can have top level search arguments that work for either backend (previously we've talked about using query builders that could output either GraphQL) we can add `hub.searchUsers(userQueryBuilder)` which internally do something like:

```ts
if (!this.isPoral) {
  searchUsersFromHub({ ...this.requestOptions, ...userQueryBuilder.getGraphQlParams() };
} else {
  searchUsersFromPortal({ ...this.requestOptions, ...userQueryBuilder.getPortalQueryParams() };
}
```

I'm even willing to take a stab at making the changes I've proposed above if needed, but I won't be able to get to it this week.

packages/search/src/users/users.ts
trescube: including @thomas-hervey (he's the PE and all feature requests should go thru him).

PR #446, opened 2021-01-06
F/169296 content paging

drspacemanphd:
[169296](https://esriarlington.tpondemand.com/entity/169296-move-pagination-and-sorting-functions-into)

This is an updated PR to address merging the pagination states of multiple search results into a single base64-encoded cursor. In a previous PR, hardcoded `ago` and `hub` properties were being used to curate the exact response the Hub Indexer needed. Instead, to make the main functionality a bit simpler and more generic, I've introduced a function that simply merges page states (each with a designated label), which should allow a search service to retrieve the correct next page of results from multiple sources.

### DIFF
--- a/packages/search/src/util/merge-pagination/merge.ts
+++ b/packages/search/src/util/merge-pagination/merge.ts
@@ -1,3 +1,4 @@
+import { encode } from "base-64";
 import { InvalidPaginationInputError } from "./invalid-pagination-input";
 
 /**
@@ -37,5 +38,5 @@ export function mergePages(pages: IDataPageNextStart[]): string {
     {} as Record<string, number>
   );
 
-  return Buffer.from(JSON.stringify(mergedPages)).toString("base64");
+  return encode(JSON.stringify(mergedPages));
 }
--- a/packages/search/src/util/merge-pagination/merge.ts
+++ b/packages/search/src/util/merge-pagination/merge.ts
@@ -23,9 +23,7 @@ interface IDataPageNextStart {
 export function mergePages(pages: IDataPageNextStart[]): string {
   if (!pages || !Array.isArray(pages)) {
     throw new InvalidPaginationInputError(
-      `Invalid Input Error. Must be array of IDataPageNextStart, received: ${JSON.stringify(
-        pages
-      )}`,
+      `Invalid Input Error. Must be array of IDataPageNextStart, received: ${typeof pages}`,
       pages
     );
   }
--- a/packages/search/src/util/merge-pagination/merge.ts
+++ b/packages/search/src/util/merge-pagination/merge.ts
@@ -1,0 +1,41 @@
+import { InvalidPaginationInputError } from "./invalid-pagination-input";
+
+/**
+ * Interface that describes the name of a datasource (label) and the next start
+ * at which to search pages of data
+ */
+interface IDataPageNextStart {
+  label: string;
+  nextPageStart?: number;
+}
+
+/**
+ * When multiple data sources underly a paginated search, the search implementation should be able to
+ * search at the appropriate starting point for each source for a given paged set of results. This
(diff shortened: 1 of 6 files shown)


dbouwman (commented):
This looks good to me - but would be good to get other reviewers on this sooner vs later. I've got two other PRs that are needed to resolve issues in -UI, so as soon as those can merge I'll cut a release (likely tomorrow) so if we can include this, so much the better

packages/search/src/util/merge-pagination/merge.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const mergedPages: Record<string, number> = pages.reduce(
+    (pageObj: Record<string, number>, page: IDataPageNextStart) => {
+      pageObj[page.label] = page.nextPageStart || 1;
+      return pageObj;
+    },
+    {} as Record<string, number>
+  );
+
+  return Buffer.from(JSON.stringify(mergedPages)).toString("base64");
```
tomwayson, line 40: AFAIK, `Buffer` is not a thing in a browser, so what's the plan there?

tomwayson (changes requested):

packages/search/src/util/merge-pagination/merge.ts
dbouwman: @tomwayson - that also looked a little suss to me... but `test:ci` runs node + chrome and firefox tests, so I assumed it was cool... but yeah - looking in chome devtools... 
image
you are correct - `Buffer` is not a thing unless it's imported from something... 

which raises a few questions re: CI tests, in addition to the problem of how to handle the base64 encoding for this

dbouwman: Since my PR's merged, I'm going to cut a release while this gets resolved

drspacemanphd: @dbouwman if you haven't already please feel free to cut a release.

Looking at the node_modes there appears to be [buffer](https://github.com/feross/buffer) package being brought in. `npm list buffer` in the root shows it's a required dependency of `karma-typescript`.

In a [previous PR](https://github.com/Esri/hub.js/pull/442), I instead specified an external dependency to ensure compatibility with both browser and NodeJS. It was my understanding based on comments that I could refactor to use Buffer for node-specific functionality, but I can easily switch back to the relatively small library.

drspacemanphd: I have updated the PR to use the external library for base64 encoding as shown in the previous PR.

packages/search/src/util/merge-pagination/merge.ts
```diff
@@ -0,0 +1,42 @@
...
+ * encoded cursor string that the implementation can use for the next set of results.
+ *
+ * @param pages an array of page states denoted by their label and the starting point for the next page
+ * @returns a base64-encoded cursor string to use when obtaining next page of results.
+ *
+ */
+export function mergePages(pages: IDataPageNextStart[]): string {
+  if (!pages || !Array.isArray(pages)) {
+    throw new InvalidPaginationInputError(
+      `Invalid Input Error. Must be array of IDataPageNextStart, received: ${JSON.stringify(
```
robertsteilberg, line 26: it might be a bit more concise to do `received: ${typeof pages}`, maybe that would help inform a bit without vomiting a huge object into the log

this might just be personal preference though.

robertsteilberg: what if `pages` is something that `JSON.stringify` cannot handle?

drspacemanphd: Good call. Removed potentially expensive stringify in favor of typeof

robertsteilberg (approved):

tomwayson (approved):

PR #447, opened 2021-01-06
Handle status polling for datasets returning "ready_unknown" status

rgwozdz:
For datasets that don't expose a lastEditDate, the Hub API returns a status of `ready_unknown` when a file is ready on S3 - this means it's "ready" but unknown whether it is up-to-date with respect to the service.

The existing polling method in hub.js does not handle such status correctly.  This PR adds required handling.  The polling method:
* now includes an argument that tracks the existing S3 file's timestamp (if exists)
* now checks for `ready_unknown` status and if found, compares the new file's date to the old file's date to ensure the re-export has finished.  If so, a success event is emitted and polling is cancelled.


drspacemanphd (approved):
Nice job on the quick turn-around.

PR #450, opened 2021-01-08
F/164823 search aggregations

drspacemanphd:
[164823](https://esriarlington.tpondemand.com/entity/164823-move-aggregation-functions-into-hubjs)

This PR introduces interfaces for Aggregations associated with search service results, as well a function to merge aggregations returned from multiple sources such that they can be returned by a single search response, as is the case for content search.


packages/search/src/util/aggregations/merge-aggregations.ts
```diff
@@ -0,0 +1,159 @@
...
+type AggregationMap = Record<string, number>;
+type AggregationResultMap = Record<string, AggregationMap>;
+
+/**
+ * The default merge function for merging aggregations. Simply sums the aggregations
+ * @param aggValueOne the value of one aggregatiom
+ * @param aggValueTwo the value of the second aggregation
+ * @returns the sum of the two values
+ */
+const defaultMergeFunc = (aggValueOne: number, aggValueTwo: number) => {
```
tomwayson, line 39: suggest calling this `sumAggregations`

packages/search/src/util/aggregations/merge-aggregations.ts
```diff
@@ -0,0 +1,159 @@
+/**
+ * Interface for an Aggregation returned from search request,
+ * where label is a value for a fieldname type for which aggregations were requested
+ * and aggValue is the aggregation value
+ */
+interface IAggregation {
+  label: string;
+  aggValue: number;
```
tomwayson, line 8: suggest `value`. To me `aggregation.value` is better than `aggregation.aggValue`.

`

tomwayson (commented):
I like the design (passing in the aggregation function). I left some (entirely optional) naming suggestions.

Again, I can understand when this code would be run on the server (joining results from portal/elastic), but I'm would this code be used in the client? If  not I recommend keeping in those packages and out of Hub.js.

packages/search/src/util/aggregations/merge-aggregations.ts
```diff
@@ -0,0 +1,159 @@
...
+ * Interface for an list of Aggregations requested for a particular fieldname
+ * Ex. { fieldName: 'access', aggregations: [ { public: 2, private: 5, ... } ] }
+ */
+export interface IAggregationResult {
+  fieldName: string;
+  aggregations: IAggregation[];
+}
+
+/**
+ * A function used when to merge the same aggregation across multiple aggregation sources.
```
robertsteilberg, line 21: typo?

packages/search/src/util/aggregations/merge-aggregations.ts
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * @param aggs a list of aggregation result sets from different requests/sources
+ * @param mergeFunction a merge function used to merge aggregation values across result sets
+ * @returns a list of merged aggregations
+ */
+export function mergeAggregations(
+  aggs: IAggregationResult[][] = [],
+  mergeFunction: mergeFunc = defaultMergeFunc
+): IAggregationResult[] {
+  if (!Array.isArray(aggs) || aggs.length === 0) {
```
robertsteilberg, line 57: we shouldn't throw an errorr if aggs isn't an array?
although i guess thiis is a pretty graceful way to handle that case.

packages/search/src/util/aggregations/merge-aggregations.ts
```diff
@@ -0,0 +1,159 @@
...
+    return [];
+  }
+
+  const mergedMap: AggregationResultMap = aggs.reduce(
+    (mergedAggs: AggregationResultMap, aggResultList: IAggregationResult[]) => {
```
robertsteilberg, line 62: can't say i'm a huge fan of having this reduction done via side effects in the helper fn--wouldn't it be more clear/testable to pass in the accumulator, and have the reduction return an object instead of having it modify mergedAggs and returrninig that?

maybe this is just the voice of Dave speaking through me

drspacemanphd: Yeah I definitely recognize the complexity of custom merging of objects at two different prop levels. I mulled on this back and forth when I initially implemented. Ultimately I reasoned that [reducing an array to an object of grouped key-value pairs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#grouping_objects_by_a_property) was common enough to justify. I re-examined this morning and believe I've done a more suitable refactor.

Keep in mind that, while there are reasons we don't include it, lodash has a [function](https://lodash.com/docs/4.17.15#mergeWith) that accomplishes custom object merging

packages/search/src/util/aggregations/merge-aggregations.ts
drspacemanphd: I'm a little torn here as well. Ultimately, I decided that because this function isn't called _directly_ by a client, I did not want to punish the client with an exception if an underlying source that provides aggregations did not format them correctly.

packages/search/src/util/aggregations/merge-aggregations.ts
drspacemanphd: Thanks for catching. This is fixed.

packages/search/src/util/aggregations/merge-aggregations.ts
drspacemanphd: Thanks I have fixed. That makes more sense.

packages/search/src/util/aggregations/merge-aggregations.ts
drspacemanphd: Good call. It's use as a default argument makes `defaultMergeFunc` redundant as a name.

drspacemanphd:
> I like the design (passing in the aggregation function). I left some (entirely optional) naming suggestions.
> 
> Again, I can understand when this code would be run on the server (joining results from portal/elastic), but I'm would this code be used in the client? If not I recommend keeping in those packages and out of Hub.js.

Thanks! While this code is not _directly_ invoked by the client, I envision it being a shared util across different search services, or rather any functionality that must fetch and merge equivalent data from multiple sources. The search services will be invoked client-side.

packages/search/src/util/aggregations/merge-aggregations.ts
robertsteilberg: i'd buy that

robertsteilberg (approved):

tomwayson:
> I envision it being a shared util across different search services, or rather any functionality that must fetch and merge equivalent data from multiple sources.

We don't have any plans to do that that in the client though. I'd still prefer this code to live in hub-indexer where we have an immediate need for it, especially because this introduces a new dependency.

Instead we should be thinking about how the client search API is going to handle search requests to 2 different back end APIs - one using a cursor, and another using paging parameters.

drspacemanphd:
> Instead we should be thinking about how the client search API is going to handle search requests to 2 different back end APIs - one using a cursor, and another using paging parameters.
>

The indexer basically already does this on behalf of the requester. The requester sends a base64-encoded cursor that contains the paging information needed for AGO and our own index.

tomwayson:
What I mean is, in enterprise, we don't use the Hub API - we search against the portal directly. Hub.js needs to provide a single API that hides that from the consumer, the way we do here:

https://github.com/Esri/hub.js/blob/0d138c678a5337feacf6fc13c694d8505f6877e0/packages/content/src/content.ts#L76-L80

drspacemanphd:
That makes sense to me and is part of the epic for unifying search [here](https://esriarlington.tpondemand.com/entity/158225-unified-search-api). But part of this epic is also moving some of these functions so that they can be shared. That's just where I am coming from.

tomwayson:
> But part of this epic is also moving some of these functions so that they can be shared.

I get that but my question is why are we sharing these functions? My current understanding is that they are only used in the indexer.

mikeringrose (approved):

PR #452, opened 2021-01-12
[DOC]: fix links to code and add GH link in header

dbouwman:
As it says - docs are already published w/ these changes


dbouwman:
tbh - I'm not clear why this has the v6.22.2 commit in it, as that's in master O_o

image

Issue #430, opened 2020-11-30
Silent failing of addAnnotations with "old" version of Hub Annotations


lisastaehli:
We have received a bug report for Urban where a user was unable to add an annotation. We identified that the problem was that the ArcGIS Online organization of the user had an older version of the Hub Annotations layer where public annotations (Author = NULL) was not yet allowed by the schema (see error message below).

![Old_Annotations_Layer (1)](https://user-images.githubusercontent.com/27951811/100595183-dfa0c400-32fa-11eb-9819-abfa9f73da26.png)

Of course there is an obvious workaround to this (probably) edge case: the user has deleted the old annotations layer (which was anyway empty) & added a new one with the current schema. 

We would like to start a discussion on how to handle old versions of Hub Annotations layers with HubJS. In my opinion, there would be a few things that could be nice, e.g. notify the user if they are trying to connect to a feature service with an outdated schema and/or (automatically) migrate in this case. 

Please note that we are currently still on version `2.2.4` of Hub JS. If this is handled in a newer version, please let me know :)

lisastaehli:
@tomwayson any thoughts on this :)?

ajturner:
two approaches to consider / provide estimates @tomwayson 
- adding a migration function (I recall there was one before)
- throwing exception when creating annotation fails.

tomwayson:
I'm looking into a migration script.

In the mean time, a potential quick fix for this would be to see if the existing service has zero records, and if so, just remove it and call [createAnnotationService()](https://esri.github.io/hub.js/api/annotations/createAnnotationService/).

tomwayson:
I see you already did that. 🤦

tomwayson:
FYI - this is the underlying issue regarding how the errors are swallowed:

https://github.com/Esri/arcgis-rest-js/issues/790

I think I'll have the rest-js functions throw an error if all records have an error. We could go a step further and have `addAnnotations()` and `updateAnnotations()` throw an error if _any_ records failed.

Let me know if you think the latter is necessary.

lisastaehli:
@tomwayson thanks for the update.

I think all having errors serves our purposes as an outdated schema will cause all of them to fail :)

tomwayson:
@lisastaehli this has been released as `6.22.3`

PR #453, opened 2021-01-19
fix(annotations): throw error when all adds or updates fail

tomwayson:
affects: @esri/hub-annotations

ISSUES CLOSED: #430

### DIFF
--- a/packages/annotations/src/add.ts
+++ b/packages/annotations/src/add.ts
@@ -8,6 +8,7 @@ import {
 
 import { IFeature } from "@esri/arcgis-rest-types";
 import { IEditFeatureResult } from "@esri/arcgis-rest-feature-layer";
+import { checkResults, IEditFeatureErrorResponse } from "./util";
 
 export interface IAnnoFeature extends IFeature {
   attributes: {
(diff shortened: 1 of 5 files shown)


tomwayson:
@lisastaehli once this is merged the add and update annotation functions will throw an error if all results failed.

packages/annotations/src/add.ts
```diff
@@ -43,14 +44,19 @@ export interface IAddAnnotationsOptions extends IAddFeaturesOptions {
...
     requestOptions.features.forEach(anno => enrichAnnotation(anno));
   }
 
-  return addFeatures(requestOptions);
+  return addFeatures(requestOptions).then(response => {
+    return {
+      ...response,
+      addResults: checkResults(response.addResults)
+    };
+  });
```
mikeringrose, line 59: Does this align with the return type?
```
Promise<{
  addResults?: Array<IEditFeatureErrorResponse | IEditFeatureResult>;
}>
```

tomwayson: yep, in fact TS complained when I made a typo and it wasn't aligned.

mikeringrose: even though it mentions nothing of the destructured response object. strange.

mikeringrose: never mind. i didn't see the destructuring at first.

mikeringrose: the destructuring doesn't seem necessary since your always overriding, but it's already merged.

tomwayson: I did the destructuring b/c thought there was additional properties in the response. looking at the [API docs](https://developers.arcgis.com/rest/services-reference/add-features.htm) there may not be.

PR #454, opened 2021-01-20
f/165476 added target change to allow locking of private non-enterpri…

drspacemanphd:
[167921](https://esriarlington.tpondemand.com/entity/167921-restricting-exports-for-portal-downloads-add)

For private downloads on non-enterprise sites, the `createReplica` API may not return the most up-to-date data if the dataset had been updated within the last 10 mins. This PR is the first step to introduce the concept of "locked" to downloads. If a download file does not exist for private non-enterprise dataset, and that dataset has been updated in the last 10 mins, it will return a status of `locked`. If an _up-to-date_ download file exists for the dataset, even if the data was updated in the last 10 mins, it still returns a `ready` status. Finally, if a _stale_ download file exists and the dataset was recently updated, it returns a `stale_locked` status.

I should note that while this is not necessarily a breaking change with respect to the UI, this introduces a new status that the UI will not know how to handle.

The following test cases were performed:
```
For test cases involving staleness, it may take 1-2 mins for feature service calls to be cleared from cache

Private Feature Service, single layer - https://downloadsidebar-qa-pre-a-hub.hubqa.arcgis.com/datasets/f7195f4598ae4864ba79d9db6c89a921
- File can be downloaded when it does not exist and the dataset was not recently updated
- File cannot be downloaded when it does not exist and the dataset was recently updated
- File can be downloaded when it is up to date
- File can be re-generated when it is stale and the dataset not recently updated
- File cannot be re-generated (but existing file can be downloaded) when it is stale and the dataset was recently updated

Private Feature Service, multiple layers - https://downloadsidebar-qa-pre-a-hub.hubqa.arcgis.com/datasets/b2c6412da0cb4e0d9e46ca87467ec360
- File can be downloaded when it does not exist and the dataset was not recently updated
- File cannot be downloaded when it does not exist and the dataset was recently updated
- File can be downloaded when it is up to date
- File can be re-generated when it is stale and the dataset not recently updated
- File cannot be re-generated (but existing file can be downloaded) when it is stale and the dataset was recently updated
```


packages/downloads/src/portal/portal-request-download-metadata.ts
```diff
@@ -134,26 +144,44 @@ function extractLastEditDate(layers: ILayerDefinition[]) {
...
+
+  let status;
+  if (recentlyUpdated && outOfDate) {
+    status = "stale_locked";
+  } else if (outOfDate) {
+    status = "stale";
+  } else {
+    status = "ready";
+  }
+
```
rgwozdz, line 181: I'd move the determination of status to it's own function, will make this more concise and easier to read.

packages/downloads/src/utils.ts
```diff
@@ -49,3 +51,17 @@ export function composeDownloadId(params: IDownloadIdParams): string {
...
+ * @private
+ */
+export function isRecentlyUpdated(
+  target: DownloadTarget,
+  lastEditDate: number
+): boolean {
+  return (
+    target === "portal" &&
+    lastEditDate &&
+    new Date().getTime() - lastEditDate <= 1000 * 60 * 10
```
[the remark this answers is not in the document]
drspacemanphd: Good call, I've converted this to a const at the top of the file

packages/downloads/test/poll-download-metadata.test.ts
```diff
@@ -11,7 +11,7 @@ describe("pollDownloadMetadata", () => {
       spyOn(mockEventEmitter, "emit");
 
       spyOn(hubPoller, "hubPollDownloadMetadata").and.returnValue(
-        new Promise((resolve, reject) => {
+        new Promise((resolve, _reject) => {
```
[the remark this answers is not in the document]
drspacemanphd: Some linters will fail if a supplied parameter is unused, but this can sometimes make it awkward for some callbacks like (err, data) in node and (resolve, reject) in promises. The underscore prefix denotes that you acknowledge that a param is being supplied but you will not use it. This is also done in a few places in the indexer!

packages/downloads/src/portal/portal-request-download-metadata.ts
drspacemanphd: Good call. This is done.

rgwozdz (approved):

packages/downloads/src/utils.ts
```diff
@@ -49,3 +53,17 @@ export function composeDownloadId(params: IDownloadIdParams): string {
...
+
+/**
+ * @private
+ */
+export function isRecentlyUpdated(
+  target: DownloadTarget,
+  lastEditDate: number
+): boolean {
+  return (
+    target === "portal" &&
```
rgwozdz, line 65: If this util is only ever used for portal downloads, maybe it should be in the `/portal` directory?

rgwozdz (approved):
Just one question about util location.

packages/downloads/src/utils.ts
drspacemanphd: I can go with that. This is now in a `utils` file in `/portal`

PR #456, opened 2021-01-21
Domains fixes

robertsteilberg:
Changes all usages of `domainEntry.domain` to `domainEntry.hostname`


dbouwman (approved):
Fancy. Pls note that this is a BREAKING change... which is le fine, but we need to ensure we do a MAJOR bump when this goes out... and also verify that it's all happy in Solutions (not sure why it would not be... but... breaking changes have a habit of breaking things)

Issue #417, opened 2020-10-29
Would like to be able to convert IModel into Content


andrewctate:
What if `getContent` could accept an `IModel` as well as an item id? It would just have to be smart enough not to request things from the API that it didn't have to, but still get the goodies and extras that can't be gleaned from the IModel.

tomwayson:
This is a good idea. I've thought about having it accept an `IItem`. 

Do you think it's common scenario where we have the `item` _and_ the `data`?

tomwayson:
related to #415

drewctate:
@tomwayson 
> Do you think it's common scenario where we have the item and the data before we want to call getContent()
That is nearly every scenario I've run into so far. But, I recognize that I've focused on initiative and solution templates so it would be wise to do a broader poll.

andrewctate:
[POA]
- Change `getContent` call signature to accept an `IItem` or an `IModel` `getContent(idOrObj: IModel | IItem | string...)` [here](https://github.com/Esri/hub.js/blob/master/packages/content/src/content.ts#L71)
- Extract item and data objects from `idOrObj` as available
- If item object is available, always use `getContentFromPortal`
- change `getContentFromPortal`
     - relax call signature to allow an item to be passed instead of an id: `getContentFromPortal(idOrItem: string | IItem,  requestOptions?: IHubRequestOptions)`
     - use item object when provided instead of requesting a new one from Portal API
- If an `IModel` was provided, and `shouldFetchData` is true, register a function that resolves to a copy of the provided `model.data` hash instead of `getContentData` [here](https://github.com/Esri/hub.js/blob/master/packages/content/src/content.ts#L93)

cc: @tomwayson

mikeringrose:
> [POA]
> 
> * Change `getContent` call signature to accept an `IItem` or an `IModel` `getContent(idOrObj: IModel | IItem | string...)` [here](https://github.com/Esri/hub.js/blob/master/packages/content/src/content.ts#L71)
> * Extract item and data objects from `idOrObj` as available
> * If item object is available, always use `getContentFromPortal`
> * change `getContentFromPortal`
>   
>   * relax call signature to allow an item to be passed instead of an id: `getContentFromPortal(idOrItem: string | IItem,  requestOptions?: IHubRequestOptions)`
>   * use item object when provided instead of requesting a new one from Portal API
[...]

PR #457, opened 2021-01-22
feat(getContent): getContent now accepts an IModel object in addition…

andrewctate:
… to an identifier string

affects: @esri/hub-content

ISSUES CLOSED: #417

### DIFF
--- a/packages/content/src/content.ts
+++ b/packages/content/src/content.ts
@@ -11,6 +11,7 @@ import {
   HubType,
   IHubContent,
   IHubRequestOptions,
+  IModel,
   includes
 } from "@esri/hub-common";
 import { IGetContentOptions, getContentFromHub } from "./hub";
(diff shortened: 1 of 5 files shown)


packages/content/src/content.ts
```diff
@@ -62,17 +63,35 @@ function getContentData(
...
+ * @param identifier - Hub API slug ({orgKey}::{title-as-slug} or {title-as-slug})
  * or record id ((itemId}_{layerId} or {itemId})
  * @param options - request options that may include authentication
  */
-export function getContent(
-  identifier: string,
-  options?: IGetContentOptions
-): Promise<IHubContent> {
-  // first fetch and format the content from the Hub or portal API
+function getContentById(identifier: string, options?: IGetContentOptions) {
```
andrewctate, line 92: We could export this function now, or wait till the next major version bump.

tomwayson (approved):
Nice work @andrewctate 

I don't see the need to export anything else until there's a need for it.

packages/content/src/content.ts
```diff
@@ -86,21 +105,31 @@ export function getContent(
...
+}
+
+/**
+ * Get content either from an IModel or an ID.
+ * @param idOrModel - An IModel (with our without data), or Hub API slug ({orgKey}::{title-as-slug} or {title-as-slug})
+ * or record id ((itemId}_{layerId} or {itemId})
+ * @param options - request options that may include authentication
+ */
+export function getContent(
+  idOrModel: string | IModel,
```
andrewctate, line 118: Could scope this down to just an `IModel` for next major version bump if we go further down the 2-path route @mikeringrose was talking about in the issue.

PR #458, opened 2021-01-25
fix(@esri/hub-downloads): add support of ready_unknown status for portal

rgwozdz:
* Adds a `ready_unknown` status for portal downloads
* Puts the portal export-complete handler in its own file
* Puts the function for acquiring exports folder id in its own file.
* Adds tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -554,71 +554,6 @@
 				}
 			}
 		},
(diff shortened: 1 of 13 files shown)


drspacemanphd (approved):
Thanks for breaking out the logic. Definitely makes a bit more readable

PR #461, opened 2021-02-02
Revert "feat(sites): enable the new download sidebar capability by default"

haoliangyu:
Reverts Esri/hub.js#460

We need to remove the capability from the default list due to an issue in the new download UI.


drspacemanphd (approved):
Are there tests for this?

PR #462, opened 2021-02-02
feat: add contentStatus to datasetToItem

mjuniper:
affects: @esri/hub-content

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -554,71 +554,6 @@
 				}
 			}
 		},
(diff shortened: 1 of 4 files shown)


tomwayson (approved):
Sorry, I now realize this is a property of an `item`, and that in `getContentFromPorta()` it's handled by https://github.com/Esri/hub.js/blob/4e9b21ec831113859919d55da58496a0532939e9/packages/content/src/portal.ts#L143

PR #463, opened 2021-02-03
S/fix release code

haoliangyu:
This PR restores the mistakenly edited release code and updates the `gh-release` to latest. FYI, these are the breaking changes for the gh-release update:

```
BREAKING CHANGES:

    Remove support for Github basic (password) auth.
    Add support for oAuth Device flow tokens.
    Existing tokens should still work fine. Basic auth generated Personal Access tokens which will work as long as you don't revoke them.
    Updated dependencies bring with it newer Node.js requirements. Run LTS or Latest.
    The --endpoint flag only works on Github Enterprise now.
    Github enterprise still uses basic auth, since it does not support the new device flow yet.
```

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,6 @@
 name: build-test
 
-on: 
+on:
   push:
     branches:
       - master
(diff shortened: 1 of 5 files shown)


mjuniper (approved):
Thank you!

tomwayson:
Thanks for including those breaking changes. I guess as long as they work for the next person to run the release (sounds like @mjuniper), we're good?

tomwayson (approved):

PR #474, opened 2021-02-25
Ensure OperationStack retains rootCause

dbouwman:
Better handling of the `rootCause` optional param in `OperationError` constructor. 

If a `rootCause` is passed in, then we use it's `.stack` vs creating a new one b/c that will make the error look like it was thrown inside the OperationError constructor, which it is not.

Also updates the Operation Stack guide w/ example outputs


benstoltz (approved):
Very nice.

PR #479, opened 2021-03-03
feat(hub-discussions): adds Discussions package to Hub.js

cpgruber:
Adds `discussions` package to monorepo, which will serve as the interface layer in front of the [Discussions API](https://devtopia.esri.com/dc/hub-discussions) and ultimately replace hub-annotations (TBDeprecated)


ajturner:
✋Let me know when it's appropriate to provide some questions or comments on this PR.

cpgruber:
> ✋Let me know when it's appropriate to provide some questions or comments on this PR.

@ajturner let em fly

packages/discussions/src/posts.ts
```diff
@@ -0,0 +1,193 @@
...
+    } = options as IDeleteChannelPostOptions;
+    url = `/channels/${channelId}` + url;
+  }
+  options.method = "DELETE";
+  return request(url, options);
+}
+
+/**
+ * update post
+ * NOTE: this method only updates a post's title and/or body
```
ajturner, line 131: This is a good limit to note.

what are your plans for adding update support to geometry or other post attributes?

packages/discussions/src/posts.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @export
+ * @param {(IUpdatePostOptions | IUpdateChannelPostOptions)} options
+ * @return {*}  {Promise<IPostDTO>}
+ */
+export function updatePost(
```
ajturner, line 137: Should there be an `updateReply` and `updateReaction` as well? Or would `updatePost` also work for those ID?

packages/discussions/src/utils/posts.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Utility that parses a discussion URI string into its component parts
+ *
+ * @export
+ * @param {string} discussion
+ * @return {*}  {IDiscussionParams}
+ */
+export function parseDiscussionURI(discussion: string): IDiscussionParams {
+  // NOTE: cannot use new URL(discussion) to parse uri bc difference in node and browser implementation
```
ajturner, line 12: It might help other developers to provide some examples here of types of Discussion URI. 
I recall you determined it was better to use a namespaced URI (arcgis://) rather than a Web URL .

ajturner: explicit parsing here makes sense because it's simple. 

if it became more complex or we allowed unique URI it might be worth considering a library for isomorphic URI parsing (e.g. [url-parse](https://www.npmjs.com/package/url-parse))

ajturner (commented):
Thanks for being open to feedback. I've added some notes based discussions with @mikeringrose  & @dbouwman for consistency in Hub.js, and some other recommendations considering longer-term maintenance.

packages/discussions/README.md
```diff
@@ -0,0 +1,85 @@
...
+//               //     attributes: {description: "Great place!", ...}
+//               //   }],
+//               //   included: [{
+//               //     id: "User1",
+//               //     type: "users",
+//               //     attributes: { firstName: "User", lastName: "Name", ...}
+//               //   }]
+//               // }
+//           });
+//     });
```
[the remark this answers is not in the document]
cpgruber: Not relevant -- will replace

packages/discussions/src/channels.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * find channel
+ *
+ * @export
+ * @param {IFindChannelOptions} options
+ * @return {*}  {Promise<IChannelDTO>}
+ */
+export function findChannel(
```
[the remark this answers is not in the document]
cpgruber: Consistency++ Semantics++, will change to `get*`

packages/discussions/src/utils/request.ts
```diff
@@ -0,0 +1,73 @@
+import { IRequestOptions } from "../types";
+
+/**
+ * returns Promise that resolves token to use in Discussions API requests
+ *
+ * @export
+ * @param {IRequestOptions} options
+ * @return {*}  {Promise<string>}
+ */
+export function authenticateRequest(options: IRequestOptions): Promise<string> {
```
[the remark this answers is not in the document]
cpgruber: TBH I didn't give much thought about the location of this method. I assumed this was the only place that needed to use platform authentication to make an off-platform request. Is that not the case?

packages/discussions/src/posts.ts
cpgruber: `updatePost` works for replies. For reactions, it makes sense to me that `create` and `delete` are the only mutations.

packages/discussions/src/utils/posts.ts
cpgruber: Thank you! When I moved this method from the API repo to here I realized the difference in browser and node `URL` implementations, and came up short in finding a polyfill. I would much prefer that over relying on my limited regex ability

packages/discussions/src/posts.ts
cpgruber: There are separate methods for updating a post's status (only allowed by moderators) and its channel/sharing. This method could support geometry and discussion/target in just a few LOC; it would be beneficial to discuss the business logic about which attributes are updatable and by whom.

packages/discussions/src/posts.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+export function searchPosts(
+  options: IQueryPostsOptions | IQueryChannelPostsOptions
+): Promise<INestPagination<IPostDTO>> {
+  let url = `/posts`;
+  if (options.params.hasOwnProperty("channelId")) {
+    const {
+      params: { channelId }
+    } = options as IQueryChannelPostsOptions;
+    url = `/channels/${channelId}` + url;
```
mikeringrose, line 39: Why just one method? Why not a method `searchPost` and `searchChannel`? It allows code to be more explicit and you remove the `if` condition and you force validation by parameters.

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,831 @@
...
+ */
+export enum PostReaction {
+  THUMBS_UP = "thumbs_up",
+  THUMBS_DOWN = "thumbs_down",
+  THINKING = "thinking",
+  HEART = "heart",
+  ONE_HUNDRED = "one_hundred",
+  SAD = "sad",
+  LAUGH = "laugh",
+  SURPRISED = "surprised"
```
mikeringrose, line 29: since these are comments, just use the UTF8 emoji character 👍, 👎, etc.

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,831 @@
...
+
+/**
+ * representation of post entity
+ *
+ * @export
+ * @interface IPostDTO
+ * @extends {IWithAuthor}
+ * @extends {IWithTimestamps}
+ */
+export interface IPostDTO extends IWithAuthor, IWithTimestamps {
```
mikeringrose, line 275: I'd refrain from naming with a DTO suffix.

packages/discussions/src/utils/channels.ts
```diff
@@ -0,0 +1,147 @@
...
+      []
+    );
+    const method = strict ? "every" : "some";
+    return sharedGroups[method](
+      group => eligibleUserGroups.indexOf(group) > -1
+    );
+  };
+}
+
+function isChannelOrgMember(channel: IChannelDTO, user: IUser): boolean {
```
mikeringrose, line 24: If channel was simply an object you could simply do:
```
channel.isOrgMember(user)
```

packages/discussions/src/posts.ts
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * search posts
+ *
+ * @export
+ * @param {(IQueryPostsOptions | IQueryChannelPostsOptions)} options
+ * @return {*}  {Promise<INestPagination<IPostDTO>>}
+ */
+export function searchPosts(
+  options: IQueryPostsOptions | IQueryChannelPostsOptions
+): Promise<INestPagination<IPostDTO>> {
```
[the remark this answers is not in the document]
dbouwman: I agree w/ @mikeringrose  on this... from what I can tell, we should be able to drop `DTO` from everything that has it... Using `Rename Symbol` in vscode, this should be pretty easy.

packages/discussions/src/posts.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+export function createReply(
+  options: ICreateReplyOptions | ICreateChannelReplyOptions
+): Promise<IPostDTO> {
+  let url = `/posts/${options.params.postId}/reply`;
+  if (options.params.hasOwnProperty("channelId")) {
+    const {
+      params: { channelId }
+    } = options as ICreateChannelReplyOptions;
+    url = `/channels/${channelId}` + url;
```
dbouwman, line 81: Since you're using string templates, you can do 
```js
url = `/channel/${channelId}${url}`;
```

packages/discussions/src/posts.ts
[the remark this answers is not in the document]
cpgruber: agreed, `*DTO` has been removed. For pagination, what's the gold standard? I've found [`IPagedResponse`](https://esri.github.io/arcgis-rest-js/api/types/IPagedResponse/) and [`IPagingParams`](https://esri.github.io/arcgis-rest-js/api/types/IPagingParams/), but I've also seen other interfaces like https://github.com/Esri/hub.js/blob/master/packages/search/src/types/common.ts#L38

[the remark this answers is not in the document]
cpgruber: Also, would it be appropriate for me to leverage [SortDirection](https://github.com/Esri/hub.js/blob/master/packages/search/src/types/common.ts#L12) from hub-search in my query DTO's?

packages/discussions/src/posts.ts
cpgruber: So this method is searching posts; `IQueryChannelPostsOptions` scopes the search to a single channel, while `IQueryPostsOptions` can search many channels. Does that change your opinion?

Otherwise, I thought having two methods, `searchPosts` and `searchChannelPosts`, to be redundant and consolidated to a single method for the convenience of dev users.

packages/discussions/src/types.ts
cpgruber: I'm sure it would work just fine, but since these values are ultimately persisted in the database I figure using human-readable string versions would eliminate any potential for things to go wrong if and when we choose to optimize by migrating these string values to numeric

packages/discussions/src/posts.ts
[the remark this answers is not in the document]
drspacemanphd: I think so @cpgruber

packages/discussions/src/utils/request.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const opts: RequestInit = {
+    headers,
+    method: options.method || "GET",
+    mode: options.mode || "cors",
+    cache: options.cache || "no-cache",
+    credentials: options.credentials || "include"
+  };
+
+  const apiBase = options.apiBaseUrl || "http://localhost/api/v1";
```
tomwayson, line 55: do we really want to default to localhost? is `/api/v1` even valid?

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,839 @@
...
+// request options
+
+/**
+ * options for making requests against Discussion API
+ *
+ * @export
+ * @interface IRequestOptions
+ * @extends {RequestInit}
+ */
+export interface IRequestOptions extends RequestInit {
```
tomwayson, line 499: I'm not super keen on how this `IRequestOptions` has a lot of overlap w/, but is also inconsistent w/ both:
- https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/
- and https://esri.github.io/hub.js/api/common/IHubRequestOptions/

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,839 @@
...
+  authentication?: IAuthenticationManager;
+  token?: string;
+  apiBaseUrl?: string;
```
tomwayson, line 502: How is this different than `IHubRequestOptions.hubApiUrl`?

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,839 @@
...
+  params?: {
```
tomwayson, line 503: This works different than https://esri.github.io/arcgis-rest-js/api/request/IRequestOptions/#params which is just a set of key/value pairs that are turned into query or body params based on the method.

packages/discussions/src/utils/request.ts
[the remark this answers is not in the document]
tomwayson: We already have:

https://github.com/Esri/hub.js/blob/f35b1a0a868916bd07e1dfd84cb084bc2c876267/packages/common/src/request.ts#L32

and [a few places where we append the `Authorization` header](https://github.com/Esri/hub.js/search?q=Authorization), although those seem to be missing the "Bearer " prefix to the token. I have some vague recollection of a discussion between John Gravois and @rgwozdz regarding that at some point.


I think that the request and auth logic should be centralized in hub-common.

packages/discussions/test/reactions.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+describe("reactions", () => {
+  let requestSpy: any;
+  const response = new Response("ok", { status: 200 });
+
+  beforeEach(() => {
+    requestSpy = spyOn(req, "request").and.returnValue(
+      new Promise(resolve => {
+        resolve(response);
+      })
```
tomwayson, line 13: There's > 5k loc in this PR, this can save 2 of them:

```suggestion
      Promise.resolve(response);
```

tomwayson (changes requested):
Impressive work @cpgruber!

Please make at least the minor changes I suggested for the package build scripts in order to catch up to recently merged PRs.

Beyond that, I think the package's API is very good other than my comments about the low level request/auth inconsistencies with the rest of Hub.js. I'd really like to see some convergence on that front. Let's discuss if/how we could do that before you merge this.

tomwayson:
Also, I strongly prefer how your tests stub `request()` instead of how other packages fetch mock everything. Nicely done.

packages/discussions/src/utils/request.ts
cpgruber: suppose this could be a required param with no default

packages/discussions/src/utils/platform.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+/**
+ * Utility that checks if a user is a portal org admin (default role)
+ *
+ * @export
+ * @param {IUser} user
+ * @return {*}  {boolean}
+ */
+export function isOrgAdmin(user: IUser): boolean {
```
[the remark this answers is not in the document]
cpgruber: That method returns a promise though while my user is already resolved. It would be nice if those rest-js helpers decoupled their argument resolution from the synchronous methods that perform the check/validation

packages/discussions/src/types.ts
cpgruber: maybe `discussionsApiUrl` is a more appropriate name?

packages/discussions/src/utils/platform.ts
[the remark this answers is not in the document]
tomwayson: ah. would be good to have a comment here to that effect

packages/discussions/src/utils/request.ts
tomwayson: I think in other places we've defaulted to the prod URL, but I've worried about that leading to subtle bugs.

packages/discussions/src/types.ts
tomwayson: why can't you use `hubApiUrl`?

packages/discussions/src/utils/request.ts
cpgruber: we don't have a prod url at the moment is the only issue here, so until we do we can leave it a required param

cpgruber:
This has been published as [`@esri/hub-discussions@1.0.0-beta.0`](https://www.npmjs.com/package/@esri/hub-discussions). @tomwayson and I have resolved that this branch will stay open until the API is moved to a QA or Prod environment. Until then the package beta versions of the package will be published outside of lerna.

packages/discussions/src/utils/platform.ts
```diff
@@ -27,6 +27,9 @@ export function reduceByGroupMembership(
  * @param {IUser} user
  * @return {*}  {boolean}
  */
+// NOTE: this is not the same as @esri/arcgis-rest-portal isOrgAdmin,
+// which first resolves `user` from `IUserRequestOptions` to make this determination
+// https://github.com/Esri/arcgis-rest-js/blob/7ab072184f89dcb35367518101ee4abeb5a9d112/packages/arcgis-rest-portal/src/sharing/helpers.ts#L45
```
tomwayson, line 32: thanks also for this comment. I've opened an issue for this https://github.com/Esri/arcgis-rest-js/issues/818

tomwayson (approved):
Thanks for all the changes @cpgruber.

The my only remaining concern is `hubApiUrl`. I probably wasn't clear before, but I only think we should use that if it ends up working _exactly_ like how it works in other packages (i.e. points to `https://hub.arcgis.com/api/v3` in prod). If that's what it ends up being, then what you have here is great, but if it ends up being something different (like `https://hub.arcgis.com/api/discussions/v1`, I wonder if we should be using a different property name.

Otherwise, this LGTM (for whenever we decide to merge it).

ajturner:
@cpgruber I'm wondering what the next steps are to merge this PR?

cpgruber:
> @cpgruber I'm wondering what the next steps are to merge this PR?

[We decided to wait for a QA or Production version of the API to be released](https://github.com/Esri/hub.js/pull/479#issuecomment-804027255), which we now have. Also we need to decide how to handle the [url convention](https://github.com/Esri/hub.js/pull/479#pullrequestreview-617652854). 

And admittedly I have been distracted with other work...

packages/discussions/src/request.ts
```diff
@@ -7,10 +7,17 @@ import { apiRequest, authenticateRequest } from "./utils/request";
...
  * @param {string} url
- * @param {IRequestOptions} options
+ * @param {IHubRequestOptions} options
  * @return {*}  {Promise<T>}
  */
-export function request<T>(url: string, options: IRequestOptions): Promise<T> {
+// NOTE: feasibly this could be replaced with @esi/hub-common hubApiRequest,
+// if that method didn't prepend `/api/v3` to the supplied path. Additionally,
+// there is the difference that hubApiRequest sets Authorization header without `Bearer`
+// https://github.com/Esri/hub.js/blob/f35b1a0a868916bd07e1dfd84cb084bc2c876267/packages/common/src/request.ts#L62
```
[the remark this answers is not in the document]
ajturner: @tomwayson @cpgruber   I suggest we should have a root _Hub API_ endpoint. Within that we can have paths to specific _microservices_ corresponding with modules.

Within these, each could have its own version - which is determine in the module. 
For example, we could have a directory of services (not saying these exist today, but would be good convention)
`/api/:serviceName/:serviceVersion`

So a Module would use the root `hubApiUrl` and then append both `serviceName = 'discussions'` and `serviceVersion = 'v1'` or something similar.

Some examples
- `/api/v3` (is really _content_, so a new version would be `/api/content/v4`)
- `/api/discussions`
- `/api/domains`
- `/api/downloads`
- `/api/community`

/cc @rweber-esri @rgwozdz @thomas-hervey for consideration how we namespace + deploy our services.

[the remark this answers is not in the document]
cpgruber: @ajturner I think that's a reasonable solution, and I think in the meantime we can live with the exceptions made in the discussions package (mostly, not using the common `request` method). I've updated the discussions request method to default to the QA API deployment. 

@tomwayson, I think I've worked out the majority of the bumps in the beta versions I've pushed out, and we're ready to merge and let lerna take over (knowing that I could continue to publish beta versions as needed).

packages/discussions/src/types.ts
```diff
@@ -0,0 +1,787 @@
...
+
+/**
+ * dto for creating a channel
+ *
+ * @export
+ * @interface ICreateChannel
+ * @extends {IWithSettings}
+ * @extends {IWithSharing}
+ */
+export interface ICreateChannel extends IWithSettings, IWithSharing {}
```
cpgruber, line 421: should be Partial<IWithSettings>

packages/discussions/README.md
```diff
@@ -0,0 +1,72 @@
...
+[coverage-url]: https://codecov.io/gh/Esri/hub.js
+
+# THIS PACKAGE IS ACTIVELY UNDER DEVELOPMENT AND IS NOT TO BE USED IN A PRODUCTION ENVIRONMENT
+# @esri/hub-discussions
+
+> Module to interact with ArcGIS Hub Discussions API in Node.js and modern browsers.
+
+### Example
+
+```bash
```
ajturner, line 22: @cpgruber  can you add some more examples of using the API with specifics usages in the response. We are trying out [RDD](https://tom.preston-werner.com/2010/08/23/readme-driven-development.html) to design/update the rest of Hub.js to consider the DX (Developer Experience) making it easy to pickup + use the library by Hub + non-Hub engineers.

/cc @tomwayson @dbouwman @rweber-esri

dbouwman: Examples in the function jsdoc will appear in the doc w/ the function... here so it's a great way to show simple usage...
image

Another good idea is to add a Guide into the docs...  guides are a great place to show how to do common things... i.e. 
image

To add a Guide, add a markdown doc in `/docs/src/guides`... at the top of that file, add an info block like

```
---
title: Working with Sites and Pages.
navTitle: Sites and Pages
description: Learn how to work with Sites, Pages and Domains using @esri/hub.js.
order: 20
group: 2-concepts
---
```

followed by the content.

Then run `npm run docs:serve` and it will fire up the doc site w/ a watcher - I don't think it will live-reload, but it will rebuild the doc a you save files

PR #480, opened 2021-03-05
feat(content): add date enrichments to content

mjuniper:

https://devtopia.esri.com/dc/hub/issues/28

Feedback welcome, my ts-fu is still weak.


mjuniper:
When I run the tests locally it reports 100% code coverage. Anybody have any suggestions?

packages/content/src/content.ts
```diff
@@ -62,6 +66,97 @@ function getContentData(
...
+export function _enrichDates(content: IHubContent): IHubContent {
+  const newContent = cloneObject(content);
+
+  // updateFrequency:
+  const updatedFrequencyValue = getValueFromMetadata(
+    newContent,
+    "updateFrequency"
+  );
+  if (updatedFrequencyValue) {
+    const updateFrequencyMap = {
```
dbouwman, line 113: Consider making the value an exported type that just is a list of the possible strings... that will allow consumers to know all the possible values and setup i18n for them all

packages/content/src/content.ts
```diff
@@ -76,11 +171,13 @@ function enrichContent(content: IHubContent, options?: IGetContentOptions) {
...
-    return content;
-  }
-  return _fetchContentProperties(propertiesToFetch, content, options);
+
+  const fetchContentPropertiesPromise: Promise<IHubContent> =
+    Object.keys(propertiesToFetch).length === 0
+      ? Promise.resolve(content)
+      : _fetchContentProperties(propertiesToFetch, content, options);
+
+  return fetchContentPropertiesPromise.then(_enrichDates);
```
dbouwman, line 179: this fn does not have a return type... not sure how the lint/ts compiler let that slide... 🤷‍♂️

packages/content/src/content.ts
```diff
@@ -12,8 +12,12 @@ import {
   IHubContent,
   IHubRequestOptions,
   IModel,
-  includes
+  includes,
+  cloneObject,
+  getProp,
+  stringToBlob
 } from "@esri/hub-common";
+import { NodeStringDecoder } from "string_decoder";
```
dbouwman: wat

packages/content/src/content.ts
andrewctate: Not sure I understand your comment @dbouwman

packages/content/src/content.ts
```diff
@@ -62,12 +65,121 @@ function getContentData(
...
+  // updatedDate is already set to item.modified, we will override that if we have reviseDate in metadata or lastEditDate
+  const reviseDate = getValueFromMetadata(newContent, "reviseDate");
+  const lastEditDate = getProp(newContent, "layer.editingInfo.lastEditDate");
+  if (reviseDate) {
+    newContent.updatedDate = new Date(reviseDate);
+    newContent.updatedDateSource = getMetadataPath("reviseDate");
+  } else if (lastEditDate) {
+    newContent.updatedDate = new Date(lastEditDate);
+    newContent.updatedDateSource = "layer.editingInfo.lastEditDate";
+  }
```
andrewctate, line 157: I'm assuming this try-and-fallback is done for reasons.

andrewctate (approved):

packages/content/src/content.ts
mjuniper: Yep, we want it to use the reviseDate if present and if not fall back to lastEditDate.

packages/content/src/content.ts
mjuniper: He means we had not specified a return type (but I have now on line 183 with `: Promise<IHubContent>`).

dbouwman (approved):
Can you add doc comment to the exported `UpdateFrequency` enum? Otherwise looks gtg to me

packages/content/src/content.ts
andrewctate: Ah. Yes, TypeScript won't complain about that unless it can't infer the return type through static code analysis. In this case, it can since you've clearly defined a return type for `_enrichDates` and the compiler sees that your promise will resolve to that return type.

andrewctate: Not saying it's a bad idea to add the explicit return type to this function. Just explaining the answer to Dave's comment
> not sure how the lint/ts compiler let that slide

PR #481, opened 2021-03-05
volta in packages should reference root package.json

mjuniper:
I think we should have the packages be on the same node version as the root of the project as described [here](https://docs.volta.sh/advanced/workspaces).


dbouwman (approved):
Awesome! I'll also make this change in templates-js

PR #485, opened 2021-03-11
[RFC] Dependency Upgrades

andrewctate:
This PR
- Bumps Karma dependencies which is long-overdue and resolves a problem in https://github.com/Esri/hub.js/pull/479
- Changes the default TS target to `ES2017` for testing purposes resolving this istanbul coverage bug: https://github.com/SitePen/remap-istanbul/issues/106
- Bumps the `rollup-typescript` plugin and locks the TS target at `ES5` so our UMD builds will still work.
- Forces an `ES5` ESM build for now so nothing changes in our distributed packages.

I checked for mentions of the `const` keyword in both the ESM and UMD builds and couldn't find anything so I think our target is successfully locked at ES5.


tomwayson (approved):

umd-base-profile.js
```diff
@@ -89,7 +89,7 @@ export default {
   context: "window",
   external: packageNames.concat(arcgisRestJsPackageNames),
   plugins: [
-    typescript(),
+    typescript({ target: 'es5' }),
```
tomwayson, line 92: I gather the reason you changed the rollup plugin is so you could pass in this target. You can't do that w/ `rollup-plugin-typescript2`?

Either way, it sounds like `@rollup/plugin-typescript` has come into it's own and may even be faster (would be nice to verify that four our project).

Also, we should explore not even passing the TS to rollup and just having it bundle the esm/es5 output since the UMD bundles are just an artifact needed at publish time (i.e. not used in tests nor needed to be continuously re-built in a watch). That has _got_ to be faster.

tomwayson (approved):
Your recent changes work for me. I would have guessed that you could have achieved the same w/ less changes by setting a karama-typescript config to override the es5 target to es2017, but the work you've done here gets us closer to publishing ES2017 output alongside the ESM/ES5 output, which I appreciate very much. 

If this solves the immediate problem, let's merge it as is. If you want to go a step further, you could try removing the rollup TS plugin altogether as I suggested and see if it speeds up the builds.

Either way I'd like to follow up soon w/ a PR to additionally publish ES2017.

umd-base-profile.js
andrewctate: It's a good idea to try to pass the esm build to rollup instead of relying on yet another typescript build.

andrewctate:
@tomwayson Sounds good. I'll merge.

PR #486, opened 2021-03-11
F/173170 disable downloads

drspacemanphd:
[177384](https://esriarlington.tpondemand.com/entity/177384-add-item-property-to-disable-downloads)
[88](https://devtopia.esri.com/dc/hub/issues/88)

Removes unused downloads locking logic and introduces logic for disabling private downloads.


rgwozdz (approved):
LGTM - is this technically a breaking change since you are removing the "locked" status?

packages/downloads/package.json
```diff
@@ -19,10 +19,12 @@
   "peerDependencies": {
     "@esri/arcgis-rest-feature-layer": "^2.21.0",
     "@esri/arcgis-rest-portal": "^2.21.0",
-    "@esri/arcgis-rest-request": "^2.21.0"
+    "@esri/arcgis-rest-request": "^2.21.0",
+    "@esri/hub-common": "^7.1.1"
   },
   "devDependencies": {
     "@esri/arcgis-rest-auth": "^2.21.0",
+    "@esri/hub-common": "^7.1.1",
```
trescube, line 27: Does this need to be in both dev and peer dependencies?

trescube (approved):

drspacemanphd:
> LGTM - is this technically a breaking change since you are removing the "locked" status?

Im waffling on keeping it in and simply changing the hub-components to ignore it temporarily. I don't necessarily think it's breaking, per se but I also think removing it probably isn't the way to go. I think I will amend this PR to keep it in, and we'll change hub-components accordingly when the sense of "locked" needs to be re-introduced.

packages/downloads/package.json
drspacemanphd: It's unclear to me, but that seems to be a pattern across several packages

rgwozdz (approved):

PR #487, opened 2021-03-13
streamline UMD builds

tomwayson:
- change UMD to build to use ESM output instead of TS source and remove @rollup/plugin-typescript
- only run UMD build when preparing to release (instead of after installs, in CI, etc) 

This cuts down the time it takes to run `bootstrap` from ~2:45 to ~1:05!!!


andrewctate:
@tomwayson  Nice! I like removing the UMD build from the bootstrap process. That will make working in Hub.js much more ergonomic.

However, since you've done that, do we really need to use the ESM build as the input or would it be beneficial to keep them on separate TS compilation routes? Wouldn't that allow us to just bump the ESM build target to `es2017` at the next major release without affecting the UMD (or requiring two versions of the ESM build)?

tomwayson:
Thanks @andrewctate, re:

> Wouldn't that allow us to just bump the ESM build target to es2017 at the next major release without affecting the UMD (or requiring two versions of the ESM build)

Yes, but, there won't be a time when we'll be doing an UMD build but not an ESM ES5 build. Think of the UMD build being useful to a subset of consumers that could use the ESM ES5 build. So might as well use the ESM ES5 build as the input for the UMD build. Hopefully the plan I laid out #488 helps clarify that.

PR #489, opened 2021-03-14
fix dev scripts

tomwayson:
dev:esm targets es5
remove dev:umd


tomwayson:
@dbouwman you are the only person in the world I can think of that might possibly have ever used the `dev:umd` scripts that this PR will remove. Let me know if this is a problem.

PR #494, opened 2021-03-24
Feat/update well known teams for membership access

benstoltz:
We are updating well known teams to account for 9.1 group changes (WRT permissions / who can be added). This PR makes these changes, and also conditionally loads old/new well known team types based on the current portal version.

### DIFF
--- a/packages/teams/src/older-well-known-teams.ts
+++ b/packages/teams/src/older-well-known-teams.ts
@@ -9,17 +9,14 @@ import { IGroupTemplate } from "./types";
  * This allows us to add/remove/edit groups by simply modifying this hash
  * instead of spreading complex construction logic all over the application
  */
-export const WELLKNOWNTEAMS: IGroupTemplate[] = [
+export const OLDERWELLKNOWNTEAMS: IGroupTemplate[] = [
   {
     config: {
       groupType: "Hub Collaboration Group",
       type: "core",
       availableIn: ["premium"],
       propertyName: "collaborationGroupId",
-      requiredPrivs: [
-        "portal:admin:createUpdateCapableGroup",
-        "portal:user:addExternalMembersToGroup"
-      ],
+      requiredPrivs: ["portal:admin:createUpdateCapableGroup"],
       titleI18n: "collaborationTitle",
       descriptionI18n: "collaborationDesc",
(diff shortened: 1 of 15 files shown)


packages/teams/src/older-well-known-teams.ts
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Hash of the team group templates.
+ * This hash is used to determin what teams can be created in what products
+ * based on what privs and stored in what properties
+ * We define the various groups via json structures. The .config hash controls
+ * the i18n, as well as the portal vs ago, and basic vs premium definition.
+ * This allows us to add/remove/edit groups by simply modifying this hash
+ * instead of spreading complex construction logic all over the application
+ */
+export const OLDERWELLKNOWNTEAMS: IGroupTemplate[] = [
```
dbouwman, line 12: can we be more specific with the naming - i.e. `PRE_9_1_WELLKNOWNTEAMS`

packages/teams/src/utils/get-user-creatable-teams.ts
```diff
@@ -12,12 +13,17 @@ import { IGroupTemplate } from "../types";
  */
 export function getUserCreatableTeams(
   user: IUser,
-  environment: HubProduct
+  environment: HubProduct,
+  currentVersion: string
 ): IGroupTemplate[] {
+  // TODO: remove this when needed
```
dbouwman, line 19: Put a date in here so next dev can know when they can remove it... i.e. after May 1 2021

dbouwman: Could we delegate to `getTeamsAvailableInProduct(...)` in this fn so we don't have the current version parse/compare logic twice?

dbouwman (changes requested):
Mainly naming things so it's really clear to users what `currentVersion` means, and what the `OLDERWELLKNOWNTEAMS` refers to. Please also ensure the jsdoc comments are updated for the changed signatures.

The duplication of the 
```js
const teams =
    parseFloat(currentVersion) < 9.1 ? OLDERWELLKNOWNTEAMS : WELLKNOWNTEAMS;
```
gives me pause... I think we can DRY that up

dbouwman (approved):

PR #495, opened 2021-03-25
fix: check user groups before confirming they can create a team

dbouwman:
This ensures that during the site deployment process we check that the user has some group capacity before


benstoltz (approved):
LGTM. Don't forget changelog. Any reason why 507 groups vs the number the api errors on?

dbouwman:
507 because that gives us room to create at least 5 groups... which is the max we're likely to ever create as part of a premium site deployment :)

PR #496, opened 2021-03-26
F/169 add telemetry and consent notice config

rweber-esri:
…new site.data.values.telemetry object. Additionally, prevent the schemaVersion from being incremented when only updating published status when saving a draft.

[169](https://devtopia.esri.com/dc/hub/issues/169)

1) Adds new `site.data.values.telemetry` config object
2) Move existing `gacode` under `site.data.values.telemetry.customAnalytics.ga`
3) Fixes bug where saving a site draft can cause loss of migration changes

### DIFF
--- a/packages/sites/src/drafts/save-published-status.ts
+++ b/packages/sites/src/drafts/save-published-status.ts
@@ -26,12 +26,11 @@ export function savePublishedStatus(
     // browser after saving the draft without first publishing,
     // migration changes can be lost
     const isUnpublished = hasUnpublishedChanges(siteOrPageModel);
-    prms = updateSite(
-      siteOrPageModel,
+    prms = updateSite(siteOrPageModel, {
+      ...hubRequestOptions,
       allowList,
-      hubRequestOptions,
-      !isUnpublished
-    );
+      updateVersions: !isUnpublished
+    });
   } else if (isPage(item)) {
     prms = updatePage(siteOrPageModel, allowList, hubRequestOptions);
   } else {
--- a/packages/sites/src/drafts/save-published-status.ts
+++ b/packages/sites/src/drafts/save-published-status.ts
@@ -22,9 +22,10 @@ export function savePublishedStatus(
 
   if (isSite(item)) {
     // when saving a draft site, we need to prevent the schemaVersion
-    // from being updated. if not, and the user refreshes the
-    // browser after saving the draft without first publishing,
-    // migration changes can be lost
+    // from being updated. otherwise, if the user does not publish the draft,
+    // functionality potentially will be broken for all users because the item
+    // reflects the most recent schemaVersion without any of the actual schema
+    // changes
     const isUnpublished = hasUnpublishedChanges(siteOrPageModel);
     prms = updateSite(siteOrPageModel, {
       ...hubRequestOptions,
--- a/packages/sites/src/drafts/save-published-status.ts
+++ b/packages/sites/src/drafts/save-published-status.ts
@@ -33,7 +33,10 @@ export function savePublishedStatus(
       updateVersions: !isUnpublished
     });
   } else if (isPage(item)) {
-    prms = updatePage(siteOrPageModel, allowList, hubRequestOptions);
+    prms = updatePage(siteOrPageModel, {
+      ...hubRequestOptions,
+      allowList
+    });
   } else {
     throw TypeError(
       "@esri/hub-sites: only page or site models have a published state"
--- a/packages/sites/src/drafts/save-published-status.ts
+++ b/packages/sites/src/drafts/save-published-status.ts
@@ -3,6 +3,7 @@ import { updateSite } from "../update-site";
 import { updatePage, isPage } from "../pages";
 import { isSite } from "../is-site";
 import { IUpdateItemResponse } from "@esri/arcgis-rest-portal";
+import { hasUnpublishedChanges } from "./has-unpublished-changes";
 
 /**
  * Saves the published status of a site or page model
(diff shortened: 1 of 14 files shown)


packages/sites/src/index.ts
```diff
@@ -58,3 +58,4 @@ export * from "./get-data-for-site-item";
 export * from "./get-site-by-id";
 export * from "./is-site";
 export * from "./get-members";
+export * from "./_ensure-telemetry";
```
tomwayson, line 61: should this be exported here? It appears this is a "private" fn that we only export for ease of testing.

packages/sites/src/drafts/save-published-status.ts
```diff
@@ -20,7 +21,17 @@ export function savePublishedStatus(
   let prms;
 
   if (isSite(item)) {
-    prms = updateSite(siteOrPageModel, allowList, hubRequestOptions);
+    // when saving a draft site, we need to prevent the schemaVersion
+    // from being updated. if not, and the user refreshes the
+    // browser after saving the draft without first publishing,
+    // migration changes can be lost
```
tomwayson, line 27: I don't quite follow how the changes would be lost, but I trust you.

packages/sites/src/update-site.ts
```diff
@@ -21,11 +21,13 @@ import { updateItem, IUpdateItemResponse } from "@esri/arcgis-rest-portal";
...
  * @param {Array} patchList Array of property paths to update
  * @param {IHubRequestOptions} hubRequestOptions
+ * @param {boolean} [updateVersions=true] Optionally update the versions, defaults to true
  */
 export function updateSite(
   model: IModel,
   patchList: string[],
-  hubRequestOptions: IHubRequestOptions
+  hubRequestOptions: IHubRequestOptions,
+  updateVersions: boolean = true
```
tomwayson, line 30: is there another way we can achieve the same w/o adding a trailing boolean argument? We try to have request options as the final argument. Maybe have `IUpdateSiteOptions extends IHubRequestOptions` that has `updateVersions` in it (also should ultimately have `patchList` in it too)?

tomwayson (commented):
Nice work @rweber-esri. I just wonder if there's another way we can deal w/ `updateVersions`.

packages/sites/src/update-site.ts
dbouwman: +1 for `IUpdateSiteOptions` vs trailing boolean

packages/sites/src/drafts/save-published-status.ts
dbouwman: I'm not following this either... @rweber-esri can you explain what at risk of getting lost?

rweber-esri: This resolves an issue I noticed that can be reproduced by the following steps:
1. Site is at schemaVersion 1.3 in AGO
2. User navigates to /edit, for which schema 1.4 is applied, but not persisted to AGO.
3. User makes local changes & saves draft.
4. When the above method is invoked, we make `typeKeywords` the only thing in the `allowList`, but the `updateSite` method automatically adds `schemaVersion` to the `allowList`, which results in `schemaVersion` being bumped to 1.4 & the typeKeyword representing we have an unpublished draft is added to the item, but none of the migration changes are committed as a result.
5. User refreshes their browser, now the item in AGO has schemaVersion 1.4, so the migration isn't applied.
6. The user saves another draft
7. Migrations changes are lost to the wind

packages/sites/src/update-site.ts
rweber-esri: I was going for a non-breaking change to avoid a major bump, but agree, the proposed is better. Looks reasonably easy to account for in Hub & Solutions.js.

packages/sites/src/index.ts
rweber-esri: I only added it b/c I noticed other private method & migration module exports, doesn't need exported, so happy to remove

rweber-esri: Scratch that, deleting the export statement breaks tests that try to mock out this migration

packages/sites/src/drafts/save-published-status.ts
rweber-esri: Updated the comment to be more accurate wrt the STR in my previous comment

packages/sites/src/update-site.ts
```diff
@@ -15,73 +16,73 @@ import { updateItem, IUpdateItemResponse } from "@esri/arcgis-rest-portal";
...
- * The patchList can include any property paths on the item.
+ * a subset of property changes to the model if a allowList is included.
+ * The allowList can include any property paths on the item.
  * If the list is empty, then the entire site model is overwritten.
  * @param {Object} model Site Model to update
- * @param {Array} patchList Array of property paths to update
- * @param {IHubRequestOptions} hubRequestOptions
+ * @param {IUpdateSiteOptions} updateSiteOptions
  */
 export function updateSite(
```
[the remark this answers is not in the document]
rweber-esri: I don't disagree, but it looks like [update-page](https://github.com/Esri/hub.js/blob/master/packages/sites/src/pages/update-page.ts#L32) already differs from `update-page` in that it can receive 2 or 3 args and there's some variable reassignment to account for this.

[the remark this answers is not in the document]
dbouwman: In an ideal world, we'd keep these signatures pretty similar... but they are different item types, so I don't think we should get too wound up about it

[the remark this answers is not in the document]
rweber-esri: Thanks for your input @dbouwman. Will hold off on making this change for now then.

rweber-esri:
I believe I've addressed all PR feedback. Hoping to get this merged & published today, scheduled to go out w/ tomorrow's release & need to huck a Solutions.js PR to bump their dependency on this package beforehand. 🤞 Any additional feedback? Not sure if @tomwayson would like to take one last peak?

packages/sites/src/update-site.ts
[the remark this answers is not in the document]
tomwayson: FYI - though, this was already on the [list of things to change at the next breaking version](https://github.com/Esri/hub.js/issues/305): 

https://github.com/Esri/hub.js/blob/master/packages/sites/src/pages/update-page.ts#L32-L35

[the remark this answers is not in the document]
rweber-esri: Pushed another commit to add `IUpdatePageOptions` and refactor `updatePage` to use per side-convo w/ @tomwayson

tomwayson (approved):
Thanks for making those changes @rweber-esri

PR #497, opened 2021-03-30
C/breaking changes

tomwayson:
addresses a few of the issues in #305, but punts on some of the thornier ones (`getItemThumbnailUrl()` support for width, removing the deprecated `getCategory()`, etc). Unfortunately, those will have to come in a later breaking change.


rweber-esri (approved):
LGTM

PR #499, opened 2021-03-31
feat(content): getContent now fetches the item from the AGO API 

mjuniper:
We now fetch the item from the AGO API even in cases where we were able to fetch it from the Hub API.

https://devtopia.esri.com/dc/hub/issues/53

affects: @esri/hub-content

### DIFF
--- a/packages/content/src/hub.ts
+++ b/packages/content/src/hub.ts
@@ -1,9 +1,10 @@
 import { ResourceObject } from "jsonapi-typescript";
-import { IItem } from "@esri/arcgis-rest-portal";
+import { IItem, getItem } from "@esri/arcgis-rest-portal";
 import {
   IHubContent,
   IHubRequestOptions,
-  hubApiRequest
+  hubApiRequest,
+  mergeObjects
 } from "@esri/hub-common";
 import { itemToContent, withPortalUrls } from "./portal";
 import { isSlug, addContextToSlug, parseDatasetId } from "./slugs";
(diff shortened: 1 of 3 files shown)


packages/content/src/hub.ts
```diff
@@ -51,9 +52,29 @@ export function getContentFromHub(
...
-    return dataset && withPortalUrls(datasetToContent(dataset), options);
-  });
+  return request
+    .then((dataset: any) => {
+      // only if authed
+      if (dataset && options.authentication) {
+        // we fetch the item - this is because if an item is contentStatus: org_authoritative
+        // we do not get that info unless we are authed in the org
+        // see https://devtopia.esri.com/dc/hub/issues/53#issuecomment-2769965
+        return getItem(dataset.id, options).then(item => {
```
tomwayson, line 62:
```suggestion
        return getItem(parseDatasetId(dataset.id).itemId, options).then(item => {
```

packages/content/src/hub.ts
```diff
@@ -51,9 +52,29 @@ export function getContentFromHub(
...
+          const itemOverrides = ["contentStatus"];
```
tomwayson, line 63: I think we should add more overrides here.

At the very least I believe that `itemControl`, like `contentStatus` is based on the authenticated user.

Also, there are bunch of properties that are not returned by the Hub API that we should consider adding here:

https://github.com/Esri/hub.js/blob/da2d344130c4b91abbd0535b1ec518362242f216/packages/content/src/hub.ts#L161-L193

Honestly, I think it'd be easier to use a block list based on known oddities (`name`, `extent`, etc) than an allow list. I don't know that full list off the top of my head, but I could help you figure it out if you want to go that route.

tomwayson (changes requested):
Thanks for doing this @mjuniper - a couple of minor changes.

packages/content/src/hub.ts
mjuniper: I think I will stick with a whitelist for now since, that is what mergeObjects wants.

tomwayson (approved):

Issue #498, opened 2021-03-31
Site Draft Schema Upgrade Strategy


andrewctate:
### Site Schema Upgrades
Schema upgrades to Hub sites are a series of transformations applied to the site item. They occur in the [`upgradeSiteSchema`](https://github.com/Esri/hub.js/blob/master/packages/sites/src/upgrade-site-schema.ts) utility and are automatically applied every time a site is fetched from the API via `getSite`. We keep track of the current schema version of a Hub site in the `item.properties.schemaVersion` property.

### Site Drafts
Site drafts are subsets of site models (`Partial<IModel>`). They are designed such that they can be applied dynamically to any site model via a naive [merge operation](https://github.com/Esri/hub.js/blob/master/packages/sites/src/drafts/apply-draft.ts#L13) that _always gives precedent to the site draft's properties_. They contain only the properties defined in [this list](https://github.com/Esri/hub.js/blob/master/packages/sites/src/drafts/site-draft-include-list.ts).

## The Problem
Imagine we have a site ("the site") at with schema v1, a migration to move the site to schema v2 ("the migration"), a draft ("the draft") and a client application such as Hub itself ("the client").

Consider this scenario
1. The client loads the site (the migration is automatically applied putting it at v2)
2. It loads the draft (a subset of the v1 site)
3. It applies the draft
4. It saves the site's item and data.

Result: parts of v2 site model will have overwritten by the draft taken from the v1 site. To make matters worse, the site's `schemaVersion` will still have been successfully updated to `2` making it so that `upgradeSiteSchema` will skip applying the migration in the future (this is because the `item.properties.schemaVersion` prop is not currently included in drafts so it won't overwrite the value on the site model).

### 1. Apply Migrations to the Draft
In this solution, it is the responsibility of Hub.js to make three important guarantees:
1. the site's schema gets upgraded when it is fetched (already happens today)
2. the merge operation will properly apply the draft
3. the site will not lose its updated schema when the draft is applied

We would most likely facilitate this by including the `schemaVersion` prop on the draft and running it through the same schema migrations that the site runs through.

![Blank diagram (5)](https://user-images.githubusercontent.com/315764/113072596-14c80080-918d-11eb-96d0-09d3f8d9b01b.png)

#### Benefits
This approach has several benefits.
- The complexity is encapsulated in Hub.js. No extra burden is shouldered by the client.
- These three guarantees lend themselves well to unit testing.
[...]

PR #500, opened 2021-03-31
draft schema migrations

andrewctate:
Implements solution 1 from https://github.com/Esri/hub.js/issues/498

### DIFF
--- a/packages/common/src/objects/set-prop.ts
+++ b/packages/common/src/objects/set-prop.ts
@@ -1,0 +1,22 @@
+/**
+ * Sets a deep object property, constructing the property path as necessary
+ *
+ * @param path - the path to the property we want to set
+ * @param val - the value we want to set it to
+ * @param obj - the target object
+ */
+export function setProp(path: string | string[], val: any, obj: any): boolean {
+  const props = Array.isArray(path) ? path : path.split(".");
+  if (props.length > 1) {
+    // does the object have this prop? if not, create it as an object
+    if (!obj.hasOwnProperty(props[0])) {
+      obj[props[0]] = {};
+    }
+    return setProp(props.slice(1), val, obj[props[0]]);
+    // recurse back into this, but with one less entry in the props
+  } else {
+    // this is the last one, the one we want to replace
+    obj[props[0]] = val;
+    return true;
+  }
+}
--- a/packages/common/src/objects/set-prop.ts
+++ b/packages/common/src/objects/set-prop.ts
@@ -1,3 +1,5 @@
+import { deepSet } from "./deep-set";
+
 /**
  * Sets a deep object property, constructing the property path as necessary
  *
@@ -5,18 +7,9 @@
  * @param val - the value we want to set it to
  * @param obj - the target object
  */
-export function setProp(path: string | string[], val: any, obj: any): boolean {
-  const props = Array.isArray(path) ? path : path.split(".");
-  if (props.length > 1) {
-    // does the object have this prop? if not, create it as an object
-    if (!obj.hasOwnProperty(props[0])) {
-      obj[props[0]] = {};
-    }
-    return setProp(props.slice(1), val, obj[props[0]]);
-    // recurse back into this, but with one less entry in the props
-  } else {
-    // this is the last one, the one we want to replace
-    obj[props[0]] = val;
-    return true;
+export function setProp(path: string | string[], val: any, obj: any) {
+  if (Array.isArray(path)) {
+    path = path.join(".");
   }
+  deepSet(obj, path, val);
 }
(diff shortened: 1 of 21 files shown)


packages/sites/src/_ensure-telemetry.ts
```diff
@@ -1,12 +1,19 @@
...
+} from "@esri/hub-common";
 
 /**
  * Add telemetry config object
  * @private
  * @param {object} model Site Model
  * @returns {object}
  */
-export function _ensureTelemetry(model: IModel) {
+export function _ensureTelemetry<T extends IModel | IDraft>(model: T): T {
```
andrewctate, line 16: This is the pattern I'm proposing we follow for future migration functions.

dbouwman (approved):

packages/common/src/objects/set-prop.ts
```diff
@@ -0,0 +1,15 @@
+import { deepSet } from "./deep-set";
+
+/**
+ * Sets a deep object property, constructing the property path as necessary
+ *
+ * @param path - the path to the property we want to set
+ * @param val - the value we want to set it to
+ * @param obj - the target object
+ */
+export function setProp(path: string | string[], val: any, obj: any) {
```
tomwayson, line 10: what's the diff w/ this and [deepSet()](https://esri.github.io/hub.js/api/common/deepSet/)?

tomwayson: nvm, I see you've updated it to use `deepSet()` now.

andrewctate:
Looks like a flakey downloads test is failing due to unsuccessful mocking in Karma+Chrome. Tests pass in Node.

tomwayson:
That flakey test has caused my last 2 publishes to fail. @rgwozdz I know you were looking at it, any update?

packages/sites/src/drafts/upgrade-draft-schema.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  if (getProp(draft, "item.properties.schemaVersion") === SITE_SCHEMA_VERSION) {
+    return draft;
+  } else {
+    let migrated = draft;
+    // apply site schema upgrade functions in order...
+    // don't have do do them all since drafts only got released
+    // at version 1.3
+    migrated = _ensureTelemetry<IDraft>(draft);
```
rweber-esri, line 24: Minor, `<IDraft>` & `<IModel>` aren't necessary before invoking `_ensureTelemetry`.

packages/sites/src/_ensure-telemetry.ts
```diff
@@ -26,6 +33,7 @@ export function _ensureTelemetry(model: IModel) {
     }
   };
   deleteProp(clone, "data.values.gacode");
```
rweber-esri, line 35: Do we want to delete this line?

rweber-esri (commented):
Looks good to me overall. Should we consider changing `getModel` to `getSiteById` on [this line](https://github.com/Esri/hub.js/blob/master/packages/sites/src/update-site.ts#L68) to make sure we're not merging changes into an unmigrated model?

rweber-esri:
Hooked up to my sandbox, things seem to work as I'd expect.

packages/sites/src/_ensure-telemetry.ts
andrewctate: I think it's okay to leave it. It will remove that prop from both the draft and the site, so no problem.

packages/sites/src/drafts/upgrade-draft-schema.ts
andrewctate: Hi @rweber-esri . I changed `_ensureTelemetry` to accept a Typescript generic [here](https://github.com/Esri/hub.js/pull/500/files#diff-eff07968780063bba314ca024cbb534e5ebd32408579a368817af36a484bfb42R16). Could you clarify what you mean?

andrewctate:
> Looks good to me overall. Should we consider changing getModel to getSiteById on this line to make sure we're not merging changes into an unmigrated model?

@rweber-esri I think this is a very good idea. Do you agree @dbouwman ?

packages/sites/src/drafts/upgrade-draft-schema.ts
rweber-esri: FWIW, this isn't _super_ important, just pointing out we could save a few bytes is all. The `T` in that generic signature represents whatever type is passed to the function (in this case either IModel or IDraft). Since the signature defines that the method will receive an object of type `T` & return an object of type `T`, it means the type can be inferred at runtime and we don't need to explicitly tell `_ensureTelemetry` whether we're passing an `IDraft` or `IModel`. E.g.

```
// migrated = _ensureTelemetry<IDraft>(draft); // can be reduced to just
migrated = _ensureTelemetry(draft);

// and

// model = _ensureTelemetry<IModel>(model); // can be reduced to just
model = _ensureTelemetry(model);
```

That being said, there's nothing _wrong_ w/ what you did, could just reduce it a bit is all.

rweber-esri (approved):

PR #502, opened 2021-04-08
F/388 add team view functions

vivzhang:
Add these functions in -

- getTeamById
- getTeamMembers
- searchTeamContent
- searchTeamMembers

Export `shareItemsToSiteGroups` from `_share-items-to-site-groups.ts`, so it can be used in opendata-ui


dbouwman (approved):
All looks good to me

PR #513, opened 2021-04-16
C/ember demo

tomwayson:
Adds an Ember demo and a `dev:ember` script that you can use to have changes you make to the Hub.js dependencies cause the Ember app to live reload.

This also does change the ESM build scripts so that they:
- build es2017 at dev time, and es5 only at publish time
- ~~put the ES2017 and ES5 code in the `dist/esm/` and `dist/es5` folders respectively~~

The latter _may_ fix https://github.com/Esri/hub-components/issues/62. It ~~will~~ would definitely break the Hub Ember app after we release and update Hub.js packages, so ~~I better do the next release~~ I reverted it and will wait for a breaking release.

There's a lot in this PR, mostly the Ember app itself and corresponding changes to package-lock.json. Probably the best way to review is to just look ~~at~~ from the second commit.


demos/ember-app/.ember-cli
```diff
@@ -0,0 +1,9 @@
+{
+  /**
+    Ember CLI sends analytics information by default. The data is completely
+    anonymous, but there are times when you might want to disable this behavior.
+
+    Setting `disableAnalytics` to true will prevent any data from being sent.
+  */
+  "disableAnalytics": false
+}
```
mjuniper, line 9: Wonder if we should disable analytics so as not to pollute their data with info about a not-real app?

packages/content/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Module to interact with ArcGIS Hub Content in Node.js and modern browsers.",
   "main": "dist/node/index.js",
   "unpkg": "dist/umd/content.umd.js",
-  "module": "dist/esm/index.js",
-  "es2017": "dist/es2017/index.js",
+  "module": "dist/es5/index.js",
+  "es2017": "dist/esm/index.js",
```
mjuniper, line 8: Ha! I think this might explain something that was unexpected (to me at least) in the built files.

mjuniper (approved):
LGTM

packages/content/package.json
tomwayson: yeah, we're still defaulting ("module") to ESM b/c I didn't want to do a breaking change yet.

PR #516, opened 2021-04-21
Move add users to group logic into hub-teams

benstoltz:


benstoltz:
Closing due to realization that hub-events is also using the add users flow.

PR #517, opened 2021-04-21
WIP: update getContentMetadata to dynamically import fast-xml-parser

tomwayson:
This causes the fast-xml-parser library to be lazy loaded.

Tests pass for me locally and I verified that the ES2017 build works in the webpack and demo builds and that the CJS build works in the node demo. In a production build of the Ember demo this results in lazy loading a ~28kB chunk.

image

This breaks in CI b/c it tries to run the ESM ES5 and UMD builds, and I haven't gotten those working yet. I'm wondering how much to invest in that (i.e. how long do we need to support IE and non-ESM loaders (i.e. AMD?).

TODO:
- get ESM ES5 and UMD builds working

### DIFF
--- a/packages/content/src/content.ts
+++ b/packages/content/src/content.ts
@@ -6,7 +6,6 @@ import {
   getUser,
   IGetUserOptions
 } from "@esri/arcgis-rest-portal";
-import { request } from "@esri/arcgis-rest-request";
 import {
   HubType,
   IHubContent,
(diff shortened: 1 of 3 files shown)


mjuniper (approved):
Badass.

tomwayson:
Also, I have yet to verify that this works when consumed by a stencil library, and given https://github.com/Esri/hub-components/issues/62 that might be a problem.

ajturner:
@tomwayson this is draft - but is it still relevant to consider and test in Stencil? Or is it superseded and should be closed?

ajturner:
This looks like a terrific enhancement - but want to make a decision whether we should have another dev review + verify or keep in the backlog

tomwayson:
closing this to be reconsidered after the v10 release

PR #519, opened 2021-04-27
fix(content): update datasetToContent method to add isProxied propert…

rweber-esri:
…y to result

affects: @esri/hub-content

ISSUES CLOSED: https://devtopia.esri.com/dc/hub/issues/71

Adds `isProxied` to the result of `datasetToContent` method so `opendata-ui` can more accurately represent proxied CSVs as `CSV` vs `Table` in type metadata.


tomwayson (approved):
thanks @rweber-esri

PR #523, opened 2021-05-06
fix(common): refactors hasBasePriv to support falsy users & simplifie…

rweber-esri:
[269](https://devtopia.esri.com/dc/hub/issues/269)

1. Refactors `hasBasePriv` to handle `undefined` users vs throwing an error.
2. Simplifies logic in `canEditItem` to leverage `itemControl` property.

### DIFF
--- a/packages/common/src/access/can-edit-item.ts
+++ b/packages/common/src/access/can-edit-item.ts
@@ -1,35 +1,13 @@
-import { IItem, IUser, IGroup } from "@esri/arcgis-rest-types";
-import { isUpdateGroup, includes } from "../utils";
-import { getProp } from "../objects";
+import { IItem, IUser } from "@esri/arcgis-rest-types";
 import { hasBasePriv } from "./has-base-priv";
+import { includes } from "../utils";
 
 /**
  * Checks if user has access to edit an item in Hub
- * @param {IItem} item
- * @param {IUser} user
+ * @param {IItem} item The item to be edited
+ * @param {IUser} [user] An optional user
  * @returns {boolean}
  */
-export function canEditItem(item: IItem, user: IUser): boolean {
-  let res = false;
-  const itemControls = ["admin", "update"];
-  const { itemControl, owner, orgId: itemOrgId } = item;
-  const { roleId, role, username, groups: userGroups, orgId: userOrgId } = user;
-  const hasItemControl = includes(itemControls, itemControl);
-  const isOwner = !!owner && owner === username;
-  const isOrgItem = !!itemOrgId && itemOrgId === userOrgId;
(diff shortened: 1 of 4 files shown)


packages/common/src/access/can-edit-item.ts
```diff
@@ -1,35 +1,13 @@
...
-      ...(getProp(item, "groupIds") || []),
-      getProp(item, "properties.collaborationGroupId")
-    ];
-    const isGroupEditable = (group: IGroup): boolean =>
-      isUpdateGroup(group) && includes(itemGroups, group.id);
-    res = userGroups.some(isGroupEditable);
-  }
-  return res;
+export function canEditItem(item: IItem, user?: IUser): boolean {
+  return hasBasePriv(user) && includes(["admin", "update"], item.itemControl);
```
cpgruber, line 12: This is obviously much simpler and easier to reason about. However, iirc the reason we had to jump through all the additional gnarly checks was that `item.itemControl` wasn't always available, depending on whether the item comes from indexer of directly from the platform. 

Also, the decision to make `user` a required argument was that it would fail fast. I don't think we should be calling these methods if there's no user.

rweber-esri: > This is obviously much simpler and easier to reason about. However, iirc the reason we had to jump through all the additional gnarly checks was that item.itemControl wasn't always available, depending on whether the item comes from indexer of directly from the platform.

Sigh. I just found a place where we're passing teams to this `canEditItem` method, which the above changes sadly won't work for either. Will mostly revert & add a comment so nobody else gets burned by this.

> Also, the rationale for making user a required argument was that it would fail fast. I don't think we should be calling these methods if there's no user.

@mjuniper care to weigh in? You requested this change [here](https://devtopia.esri.com/dc/hub/issues/269)

mjuniper: Based on your findings it seems like my request is not practicable at this time.

rweber-esri: > Based on your findings it seems like my request is not practicable at this time.

Ok, thanks for the feedback guys. Closing this PR then.

cpgruber: Valiant effort, Randy. Re: passing a team to `canEditItem`, that just seems like an incorrect use of the method. I would expect there to be a similar method, like `canEditTeam`, to handle that scenario.

A lot of the history is coming back to me now, I think it mostly hinged on the fact that an edit pencil would never appear for an unauthenticated user, so we can reasonably assume that `user` to be defined.

tomwayson: FTR - `item.itemControl` is only returned if the user is authenticated

PR #524, opened 2021-05-07
f/522 metadata dates

mjuniper:
https://devtopia.esri.com/dc/hub/issues/918
https://devtopia.esri.com/dc/hub/issues/920

Adds metadataUpdatedDate and metadataUpdatedDateSource to getContent.

~Returns the value we get from the metadata (if it is parsable as a Date) without trying to parse it to a Date. This is because the home app allows people to put in any part of an ISO string, like `2018`, `2018-02`, `2018-02-07`, etc, etc. There is actually no validation so even that is just on the honor system. If a user enters, for example, `2018`, we do not want to just parse that to a date and return it because then we lose the fact that the user said `2018`, not `01/01/2018`.~

Returns additional fields metadataUpdatedDatePrecision, updatedDatePrecision, & publishedDatePrecision that indicate the level of precision provided by the user.


tomwayson (changes requested):
I get why you don't want to parse the partial dates, but I don't like `updatedDate: Date | string` or `publishedDate: Date | string` at all. I see the point of those properties is to provide a `Date`. I know our app likely just uses those for display (i.e. not comparison w/ other dates), but is your plan to just render that string as is in the app w/o parsing into a date and then formatting?

We won't _lose_ the source date (it will still be buried in metadata) but I do think that providing a parsed date can be misleading (`new Date(2018, 0, 1)` is not the same as '2018').

mjuniper:
My plan is to use that date info to try to render it at the intended level of precision.

https://github.com/ArcGIS/opendata-ui/pull/9531

tomwayson (approved):
thanks for adding `parseISODateString()` and `DatePrecision`

PR #530, opened 2021-05-12
[WIP] hub-core package

dbouwman:
Adds:
- `@esri/hub-core` package, which has:
  - `HubSession`
  - `buildSession(...)`
  - `getDefaultRequestOptions(...)` (for unauthd calls) 

Ended up calling it `hub-core`, in `/packages/core` as there is tooling that _really_ wants that pattern.

I also had to change the `karma.e2e.conf` to ref `tsconfig.e2e.conf` which does an es5 build... was getting a regenerator error w/o that. Maybe @tomwayson has some ideas around that.

Zero unit tests, but did write e2e's to flex the basic pattern and it works.

Things to specifically suss...
- builder pattern. This is almost 100% the final version of what @rweber-esri had in my repl
- patterns for getters that abstract "cognative load" 
   - As a start around this I added `session.canCreateUpdateGroup`... which is ok, but there could be many many permutations... alternative would be `session.canCreateGroup(group)` and that introspects the structure and checks privs depending on what it finds. Open to any ideas


packages/core/src/HubSession.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  /**
+   * Hub API Url
+   * Will return `null` in Enterprise
+   * @readonly
+   * @memberof HubSession
+   */
+  get hubApiUrl() {
+    // if portal this should be null
+    if (this.isPortal) {
```
rweber-esri, line 83: Do we care about multiple returns in Hub.js? If so, could be reduced to return the results of a ternary.

packages/core/src/HubSession.ts
```diff
@@ -0,0 +1,116 @@
...
+    return this._userSelf.groups.length;
+  }
+
+  get canCreateUpdateGroup() {
+    const requiredPrivs = ["portal:admin:createUpdateCapableGroup"];
+    return this.canCreateGroup && this.hasAllPrivs(requiredPrivs);
+  }
+
+  get canCreateGroup() {
+    if (this.userGroupCount < 511) {
```
rweber-esri, line 100: Similar, could be reduced to single return of a ternary vs multiple returns

```js
  return this.userGroupCount < 511
    ? this.hasAllPrivs(["portal:user:createGroup"])
    : false;
```

packages/core/src/HubSession.ts
```diff
@@ -0,0 +1,116 @@
...
+      const requiredPrivs = ["portal:user:createGroup"];
+      return this.hasAllPrivs(requiredPrivs);
+    } else {
+      return false;
+    }
+  }
+
+  hasAllPrivs(privs: string[]): boolean {
+    return privs.reduce((acc: boolean, val) => {
```
rweber-esri, line 109: Array.prototype.every would be more performant.

`return privs.every(priv => includes(this._userSelf.privileges, priv);`

packages/core/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { HubSession } from "./HubSession";
+export * from "./utils";
```
rweber-esri, line 2: Do we want to be more specific w/ our exports here?

packages/core/src/factories.ts
```diff
@@ -0,0 +1,9 @@
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { HubSession } from "./HubSession";
+
+export async function buildHubSession(
```
rweber-esri, line 4: Can drop `async` since no `await`.

Do we want to add comments/JSDoc here and all method class members for consistency?

packages/core/src/utils/getDefaultRequestOptions.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Return a minimal IHubRequestOptions object for
+ * use in functions that don't require authentication
+ * but do need some url information
+ *
+ * @export
+ */
+
+export function getDefaultRequestOptions(
+  portalUrl: string = "https://maps.arcgis.com/sharing/rest"
```
rweber-esri, line 12: Should `portalUrl` be `portalUrl?` since we're assigning a default value? `portalUrl?: string = "https://maps.arcgis.com/sharing/rest"`

Also, update method doc w/ args?

packages/core/src/HubSession.ts
```diff
@@ -0,0 +1,116 @@
...
+    // if portal this should be null
+    if (this.isPortal) {
+      return null;
+    } else {
+      return this._hubApiUrl;
+    }
+  }
+
+  get userGroupCount() {
+    return this._userSelf.groups.length;
```
rweber-esri, line 91: How do we plan to handle anonymous users? Curious if this should be `this._userself?.groups?.length ?? 0;`

packages/core/src/HubSession.ts
```diff
@@ -0,0 +1,116 @@
...
+  includes
+} from "@esri/hub-common";
+import { IUser, UserSession } from "@esri/arcgis-rest-auth";
+import { getSelf, IPortal } from "@esri/arcgis-rest-portal";
+
+export class HubSession implements IHubRequestOptions {
+  // Note: can't use a getter for this or we get failures
+  // from inside rest-js where object.assign is used frequently
+  // and getters are not copied over
+  authentication: UserSession;
```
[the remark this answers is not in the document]
dbouwman: Maybe - we can try in the e2e's, but from what I saw, the `authentication` property was simply not present on the object that came out of `Object.assign` when authentication used a getter.

packages/core/src/HubSession.ts
dbouwman: should definitely have `?`'s

packages/core/src/utils/getDefaultRequestOptions.ts
dbouwman: TS complains (for me anyhow) if you use `?` and assign a default value...

packages/core/src/HubSession.ts
tomwayson: So `?.` and `??` are some of the few language features that I'm not sure about using yet. They're both relatively recent, and neither is in ES2017 and I haven't looked into the consequences of using them in TS when that is your target vs just using our own functions like `getProp` (assuming we intend to keep shipping that).

packages/core/src/index.ts
dbouwman: Sure - I had some vague recollection from @tomwayson about moving the super simple array/object util fns here for some tree shaking reasons... if we're not doing that, we can drop the second line

packages/core/src/utils/getDefaultRequestOptions.ts
dbouwman: image

packages/core/src/HubSession.ts
dbouwman: fwiw - I created the `getDefaultRequestOptions(...)` fn for un-authenticated workflows... it just seems overly complex to have HubSession support not having a session.

packages/core/src/HubSession.ts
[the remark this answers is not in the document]
dbouwman: Addressed everything else in a new commit - and we can try this, but I suspect it will not work.

dbouwman:
Closing as this was more of an experiment / ideation exercise vs shippable code. We've since moved to ArcGISContext{Manager}, and have keeping code in the `hub-common` package

PR #531, opened 2021-05-13
feat(sites): preserve a created initiative's id

MikeTschudi:
affects: @esri/hub-sites

By adding the id of a created initiative to its record within a hub site, it is easier to trace from the site to the initiative during cleanup


dbouwman:
I believe that property you want is already present in `item.properties.parentInitiativeId`

image

MikeTschudi:
Will use `removeSite` instead.

Issue #538, opened 2021-06-09
getContent() allows callers to only fetch specific enrichments


tomwayson:
Carried over from https://github.com/Esri/hub.js/issues/322#issuecomment-708617199

We were in agreement that `getContent()` should make "smart" default choices about whether or not to fetch additional enrichments, based on things like whether or not the data was fetched from the API (and already includes the enrichments), and the family or type of the content (i.e. we need much more data for feature layers than we do for documents). 

This issue is about providing a way to override those "smart" choices for example to avoid additional xhrs when we know they are not needed.

I propose adding `IGetContentOptions.enrichments: string[]` which is an array of the specific enrichments (`IHubContent` properties that are not part of `IItem`) that the user wants to include. If the `enrichments` options is not specified (i.e. falsey), `getContent()` would continue the current behavior which is to fetch all enrichments that based on API, family and type, etc. To skip fetching _any_ enrichments, you'd pass an empty array like: `getContent(identifier, {...requestOptions, enrichments: []})`. Note that even in that case if the content was returned from the Hub API, it would still include any enrichments that the API returned.

PR #540, opened 2021-06-10
allow consumers to control what enrichments are fetched

tomwayson:
The reason for this PR is to allow consumers to control what enrichments are fetched via a `requestOptions.enrichments` property as proposed in #538.

It also refactors how we enrich content by exposing new  function`enrichContent(content, requestOptions)` which calls a new `fetchContent(content, requestOptions)`.  `enrichContent()` is called by `getContentFromPortal()` and `getContentFromHub()` (instead of at the end of `getContent()`). This means that you can expect those functions to include the same enrichments as `getContent()`. This also means that a lot of 🧀 has moved out of the content module and corresponding test and into a new enrichments module and test. 

All in all I'm much happier w/ the refactored functions and new APIs and they should make it easier to implement and test #322.


packages/content/src/enrichments.ts
```diff
@@ -0,0 +1,443 @@
...
+enum DatePrecision {
+  Year = "year",
+  Month = "month",
+  Day = "day",
+  Time = "time"
+}
+
+function getMetadataPath(identifier: keyof IMetadataPaths) {
+  // NOTE: i have verified that this will work regardless of the "Metadata Style" set on the org
+  const metadataPaths: IMetadataPaths = {
```
rweber-esri, line 60: Could probably just define this as a `const` outside the scope of this `fn` in place of `IMetadataPaths`?

packages/content/src/enrichments.ts
```diff
@@ -0,0 +1,443 @@
...
+    date = new Date(+parts[0], +parts[1] - 1, +parts[2]);
+    precision = DatePrecision.Day;
+  } else if (!Number.isNaN(Date.parse(isoString))) {
+    // any other string parsable to a valid date
+    date = new Date(isoString);
+    precision = isoString.includes("T")
+      ? DatePrecision.Time
+      : DatePrecision.Day;
+  }
+  return date && precision && { date, precision };
```
rweber-esri, line 116: Maybe deserving of an interface?

packages/content/src/enrichments.ts
```diff
@@ -0,0 +1,443 @@
...
+ * and b) you cannot naively call new Date with an ISO 8601 string that does not include time information
+ * For example, when I, here in mountain time, do new Date('2018').getFullYear() I get "2017".
+ * This is because when you do not provide time or timezone info, UTC is assumed, so new Date('2018') is 2018-01-01T00:00:00 in UTC
+ * which is actually 7 hours earlier here in mountain time.
+ *
+ * @param {string} isoString
+ * @return { date: Date, precision: DatePrecision }
+ */
+export function parseISODateString(isoString: string) {
+  isoString = `${isoString}`;
```
rweber-esri, line 92: For some reason I _thought_ our lint rules complained about argument reassignment.This is probably fine, but might be better practice to not mutate/reassign args.

packages/content/src/enrichments.ts
```diff
@@ -0,0 +1,443 @@
...
+    return [...admin, ...member, ...other].map(group => group.id);
+  });
+};
+
+const fetchMetadata = (
+  content: IHubContent,
+  requestOptions?: IHubRequestOptions
+) => getContentMetadata(content.id, requestOptions);
+
+const fetchOwnerOrgId = (
```
rweber-esri, line 245: can we expose this? I need for [PR-9667](https://github.com/ArcGIS/opendata-ui/pull/9667/files#diff-076d4969300d0bb9d112746b1dfc71f68ff2e0abe1ecde160cd68e21885f4f6fR258) unless changes in this PR will result in `orgId` being present on content created from `composer` datasets.

packages/content/src/enrichments.ts
```diff
@@ -0,0 +1,443 @@
...
+  return (
+    content.type === "Web Map" &&
+    contentTypeKeywords.some(
+      typeKeyword => hubTypeKeywords.indexOf(typeKeyword) > -1
+    )
+  );
+};
+
+const shouldFetchOrgId = (content: IHubContent) => {
+  return !content.orgId && isHubCreatedContent(content);
```
rweber-esri, line 299: Maybe we don't need to expose `fetchOwnerOrgId` if this handles detecting content coming from composer?

rweber-esri (approved):
Looks good, only minor comments.

mjuniper (commented):
Love it, nice work! Thanks, Tom!

packages/content/src/enrichments.ts
tomwayson: Yeah, seems to me our rules got a little looser a while back, I assume when we added prettier

packages/content/src/enrichments.ts
tomwayson: so, I swear I left a comment about this last night, but it seems to have gotten lost.

I definitely can, but you can also achieve the same w/ `const { orgId } = fetchEnrichments(content, { ...requestOptions, enrichments: ['orgId'] }`

That is a little awkward, so if you think it's better just to expose each function, I can do that too.

packages/content/src/enrichments.ts
tomwayson: this does not detect content coming from composer.

PR #545, opened 2021-06-15
Content family and org

tomwayson:
This PR gets closer to resolving #322. In essence, everything we need should be there _if_ the content was fetched from the Hub API. If the content is coming from portal, we still have to add logic to `enrichContent()` to fetch additional enrichments such as `server`, `layers`, `recordCount`, etc.

This PR also makes a few fixes to the way `itemToContent` and `datasetToContent` were handling `hubId` and `isPortal`. TLDR: we should only set `hubId` if we are fairy confident that it is in the index. Also, `isPortal` is not returned by the API, we'll need to set it (and the proxied `url`) in `enrichContent()` before calling `fetchEnrichments()`.

In service of those, the PR also moves several low level utils from the Hub app to Hub.js (getFamily, getItemHubId, etc), and it also re-orders how properties are set in `datasetToConent()` to logically group them (common, server, layer). So it may seem like a lot of 🧀 was moved.

### DIFF
--- a/packages/common/src/content.ts
+++ b/packages/common/src/content.ts
@@ -1,6 +1,6 @@
 /* Copyright (c) 2019 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
-
+import { IItem } from "@esri/arcgis-rest-portal";
 import { collections } from "./collections";
 import { categories } from "./categories";
 import { includes } from "./utils";
@@ -138,3 +138,55 @@ export function getCollection(itemType: string = ""): string {
     }
   }
 }
+
+/**
+ * Case-insensitive check if the type is "Feature Service"
+ * @param {string} type - item's type
+ * @returns {boolean}
+ */
+export const isFeatureService = (type: string) => {
+  return type && type.toLowerCase() === "feature service";
+};
+
+/**
+ * parse layer id from a service URL
+ * @param {string} url
+ * @returns {string} layer id
+ */
+export const getLayerIdFromUrl = (url: string) => {
+  const endsWithNumberSegmentRegEx = /\/\d+$/;
+  const matched = url && url.match(endsWithNumberSegmentRegEx);
+  return matched && matched[0].slice(1);
+};
+
+/**
+ * return the layerId if we can tell that item is a single layer service
+ * @param {*} item from AGO
+ * @returns {string} layer id
+ */
+export const getItemLayerId = (item: IItem) => {
+  // try to parse it from the URL, but failing that we check for
+  // the Singlelayer typeKeyword, which I think is set when you create the item in AGO
+  // but have not verified that, nor that we should alway return '0' in that case
+  return (
+    getLayerIdFromUrl(item.url) ||
+    (isFeatureService(item.type) &&
+      includes(item.typeKeywords, "Singlelayer") &&
+      "0")
+  );
+};
+
+/**
+ * given an item, get the id to use w/ the Hub API
+ * @param item
+ * @returns Hub API id (hubId)
+ */
+export const getItemHubId = (item: IItem) => {
+  if (item.access !== "public") {
+    // the hub only indexes public items
+    return;
+  }
+  const id = item.id;
+  const layerId = getItemLayerId(item);
+  return layerId ? `${id}_${layerId}` : id;
+};
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -270,8 +270,12 @@ export interface IEnrichmentErrorInfo {
 export interface IHubContent extends IHubResource, IItem {
   /**
    * The content's ID for use with the Hub API
+   * For most content this will be the item's id
+   * For layers this will be `<itemId>_<layerId>`
+   * This will be undefined for private items and in enterprise
+   * because only public online items are included in the Hub API
    */
-  hubId: string;
+  hubId?: string;
   // NOTE: we may want to elevate this to IHubResource if it's needed for other subtypes
   /**
    * Content visibility and access control, including groups
--- a/packages/common/src/types.ts
+++ b/packages/common/src/types.ts
@@ -141,6 +141,7 @@ export type IRevertableTaskResult =
 
 /**
  * Types of Hub resources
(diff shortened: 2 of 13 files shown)


packages/common/src/types.ts
```diff
@@ -319,11 +348,25 @@ export interface IHubContent extends IHubResource, IItem {
   data?: {
     [propName: string]: any;
   };
+  // service types: Feature Service, Map Service
+  /** service information (currentVersion, capabilities, maxRecordCount etc) */
+  server?: Partial<IFeatureServiceDefinition>;
```
dbouwman, line 353: Should this be `service`? vs `server`?

tomwayson: the Hub API returns `server`, so we rollin' w/ that.

packages/common/src/types.ts
```diff
@@ -319,11 +348,25 @@ export interface IHubContent extends IHubResource, IItem {
...
+  /** layer information (geometryType, fields, etc) for related layers in the service */
+  layers?: Array<Partial<ILayerDefinition>>;
+  // layer types: Feature Layers, Raster Layers
+  /** layer information (geometryType, fields, etc) */
+  layer?: Partial<ILayerDefinition>;
```
dbouwman, line 358: Thoughts on just using `layers` and if there is only one entry... treating it like a `layer`?

packages/common/src/types.ts
```diff
@@ -319,11 +348,25 @@ export interface IHubContent extends IHubResource, IItem {
...
+  recordCount?: number;
```
dbouwman, line 359: how does `recordCount` work w/ `layers`?

packages/content/src/hub.ts
```diff
@@ -269,8 +297,13 @@ export function datasetToItem(dataset: DatasetResource): IItem {
     id: itemId,
     owner: owner as string,
     orgId,
-    created: (created || createdAt) as number,
-    modified: (modified || updatedAt) as number,
+    created: created as number,
```
mjuniper, line 300: I sorta thought createdAt came from the hub api but i don't see it. Are we sure we will never have createdAt?

mjuniper (approved):

packages/content/src/hub.ts
mjuniper: Hmm thinking more about this, i guess created at was from our composer service.

tomwayson: it comes from the model serializer (or composer). I've tested this locally sending models that come out of composer in the app's `datasetModelToContent()` and it does the right thing.

tomwayson: I have a massive spreadsheet that maps all the attributes returned from the API for various content types and maps them to the dataset model, which is a superset w/ CPs, and such.

packages/common/src/content.ts
```diff
@@ -138,3 +138,55 @@ export function getCollection(itemType: string = ""): string {
...
+ */
+export const getLayerIdFromUrl = (url: string) => {
+  const endsWithNumberSegmentRegEx = /\/\d+$/;
+  const matched = url && url.match(endsWithNumberSegmentRegEx);
+  return matched && matched[0].slice(1);
+};
+
+/**
+ * return the layerId if we can tell that item is a single layer service
+ * @param {*} item from AGO
```
rweber-esri, line 164: Minor, comment formatting. Outside scope of this PR, but have we ever explored adding `eslint-plugin-jsdoc` or similar for enforcing consistent commenting/docing styles?

dbouwman (approved):
Not sure I quite follow all of that, but seems legit :)

packages/common/src/types.ts
tomwayson: so, `layer` represents the one layer out of all the ones on the server that you are currently looking at.  If this we an instance of a class, we'd have a `layerId` property and a layer getter like `this.layers.filter(layer => layer.id === this.layerId)`.

packages/content/src/hub.ts
```diff
@@ -269,8 +297,13 @@ export function datasetToItem(dataset: DatasetResource): IItem {
...
+    // for feature layers, modified will usually come from the layer so
+    // we prefer itemModified, but fall back to modified if it came from the item
+    modified: (itemModified ||
```
rweber-esri, line 303: The nested ternary is hard to read, might be easier on the eyes written as

```js
modified: (itemModified || (modifiedProvenance === "item.modified" && modified) as number,
```

packages/common/src/types.ts
tomwayson: `recordCount` only refers the the "current" layer (see https://github.com/Esri/hub.js/pull/545#discussion_r652014708). and is not returned for service types (`Feature Service`, `Map Service`), only for the `Feature Layer` pseudo type.

tomwayson: i.e it must be re-fetched if you switch layers

rweber-esri (approved):
Looks good to me, only minor comments.

packages/common/src/content.ts
tomwayson: so you are saying that the doc comments shouldn't have types, right? b/c that is taken care of by typedoc?

packages/content/src/hub.ts
tomwayson: I agree, and that's actually how I had it written originallyI.  think I when I changed it I was being preemptively defensive against the insults TS was going to lob at me for having the audacity to assign `false` to a `number`. Resolved in 02d08c69e336e9e960e7ecd3d441d9148996fb9a

packages/common/src/content.ts
rweber-esri: I hadn't considered that. I guess it'd just be nice if our inline comments were aligned w/ the docs generated by typedoc.

tomwayson: I'm not sure what's out of alignment, either specifically w/ this comment (other than I think it should not have the types), or in general, but if there are improvements w/ can make w/ `eslint-plugin-jsdoc` I say open an issue.

PR #547, opened 2021-06-17
handle trial, et al sub types wrt group creation

benstoltz:
https://devtopia.esri.com/dc/hub/issues/954 

Issue: Online isn't properly respecting addExternalMembersToGroup for certain subscription types. We need to work around this to allow site creation for those types (trial, personal use, etc)


benstoltz:
@dbouwman I didn't notice the package.lock changes till now. Suggestions on what to do or are those changes fine?

packages/teams/src/utils/filter-user-privs-by-subscription-type.ts
```diff
@@ -0,0 +1,40 @@
...
+  "Education Site License",
+  "Education",
+  "HUP Online"
+];
+
+/**
+ * Returns a cloned copy of the user object with updated privileges
+ * based on whether or not the user has a subscription type not in the
+ * allowed list
+ * @param {object} user
```
dbouwman, line 20: Add a note that if/when AGO implements this logic on their side, we can remove this function or simply return the passed in user

packages/teams/src/utils/update-template-based-on-privs.ts
```diff
@@ -0,0 +1,23 @@
+import { IGroupTemplate } from "../types";
+import { IUser } from "@esri/arcgis-rest-types";
+import { cloneObject, includes } from "@esri/hub-common";
+
+export function updateTemplateBasedOnPrivs(
```
dbouwman, line 5: comment needed. also can we name this something that connects it to the `privPropValues` hash? i.e. `applyPrivPropValues(..)`?

packages/teams/src/utils/update-template-based-on-privs.ts
```diff
@@ -0,0 +1,23 @@
...
+  user: IUser,
+  template: IGroupTemplate
+): IGroupTemplate {
+  const copyTemplate = cloneObject(template);
+  // Only two templates actually have this in config atm, so we want to be safe.
+  if (template.config.privPropValues && template.config.privPropValues.length) {
```
dbouwman, line 11: make this less `if`y
```
const ppv = getWithDefault(template, 'config.privPropValues', []);
ppv.forEach(...
```

packages/teams/src/utils/update-template-based-on-privs.ts
```diff
@@ -0,0 +1,23 @@
+import { IGroupTemplate } from "../types";
+import { IUser } from "@esri/arcgis-rest-types";
+import { cloneObject, includes } from "@esri/hub-common";
+
+export function updateTemplateBasedOnPrivs(
+  user: IUser,
+  template: IGroupTemplate
+): IGroupTemplate {
+  const copyTemplate = cloneObject(template);
```
dbouwman, line 9: `copyTemplate` sounds like a function, `templateCopy` sounds better

packages/teams/src/utils/get-user-creatable-teams.ts
```diff
@@ -4,27 +4,38 @@ import { WELLKNOWNTEAMS } from "../well-known-teams";
...
+ * @param {string} subscriptionInfoType
  */
 export function getUserCreatableTeams(
   user: IUser,
   environment: HubProduct,
-  portalApiVersion: string
+  portalApiVersion: string,
+  subscriptionInfoType: string
 ): IGroupTemplate[] {
   // TODO: remove this when needed (after may 1st 2021)
```
dbouwman, line 24: Since we are after May 1, can we do the removal? i.e. nix the PRE_9_1_WELLKNOWNTEAMS file, and this block of logic?

packages/teams/src/utils/get-user-creatable-teams.ts
```diff
@@ -4,27 +4,38 @@ import { WELLKNOWNTEAMS } from "../well-known-teams";
...
   // choosing the type of well known team based on the current portal version
   const teams =
     parseFloat(portalApiVersion) < 9.1
       ? PRE_9_1_WELLKNOWNTEAMS
       : WELLKNOWNTEAMS;
+  // Online is not properly allowing certain privs certain abilities for certain sub types
```
dbouwman, line 30: Let's be very specific in this comment, or don't have a comment.

packages/teams/src/utils/filter-user-privs-by-subscription-type.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param {string} subscriptionInfoType
+ * @returns
+ */
+export function filterUserPrivsBySubscriptionType(
```
dbouwman, line 24: thinking about what this _does_, it's removing invalid privs from the user, based on the subscription type... so lets call it that... 
`removeInvalidPrivs(user, subscriptionType)`

packages/teams/src/utils/get-user-creatable-teams.ts
```diff
@@ -4,27 +4,38 @@ import { WELLKNOWNTEAMS } from "../well-known-teams";
...
  * @param {string} environment
+ * @param {string} portalApiVersion
+ * @param {string} subscriptionInfoType
  */
 export function getUserCreatableTeams(
   user: IUser,
   environment: HubProduct,
-  portalApiVersion: string
+  portalApiVersion: string,
+  subscriptionInfoType: string
```
dbouwman, line 22: tbh - not super wild about passing in this deeply nested property value. makes testing easy, but seems like it's "leaking" an abstraction. Since it's currently a breaking change (i.e. we're adding an argument, that does not have a default value, ergo clients that don't pass in a value will break)... either, give this a default value (so it's a minor release) or just ask for an IPortal, vs the version + subscription info.

packages/teams/src/types.ts
```diff
@@ -28,9 +28,16 @@ export interface IGroupTemplate extends Partial<IGroup> {
     titleI18n: string;
     descriptionI18n: string;
     snippetI18n: string;
+    privPropValues?: IPrivPropValues[];
   };
 }
 
+export interface IPrivPropValues {
```
dbouwman, line 35: this will be in the docs, so please add comment

dbouwman (changes requested):
lots of comments :) 

I'd like it if we could release this as a minor bump, so let's ensure we have default values for the subscriptionInfo.type argument on `getUserCreatableTeams`

Also leave a comment re: breaking changes, as it would be _ideal_ to take the portalSelf or IHubRequestOptions on that fn, and have the fn pluck the right values out of the deep hash vs making a consumer know where to find these two things.

packages/teams/src/utils/get-user-creatable-teams.ts
benstoltz: As discussed going with the default empty string value.

packages/teams/src/utils/get-user-creatable-teams.ts
benstoltz: Will be done in it's own PR.

dbouwman (approved):

PR #551, opened 2021-06-23
1263/move is valid extent

juliannaeapicella:
## Issue Link(s)
https://devtopia.esri.com/dc/hub/issues/1263

## Description
Added isValidExtent and necessary helper functions into Hub.js.


packages/common/src/extent/index.ts
```diff
@@ -2,3 +2,35 @@ export * from "./get-geographic-org-extent";
...
+
+function isExtentCoordinateArray(extent: object) {
+  return (
+    Array.isArray(extent) &&
+    Array.isArray(extent[0]) &&
+    Array.isArray(extent[1])
+  );
+}
+
+function isExtentEnvelope(extent: any) {
```
andrewctate, line 14: Why are some of these arguments of type `any`, and some of type `object`?

tomwayson (commented):
Well done @juliannaeapicella 

I'm surprised that you didn't need more tests of null cases (`undefined`, `[]`) to get 100% coverage, but if CI is happy, so am I.

I'd like to add a commit that puts these fns into use here: 

https://github.com/Esri/hub.js/blob/399acaee02f482e590be4f23bc9806a95523b83c/packages/content/src/hub.ts#L165

That should be 

```ts
if (!isExtentCoodinateArray(item.extent)) && extent && isExtentCoodinateArray(extent.coordinates)) { 
```

I realize that means we need to export that function. It also means we can drop `isExtentEnvelope()` and related tests.

andrewctate (changes requested):
First off—great work. You've taken a 🤿 into the world of TypeScript!

However, we should move the functions to their own file(s). We use the `index.ts` files in Hub.js mainly for simplifying import statements in the consuming application. For example, without the [export statement](https://github.com/Esri/hub.js/pull/551/files#diff-e683ee82a61ee5ca12a797e2b255b3851e1527780535089f0cc291fa2afd964aR2) for the `extent-to-bbox` utility, importing it would look like

```js
import { extentToBbox } from '@esri/hub-common/extent/extent-to-bbox`;
```

since we export it from `index.ts` at both the extent and `src` levels, the import can be simplified:

```js
import { extentToBbox } from '@esri/hub-common`;
```

So, instead of putting them in `index.ts` (which technically works but is against our convention), let's create a new file at `packages/common/src/is-valid-extent.ts`, put the functions there, and re-export them from the `index.ts`

```js
export * from "./is-valid-extent";
```

tomwayson:
I asked her to put the fns in `index.ts` - I am pretty strongly against the convention we have adopted in some parts of this repo of one fn per file. It negates one of the benefits of modules in that you can have truly private fns. The bigger the surface you expose (export) the larger your contract w/ consumers, and the harder it is to make non-breaking changes.

andrewctate:
@tomwayson I agree with you. But, to clarify, my suggestion was to create an `is-valid-extent.ts` file and put _all_ the functions she's moved over in this PR in there, with only `isValidExtent` being exported. 

Either way, I defer to you.

andrewctate (commented):
Leaving it to @tomwayson

andrewctate (dismissed):

packages/common/src/extent/index.ts
```diff
@@ -2,3 +2,35 @@ export * from "./get-geographic-org-extent";
...
+    key => typeof extent[key] === "number"
+  );
+}
+
+/**
+ * Check if the given extent is in a known format
+ * @param  {Object} extent extent in any format
+ * @return {Boolean}       indicator
+ */
+export function isValidExtent(extent: object) {
```
dbouwman, line 29: Can we move this fn (and it's deps) to it's own file vs adding into index.ts?

tomwayson: for this PR we're going to replace the `src/extent` folder w/ `src/extent.ts` by moving these fns and all fns in `/extent/*.ts` files into the new file.

packages/content/src/hub.ts
```diff
@@ -9,6 +9,7 @@ import {
 import { itemToContent } from "./portal";
 import { isSlug, addContextToSlug, parseDatasetId } from "./slugs";
 import { enrichContent, IFetchEnrichmentOptions } from "./enrichments";
+import { isExtentCoordinateArray } from "../../common/src/extent";
```
tomwayson, line 12: When we import across package boundaries we should import from the package name:

```suggestion
import { isExtentCoordinateArray } from "@esri/hub-common";
```

tomwayson (commented):
great work, just one small change and we're good to merge!

tomwayson (approved):
🚀

PR #554, opened 2021-06-24
Update well known teams to default to org membershipAccess and update…

benstoltz:
… it if they have privs

We ended up needing to repeat the pattern from core teams for the other well known teams that rely on addExternalMembersToGroup. This finishes those out.

### DIFF
--- a/packages/teams/test/utils/get-user-creatable-teams.test.ts
+++ b/packages/teams/test/utils/get-user-creatable-teams.test.ts
@@ -10,9 +10,11 @@ describe("getUserCreatableTeams", () => {
         "opendata:user:designateGroup"
       ]
     };
+
     expect(
       getUserCreatableTeams(user, "premium", "8.4", "In House").length
     ).toBe(5, "Premium, all privs, should be 5 creatable team types");
+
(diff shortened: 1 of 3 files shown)


packages/teams/test/utils/get-user-creatable-teams.test.ts
```diff
@@ -60,50 +61,62 @@ describe("getUserCreatableTeams", () => {
...
     // remove updateGroups...
     user.privileges = [
       "portal:user:createGroup",
       "opendata:user:designateGroup",
       "portal:user:addExternalMembersToGroup"
     ];
+
     expect(
       getUserCreatableTeams(user, "premium", "9.1", "In House").length
     ).toBe(4, "Premium, all privs, should be 5 creatable team types");
```
dbouwman, line 87: This message seems wrong... user does not have all privs, and only 4 groups can be created

dbouwman (approved):
pls review the tests and ensure the messages are correct. otherwise, gtg

dbouwman:
Unclear why fetchMocks are failing, but this should be good to go regardless

dbouwman (approved):

PR #555, opened 2021-06-24
chore: add doc and commit-lint

dbouwman:
Number of chores:
- add commitlint deps and configuration
- add husky commit_msg hook
- add doc on how to work with hub.js for Hub devs
- add PR template (although not clear why that's not used w/ this PR)
- drop existing GH Action workflows and move to same flows as rest-js
- drop node-sass (only used for docs) and use css vs sass
   - this drops int `npm install` from ~6 minutes to ~1min, and we really were not using the sass anyhow
- run chrome & ff tests _first_ so we "fail fast". They take ~30s, where as `jasmine` can take ~6min to build and test
- use unique ids in download tests to we can figure out specifically which one has the issue


tomwayson (approved):
heroic work @dbouwman!

PR #556, opened 2021-06-25
Added getContentIdentifier() and moved all slug helpers from @hub-content to @hub-common

sonofflynn89:
Note: all slug helpers in hub content had deprecation notices added onto them.


packages/content/src/slugs.ts
```diff
@@ -2,7 +2,7 @@ import { isGuid } from "@esri/hub-common";
 
 /**
  * Parse item ID and layer ID (if any) from dataset record ID
- *
+ * **DEPRECATED: import from hub-common instead**
```
tomwayson, line 5: we want to remove the implementations in this file, so it should be just a deprecation message and then:

```ts
export { parseDatasetId, isSlug, addContextToSlug, removeContextFromSlug } from `@esri/hub-common`;
```

We'll then want to remove slugs.test.ts too.

tomwayson (changes requested):
Great work @sonofflynn89!

Just need to re-export the moved fns and remove the implementations and tests from in hub-content.

sonofflynn89:
Issues with github actions, creating a copy branch

PR #576, opened 2021-07-21
One Approach

andrewctate:


packages/search/src/content/index.ts
```diff
@@ -89,8 +102,15 @@ export class ContentSearchService
  * ```
  */
 export function searchContent(
-  request: IContentSearchRequest = { filter: {}, options: {} }
+  request: IContentSearchRequest = { filter: {}, options: {} },
+  hubRequestOptions: IHubRequestOptions
```
andrewctate, line 106: Now you _can_ pass in your own `IHubRequestOptions`, but if you don't it will do its best to build one for you.

packages/sites/src/domains/_lookup-portal.ts
```diff
@@ -9,7 +10,7 @@ import { IHubRequestOptions, includes } from "@esri/hub-common";
  */
 export function _lookupPortal(
   hostname: string,
-  hubRequestOptions: IHubRequestOptions
+  requestOptions: IRequestOptions
```
andrewctate, line 13: Downgraded this call sig, since all this function really needs is an `IRequestOptions`

packages/search/src/content/index.ts
dbouwman: I think this is reasonable

PR #577, opened 2021-07-22
refactor(hub-sites): move domain utils to hub-common

andrewctate:
affects: @esri/hub-common, @esri/hub-sites

1. Description:
I need to use `lookupDomain` in `@esri/hub-search` for some upcoming Hub.js work. This moves everything to hub-common so that I don't have to add a new peer dependency to the hub-search package.

2. Instructions for testing:
`npm test`

3) Screenshot/GIF:
n/a (why do we have this?)

4) Closes Issues: none

5) [x] ran commit script (`npm run c`)


packages/sites/src/index.ts
```diff
@@ -1,7 +1,7 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-export * from "./domains";
+export * from "@esri/hub-common";
```
dbouwman, line 4: Something about this... seems problematic... this will export all of hub common in hub-sites... is there some way to scope this?

tomwayson (approved):
Thanks for taking on this burden @andrewctate

dbouwman (commented):
We need @tomwayson's input on this... I like re-exporting the domain fns from hub-sites to avoid a breaking change, but the way it's done here, it is exporting all of hub-common, not just the domain fns... I'm not clear if that's "ok" or opening up future problems.

tomwayson:
looking

packages/sites/src/index.ts
tomwayson: oh yeah, this is bad - I suspect @andrewctate was just doing a global find and replace on `from "./domains"`. Just need to undo that in this instance

packages/sites/src/domains/index.ts
```diff
@@ -1,19 +1,30 @@
...
-export * from "./is-domain-used-elsewhere";
-export * from "./is-valid-domain";
-export * from "./lookup-domain";
-export * from "./remove-domain";
-export * from "./types";
-export * from "./update-domain";
-export * from "./_ensure-safe-domain-length";
-export * from "./ensure-unique-domain-name";
+export {
+  /* istanbul ignore next */
```
tomwayson, line 2: Bummer you needed these ignore comments.

tomwayson (changes requested):
Just make the one change on the exports that @dbouwman caught (good catch Dave!)

packages/sites/src/domains/index.ts
andrewctate: Yeah, I tried to avoid. You can see that I wrote a test to cover this file, but it's no good.

packages/sites/src/index.ts
andrewctate: Thanks for noting this. Definitely unintentional.

tomwayson (approved):
🚀

PR #582, opened 2021-07-23
search content scopes to a site

andrewctate:
1. Description:
This PR adds the ability to scope a `searchContent` request to a particular site's catalog.

2. Instructions for testing:
`npm test`

4) Closes Issues: https://devtopia.esri.com/dc/hub/issues/1475

5) [x] ran commit script (`npm run c`)


.github/pull_request_template.md
```diff
@@ -1,12 +1,10 @@
...
-2. Instructions for testing:
+1. Instructions for testing:
 
-3) Screenshot/GIF:
+1. Closes Issues: #<number> (if appropriate)
 
-4) Closes Issues: #<number> (if appropriate)
-
-5) [ ] ran commit script (`npm run c`)
+1. [ ] ran commit script (`npm run c`)
```
andrewctate, line 7: Question: do we even need this? I see that we now have a CI check.

packages/search/src/types/content.ts
```diff
@@ -61,6 +61,7 @@ export interface IContentSearchFilter {
  * Properties correspond to their Portal API or arcgis-rest-js equivalents
  */
 export interface IContentSearchOptions {
+  site?: string;
```
andrewctate, line 64: switch to "domain?" "siteDomain"?

packages/search/src/content/index.ts
```diff
@@ -88,9 +125,17 @@ export class ContentSearchService
...
-export function searchContent(
+export async function searchContent(
   request: IContentSearchRequest = { filter: {}, options: {} }
 ): Promise<IContentSearchResponse> {
+  const siteCatalog = await getSiteCatalogFromOptions(request.options);
+  if (siteCatalog) {
+    request.filter = {
+      ...request.filter,
+      ...catalogToContentFilter(siteCatalog),
+    };
```
andrewctate, line 136: This will overwrite any `group` array that was passed in on the `request.filter`. I'm thinking that's the right behavior to avoid unpredictable scoping. Open for feedback, though.

sonofflynn89 (approved):
LGMT, nj! Let's wait for @tomwayson or @dbouwman to weigh in before merging, though.

packages/search/src/content/index.ts
dbouwman: Only scenario I can think of would be showing the content in a specific team, connected to a site... in our app we could just construct a IContentFilter that's specific to the team, or use `team.searchContent(...)` which will be part of the coarse-grained api. Long way of saying I think this is a good way to start

packages/common/src/types.ts
```diff
@@ -458,3 +458,8 @@ export interface IDomainEntry {
   sslOnly: boolean;
   updatedAt?: string;
 }
+
+export interface ISiteCatalog {
```
dbouwman, line 462: doc comments on interface and members

packages/search/src/content/index.ts
```diff
@@ -131,11 +176,32 @@ function performHubContentSearch(
...
   }).then((response: any) =>
     convertHubResponse(requestParams, response, authentication)
   );
 }
+
+function getSiteCatalogFromOptions(options: IContentSearchOptions) {
+  if (!options || !options.site) return null;
+
+  const ro = getHubRequestOptions(options);
+
```
dbouwman, line 192: We need to add logic here to accept a wide range of values for `site`:
- `opendata.dc.gov` ideal case
- '3adff78f286e4ecf936bdefdfeca9890' (isGuid(...) - skip domain lookup, AGO & Ent)
- `http://opendata.dc.gov` - drop protocol
- `https://opendata.dc.gov/other/route` - drop protocol and split off trailing paths
- `beta-dcgis` (Enterprise) - ensure no `.`'s

packages/search/src/types/content.ts
dbouwman: leave as site, as that works for enterprise and ago, and allows a range of things to be passed in

dbouwman (changes requested):
This looks great and nails your needs - in talking with Andrew re `site` he also wants it to be very open in terms of what it accepts... see comment. Please add that and then I think this is good to go.

packages/search/src/content/index.ts
andrewctate: @dbouwman thanks for pointing this out.

packages/search/src/content/index.ts
```diff
@@ -76,6 +86,33 @@ export class ContentSearchService
...
+ *   orgId: 'be55891b4'
+ * };
+ *
+ * const filter: IContentSearchFilter = catalogToContentFilter(siteCatalog);
+ * ```
+ *
+ * @param catalog - The site catalog
+ * @returns
+ */
+export function catalogToContentFilter({
```
tomwayson, line 106: I don't want to export this at this time, see https://devtopia.esri.com/dc/hub/issues/1475#issuecomment-2960329

The params it's expecting are not a catalog. I would call this `siteCatalogToContentFilter()` since it takes `site.data.values.catalog`.

packages/search/src/content/index.ts
tomwayson: This should be the opposite:

```ts
    request.filter = {
      ...catalogToContentFilter(siteCatalog),
      ...request.filter,
    };
```

I'm pretty sure we always prefer explicitly passed in params over any any params that come from context, in this case the site.

In our glorious future, this fn will be exectued from an instance of a site that will provide all the context, but also allow users to pass in overrides to that context.

tomwayson (changes requested):
Great work @andrewctate 

A couple of minor (but important) changes. I also agree w/ @dbouwman  on docing the params.

packages/search/src/content/index.ts
tomwayson: actually, this fn doesn't do anything and probably never will if `site.data.values.catalog` is only ever `orgId` and/or `groups`.

Let's drop this fn and just do `request.filter = { ...site.data.values.catalog, ...requst.filter }`

Later we can add a different fn for converting a ["platform" catalog](https://github.com/ArcGIS/Hub/pull/54/files)  to `IContentSearchRequest`  b/c I think some things in `catalog.params` will go in `request.filter` and other things will go in `request.options`

packages/search/src/content/index.ts
tomwayson: I think this calls for a standalone `fetchSite(idenifier: string): ISiteModel` function where `identifier` is an id (guid), domain (hostname), URL, or enterprise id(, slug?) that does all that. Then this fn calls `fetchSite(options.site)`.

dbouwman (approved):
Looks good to me! Thanks @andrewctate

packages/common/test/sites/index.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(getSiteSpy).toHaveBeenCalledWith(siteId, requestOptions);
+  });
+
+  it("accepts a site slug", async () => {
+    const portalRO = { isPortal: true } as IHubRequestOptions;
+    const chk = await fetchSite("okokokko", portalRO);
+    expect(chk).toEqual(site);
+    expect(lookupDomainSpy).toHaveBeenCalledWith("okokokko", portalRO);
+    expect(getSiteSpy).toHaveBeenCalledWith(siteId, portalRO);
+  });
```
andrewctate, line 65: @dbouwman to your eyes does this test case cover what it needs to for Enterprise Sites?

dbouwman: I think that's reasonable... `lookupDomain` is responsible for sorting out the enterprise vs ago logic

packages/discussions/src/utils/posts.ts
```diff
@@ -15,7 +15,13 @@ export function parseDiscussionURI(discussion: string): IDiscussionParams {
...
   const type = uri.hostname;
   const [, identifier] = uri.pathname.split("/");
-  const [id, layer = null] = identifier.split("_");
+  let id, layer;
+  if (!identifier) {
+    id = null;
+    layer = null;
+  } else {
+    [id, layer = null] = identifier.split("_");
+  }
```
tomwayson, line 24: is there a reason to not use [parseDatasetId()](https://esri.github.io/hub.js/api/common/parseDatasetId/) here?

tomwayson (approved):
great work @andrewctate!

My only question is about the discussion package changes. they seem unrelated, and I just want to ensure they are intentional.

Issue #543, opened 2021-06-10
expand orgId enrichement to other content types, and/or replace with enrichment w/ ownerUser?


tomwayson:
Hub.js currently only fetches the orgId for a content for Hub-created web maps (we were being conservative in order to avoid making this xhr when we didn't need it).

However, now I've learned that the Hub app's content views send `contentOrgId` w/ [telemetry](https://github.com/ArcGIS/opendata-ui/pull/9667/files#diff-1d76a9b7b87d42647bae7871365770cf3b84982ecc6844e51652b92d0d5beff6R13). The app fetches the `ownerUser` for _all content types_ and we get the id from `content.ownerUser.org.id`.

See: https://github.com/ArcGIS/opendata-ui/pull/9667#issuecomment-858841298 for more info.

Determine if hub app needs the whole `ownerUser`, or just the `orgId` and move that logic into `fetchEnrichments`.

cc @rweber-esri @juliannemarik

tomwayson:
I forgot about #414, so looks like we need at least the user's name in addition to the org id, so sounds like we should update the enrichment to set `ownerUser` and not just `orgId`. We can then also set `orgId` (and perhaps other org properties)? later in `enrichProperties`.

Also, I guess since we are

tomwayson:
Plan:
- replace `orgId` enrichment w/ a new `ownerUser` enrichement
- after enrichments are fetched, call a new `_processFetchedEnrichments` fn instead of just `_enrichDates`
- `_processFetchedEnrichments` will now call `_enrichDates` and then also have new logic for updating the content's `orgId` and `org` properties based on `ownerUser`

NOTE: that the current app code for fetching the owner user is here:

https://github.com/ArcGIS/opendata-ui/blob/master/packages/opendata-ui/app/services/content.js#L307-L332

That code not only fetches the user, but enriches the user object and makes a few additional requests for org info, etc. TBD how much of that needs to be done w/in `enrichContent()`.

dbouwman:
# Enrichment Streamlining

Single entry fn

```
enrichContent(
  type:string,
  content:Partial<IHubContent>,
  requestOptions: IHubRequestOptions
  ):Promise<IHubContent> {...}
```

For composition, we should have a single input and return type for all the enrichment functions

```
export interface IEnrichmentOpts {
  content: Partial<IHubContent>,
  requestOptions?: IHubRequestOptions
}
```

Each enrichment should handle merging in any errors so we have no processing "at the end"

[...]

PR #586, opened 2021-07-27
refactor(hub-content): refactor fetchContent to use createOperationPi…

tomwayson:
…peline

1. Description:

what it says on the tin

1. Instructions for testing:

1. Closes Issues: n/a. this is the foundation for addressing #543, but will do that in a follow up PR.


packages/content/src/enrichments.ts
```diff
@@ -262,55 +259,99 @@ export function _enrichDates(content: IHubContent): IHubContent {
...
-): Promise<string[]> => {
-  return getItemGroups(content.id, requestOptions).then((response) => {
-    const { admin, member, other } = response;
-    return [...admin, ...member, ...other].map((group) => group.id);
-  });
+const enrichGroupIds = (
+  input: IPipeable<IHubContent>
+): Promise<IPipeable<IHubContent>> => {
+  const { data: content, stack, requestOptions } = input;
+  const opId = stack.start("enrichGroupIds");
```
dbouwman, line 266: If these are ever going to be run on a list of content objects, i'd suggest using 
```js
...startOperation({ id: `enrichGroupIds-${data.id}`, type: 'enrichGroups', inputs: { id: data.id}})`
```
it's a little more verbose, but you'll be able to trace back to _which_ item was involved in a failure vs just knowing something failed... I'd do this in all of these enrichment fns

packages/content/src/enrichments.ts
```diff
@@ -509,22 +561,20 @@ export const enrichContent = (
...
-        ...portalUrls,
-        ...enrichments,
-        // include any previous errors (if any)
-        errors: [...serverErrors, ...enrichments.errors],
-      };
+  return fetchEnrichments(contentWithPortalUrls, requestOptions).then(
+    (fetched) => {
       // enrich dates now that we have metadata (if any)
-      const enriched = _enrichDates(merged);
+      const enriched = _enrichDates(fetched);
```
dbouwman, line 577: Since this is done for all content, it could be included in the pipeline...

dbouwman: Odd - I can't leave a comment on a line that's not changed in this PR... but I'd also suggest looking at refactoring `getLayerContent` so it's also in the pipeline. As it stands, that fn does a check and early exits... so I think you could do a small refactor to make it part of the pipeline, which makes it all consistent.

dbouwman (changes requested):
Some other tests area failing as a result of implementing this pipeline - I think you need to mock some more things in the `get content from IModel ...` set of tests.

I'd also consider some small refactors to make all aspects of the enrichments run through the pipeline (comments added), and to have the itemId in the operation stack - as part of the operation name, or as inputs... this will make it _much_ easier to track down problems.

packages/content/src/enrichments.ts
tomwayson: It's odd the other tests didn't fail for me locally, but they are now.

packages/content/src/enrichments.ts
dbouwman: As the 😈 's advocate - `searchContent()` is async already, and wrapping a sync fn into a `Promise.resolve()` adds ~1 tick... now if you got 10,000 things back from the api, this may add a measurable amount of time, but it would still be minuscule compared to the first xhr to get the data from the api :)

Anyhow - I'm fine if you rename the fn, and leave it sync... I just see value in things being consistent w/in a sub system.

packages/content/src/enrichments.ts
tomwayson: tests are passing. I think we should merge this as is. I'll be revisiting this code soon for #543 at which point I can do the more verbose `startOperation()` call.

dbouwman (approved):
Huzzah! Thanks Tom!

PR #601, opened 2021-08-06
[DNM] 4.1.1 force POST when Searching on Portal

andrewctate:
1. Description:
Add `httpMethod: 'POST'` to avoid URLs that are too long for Portal-10.8.

1. Instructions for testing: 
Follow the instructions on the issue.

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/1215

1. [ ] ran commit script (`npm run c`)


andrewctate:
@tomwayson we validated that this change is only necessary on the Portal 10.8.1 branch. No need to wrap it back into `master`.

packages/search/src/ago/get-items.ts
```diff
@@ -28,7 +28,10 @@ export async function getItems(
           countSize: agoParams.countSize
         },
         portal,
-        authentication
+        authentication,
+        // this is the magic number of bytes that PORTAL-10.8
+        // allows in a GET request
```
tomwayson, line 33: This is actually probably the magic number for the IIS Web Adaptor on Windows, or maybe even [a setting in IIS](https://www.google.com/search?q=iis+max+url+length&oq=iis+max+url&aqs=edge.0.0i512j69i57j0i512j0i22i30l4.4418j0j1&sourceid=chrome&ie=UTF-8), and not a limitation of portal itself. There are other (Java) versions of the web adapter for Windows and Linux:

image

I would hope that those would support longer URLs.

I'm not saying that we need to get to the bottom of this in this PR, but I think it's worth calling out in the code comments here the OS and Web Adapter version that were reported in the bug(s) and test environment where this problem was observed.

tomwayson (approved):
nicely done. might be worth being more specific in the comments if we have the info I suggest.

PR #613, opened 2021-08-17
docs(hub-search): add inline docs and examples to searchContent()

sonofflynn89:
affects: @esri/hub-search

1. Description:
Documentation-only changes

### DIFF
--- a/packages/search/src/content/index.ts
+++ b/packages/search/src/content/index.ts
@@ -86,10 +86,12 @@ export class ContentSearchService
 }
 
 /**
- * A standalone function for searching content across the Portal API only or Portal API and the
- * Hub Indexer V3 API.
+ * A standalone function for searching Hub content
  *
- * The `filter` argument allows the caller to filter results based on attributes exposed by AGO and
+ * A content search is configured by passing `searchContent` an object of type `IContentSearchRequest`.
+ * This configuration object is composed of two important child objects: `filter` and `options`.
+ *
+ * The `filter` object allows the caller to filter results based on attributes exposed by AGO and
  * the Hub API. A complete list of attributes can be found at the docs for `IContentSearchFilter`,
  * but some examples include:
  * - access
@@ -97,8 +99,8 @@ export class ContentSearchService
  * - group
  * - orgid
  *
- * The `options` argument allows the caller to specify more general attributes about the request itself.
- * A complete list of options can be found at the docs for `IContentSearchFilter`, but some examples include:
+ * The `options` object allows the caller to specify more general attributes about the request itself.
+ * A complete list of options can be found at the docs for `IContentSearchOptions`, but some examples include:
  * - the authenticated user
  * - whether the call is happening in an enterprise environment
  * - what site catalog to search
@@ -116,7 +118,7 @@ export class ContentSearchService
  * const options: IContentSearchOptions = {
  *    site: 'https://my-site.hub.arcgis.com',
  *    portal: 'https://www.arcgis.com',
- *    // Note: only private content visible to the authenticated user will be returned
+ *    // Any private content that the authenticated user can access will be included in the results
  *    authentication: new UserSession(...),
  * }
  * const searchResults = await searchContent({ options });
@@ -131,7 +133,7 @@ export class ContentSearchService
  *    authentication: new UserSession(...),
  * }
  * const filter: IContentSearchFilter = {
- *    access: 'public' // Can also be expressed as { value: ['public'] }
+ *    access: 'public'
  * }
  * const searchResults = await searchContent({ filter, options });
  * ...
@@ -147,7 +149,7 @@ export class ContentSearchService
  * const options: IContentSearchOptions = {
  *    site: 'https://my-site.hub.arcgis.com',
  *    portal: 'https://www.arcgis.com',
- *    // Note: only private content visible to the authenticated user will be returned
+ *    // Any private content that the authenticated user can access will be included in the results
  *    authentication: new UserSession(...),
  * }
  * const filter: IContentSearchFilter = {
@@ -167,7 +169,7 @@ export class ContentSearchService
  *    authentication: new UserSession(...),
  * }
  * const filter: IContentSearchFilter = {
- *    id: 'my_item_id' // Can also be expressed as { value: ['my_item_id'] }
+ *    id: 'my_item_id'
  * }
  * const searchResults = await searchContent({ filter, options });
  * ```
@@ -177,10 +179,17 @@ export class ContentSearchService
  * items that match a given query, you'll need to utilize the `hasNext` flag as well as the
  * `next` function included on the return object and make multiple XHR requests.
  *
- * 2) There are differences in item structure for results returned from the
- * hub index vs AGO/Enterprise (i.e. 'public' items vs 'private' items).
- * Here is an example of a 'private' map result:
+ * 2) Results returned from the hub index (i.e. 'public items') will be structured differently than
+ * results returned from AGO/Enterprise (i.e. 'private' items). A comprehensive list of differences cannot
+ * be given here, but developers should be aware that they exist. Additionally, the structure of 'public'
+ * items can change when either the indexing process or schema of the Hub API is modified. To showcase
+ * some of the possible differences between 'public' and 'private' items, we've provided examples of both
+ * types of results below:
+ *
  * ```js
+ * ///////////////
+ * // Private
+ * ///////////////
  * {
  *  access: "myself",
  *  appCategories: [],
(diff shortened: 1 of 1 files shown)


packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
  * ```
+ * There are a couple gotchas that need to be accounted for:
+ *
+ * 1) There is no way to specify the number of results per page. If you need to fetch all
+ * items that match a given query, you'll need to utilize the `hasNext` flag as well as the
+ * `next` function included on the return object and make multiple XHR requests.
+ *
+ * 2) There are differences in item structure for results returned from the
+ * hub index vs AGO/Enterprise (i.e. 'public' items vs 'private' items).
+ * Here is an example of a 'private' map result:
```
sonofflynn89, line 182: This section was added as per @ajturner's comment on the issue:

> How do I search a Site?
Does this include both public + private content?
What are the differences in metadata attributes?

I couldn't find an easy way to see exactly what was different between the two and I'm not sure which attribute differences are significant (other than obvious ones like `slug` and `composeStatus`). We have to remember as well that https://devtopia.esri.com/dc/hub/issues/1713 will change the results into `IHubContents` in the near future

andrewctate:
@sonofflynn89 how can I view these nice docs in their final format?

sonofflynn89:
@andrewctate you have to pull the branch down and run `npm run docs:serve` from the root directory. You might have to do it a couple times cause the build process is janky, but that will serve it locally.

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
+ * The `filter` argument allows the caller to filter results based on attributes exposed by AGO and
+ * the Hub API. A complete list of attributes can be found at the docs for `IContentSearchFilter`,
+ * but some examples include:
+ * - access
+ * - id
+ * - group
+ * - orgid
+ *
+ * The `options` argument allows the caller to specify more general attributes about the request itself.
+ * A complete list of options can be found at the docs for `IContentSearchFilter`, but some examples include:
```
sonofflynn89, line 101: Typo: should be `IContentSearchOptions`

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
  * A standalone function for searching content across the Portal API only or Portal API and the
  * Hub Indexer V3 API.
  *
- * @param request - the IContentSearchRequest instance for searching
+ * The `filter` argument allows the caller to filter results based on attributes exposed by AGO and
```
andrewctate, line 92: I would insert a short intro before jumping into `filter` and `options`. For example
> A content search is configured by passing `searchContent` an object of type `IContentSearchRequest`. This configuration object is composed of two important child-objects, `filter` and `options`.

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
+ * import { searchContent, IBooleanOperator } from '@esri/hub-search';
+ * ...
+ *
+ * /////////////////////////////////////////////////////////////////////
+ * // Search for all public and private content associated with a site
+ * /////////////////////////////////////////////////////////////////////
+ * const options: IContentSearchOptions = {
+ *    site: 'https://my-site.hub.arcgis.com',
+ *    portal: 'https://www.arcgis.com',
+ *    // Note: only private content visible to the authenticated user will be returned
```
andrewctate, line 119: This note could be misinterpreted to mean that no public content will be returned.
```suggestion
 *    // Note: no private content that is unavailable to the authenticated user will be included
```

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
+ * ///////////////////////////////////////////////////////////
+ * // Search for all public content associated with a site
+ * ///////////////////////////////////////////////////////////
+ * const options: IContentSearchOptions = {
+ *    site: 'https://my-site.hub.arcgis.com',
+ *    portal: 'https://www.arcgis.com',
+ *    authentication: new UserSession(...),
+ * }
+ * const filter: IContentSearchFilter = {
+ *    access: 'public' // Can also be expressed as { value: ['public'] }
```
andrewctate, line 134: I'm thinking this comment may be confusing without further context that we get later. I could see someone asking: does it do anything different than just `public`? Why is it an array?

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
  * A standalone function for searching content across the Portal API only or Portal API and the
  * Hub Indexer V3 API.
```
andrewctate, line 90: This feels like too much technical detail for the first sentence. Maybe something like this?
> A standalone function for searching Hub content.

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
+ * ///////////////////////////////////////////////////////////
+ *
+ * // TODO: As of 8/17/21 the hub api has a bug in which
+ * // Any query with a not filter on the access field
+ * // (e.g access: not('public')) will return no results
+ *
+ * const options: IContentSearchOptions = {
+ *    site: 'https://my-site.hub.arcgis.com',
+ *    portal: 'https://www.arcgis.com',
+ *    // Note: only private content visible to the authenticated user will be returned
```
andrewctate, line 150:
```suggestion
 *    // Note: no private content that is unavailable to the authenticated user will be included
```

packages/search/src/content/index.ts
```diff
@@ -89,13 +89,195 @@ export class ContentSearchService
...
+ *  searchDescription: “A description”,
+ *  size: -1,
+ *  subInfo: 0,
+ *  tags: ["DC GIS", "police", "speed", "speeding", "ticket", "traffic", "violation"],
+ *  thumbnail: "thumbnail/ago_downloaded.png",
+ *  title: "Traffic Camera Enforcement Locations in Washington DC",
+ *  type: "Web Map",
+ *  typeKeywords: ["ArcGIS Online", "Explorer Web Map", "Map", "Online Map", "Web Map"],
+ * }
+ * ```
```
andrewctate, line 223: For some reason I get syntax highlighting on the next snippet, but not this one. Not a huge deal.

andrewctate (commented):
Lookin good! Left a few comments

packages/search/src/content/index.ts
sonofflynn89: Good catch! I didn't even think about that!

packages/search/src/content/index.ts
sonofflynn89: This was the original description on the function, but I'm happy to change it!

packages/search/src/content/index.ts
sonofflynn89: Really? I thought the syntax highlighting worked on both for me but I'll double check

packages/search/src/content/index.ts
sonofflynn89: Yeah, I can see that. Do you think I should make more examples using this extended structure with boolean operators @andrewctate ?

packages/search/src/content/index.ts
drspacemanphd: Hate to be that guy, but including both `no` and `unavailable` seems like a double negative. How about `All public, as well as private content accessible to authenticated user, will be included`?

packages/search/src/content/index.ts
drspacemanphd: Unfortunately, this seems to be par for the course. There are a default list of attributes that are returned, but these are incomplete compared to Portal API. See [schema.json](https://github.com/ArcGIS/hub-indexer/blob/master/packages/api/schema.json) for how the API properties of the V3 API are mapped to the elasticsearch document properties. Those that appear to be identical attributes to their Portal API equivalents, but do not come from the `item` ES document property, may be different.

drspacemanphd (approved):
Approved with a minor change to request, as well as any unaddressed comments from Tate.

packages/search/src/content/index.ts
andrewctate: I know :)

packages/search/src/content/index.ts
andrewctate: Fair point. Not to be "that other guy" but what about

`Any private content that the authenticated user can access will be included in the results`

packages/search/src/content/index.ts
andrewctate: I think you've got a good example below. I'm just wondering if there's a reason to keep this comment?

packages/search/src/content/index.ts
sonofflynn89: Personally, I like that suggestion @andrewctate because it will flow a little better as an inline comment and it's scoped to just the private content

packages/search/src/content/index.ts
sonofflynn89: No, I just threw in there because someone could theoretically see this in the future once searchContent is used in the app, but you're right in that we have another example and someone who runs into this should be able to parse out the structure from these and other docs 👍

packages/search/src/content/index.ts
ajturner: Feel free to change any text that results in an easier-to-understand library

andrewctate (approved):
Nice work!

PR #619, opened 2021-08-30
B/layer datasets have wrong item type

tomwayson:
1. Description: add `getServiceTypeFromUrl()` to derive the item type based on URL b/c the `type` returned from the Hub API will be the layer's type for feature layers, tables, and raster layers. Fix `content.item.type` so that it has the correct type in these cases.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/util.ts
+++ b/packages/common/src/util.ts
@@ -395,3 +395,15 @@ export function chunkArray(arr: any[], size: number) {
  * @returns
  */
 export const isNil = (value: unknown) => value == null;
+
+/**
+ * Upper case first letter (only) of a string
+ * @param word
+ * @returns Word
+ */
+export const capitalize = (word: string) => {
+  // upper case first letter and return as element in array for backwards compatibility
+  const chars = Array.from(word);
+  chars[0] = chars[0].toUpperCase();
+  return chars.join("");
+};
--- a/packages/content/src/portal.ts
+++ b/packages/content/src/portal.ts
@@ -82,6 +82,7 @@ export function itemToContent(item: IItem): IHubContent {
     // we don't store item.name in the Hub API and we use name for title
     name: item.title,
     family: getFamily(normalizedType),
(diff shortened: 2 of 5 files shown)


tannerjt (approved):

packages/common/src/util.ts
```diff
@@ -395,3 +395,15 @@ export function chunkArray(arr: any[], size: number) {
...
 export const isNil = (value: unknown) => value == null;
+
+/**
+ * Upper case first letter (only) of a string
+ * @param word
+ * @returns Word
+ */
+export const capitalize = (word: string) => {
+  // upper case first letter and return as element in array for backwards compatibility
+  const chars = Array.from(word);
```
rweber-esri, line 406: This is perfectly fine, but fwiw `String` is already iterable, so we could achieve the same w/out the `Array.from`.

```js
const [head, ...tail] = word;
return [head.toUpperCase(), ...tail].join('');
```

packages/content/src/portal.ts
```diff
@@ -117,11 +118,16 @@ export function itemToContent(item: IItem): IHubContent {
...
 
 /**
+ * DEPRECATED: Use getFamily() instead.
+ *
  * get the HubType for a given item or item type
  *
  * @param itemOrType an item or item.type
  */
 export function getItemHubType(itemOrType: IItem | string): HubType {
+  console.warn(
```
rweber-esri, line 128: Any reason to not use `Logger.warn`?

rweber-esri (approved):
Looks good to me, only minor/optional comments.

rweber-esri:
> Looks good to me, only minor/optional comments.

Disregard, didn't see this got merged already.

packages/content/src/portal.ts
tomwayson: Like all the other deprecations, this is temporary code, so I think `console.log()` is fine, even better that `Logger` in cases where that would have to be added to the `import`s.

PR #625, opened 2021-09-07
chore(hub-teams): migrate a number of functions to hub-teams

benstoltz:
1. Description: This PR migrates a number of functions over to hub.js from ember land.

1. Instructions for testing:


packages/teams/src/remove-team-from-items.ts
```diff
@@ -0,0 +1,45 @@
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { IUpdateItemResponse, updateItem } from "@esri/arcgis-rest-portal";
+import { cloneObject, IModel, without } from "@esri/hub-common";
+import { removeTeam } from "./remove-team";
+
+/**
+ * Removes a Team from N hub models.
+ * First, if passed deleteTeam, it will delete the team
```
ajturner, line 8: Would this be better as a separate call rather than trying to add complexity to this method?

packages/teams/src/remove-team-from-items.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export async function removeTeamFromItems(
+  teamId: string,
+  models: IModel[],
+  deleteTeam: boolean,
+  authentication: UserSession
+): Promise<IUpdateItemResponse[]> {
+  if (deleteTeam) {
+    // first kill the team...
+    await removeTeam(teamId, authentication);
```
ajturner, line 26: Per comment about separate delete team - what is delete team fails?

In particular the user removing a team from items may not be a manager of that team. 
They do need to be an editor of the item - but that's all.

packages/teams/src/remove-team-from-items.ts
```diff
@@ -0,0 +1,45 @@
...
+    models.map((model) => {
+      // clone the item before modifying
+      const clonedModel = cloneObject(model);
+      // remove the id from the item.properties.teams array
+      clonedModel.item.properties.teams = without(
+        clonedModel.item.properties.teams,
+        teamId
+      );
+      // update the item
+      return updateItem({
```
ajturner, line 39: What happens if the user doesn't have permissions to update this item?

packages/teams/src/remove-team.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * removeTeam unprotects a group, then deletes it
+ *
+ * @export
+ * @param {string} id Team Id
+ * @param {UserSession} authentication authentication
+ * @return {*}  {Promise<any>}
+ */
+export async function removeTeam(
```
ajturner, line 12: This method should be called deleteTeam

Remove ~ unassociate (but not delete)

Delete ~ destroy from system

packages/teams/src/remove-team.ts
```diff
@@ -0,0 +1,23 @@
...
+  id: string,
+  authentication: UserSession
+): Promise<any> {
+  // unprotect the group
+  const unprotectResult = await unprotectGroup({ id, authentication });
```
ajturner, line 17: This seems dangerous - I suggest we reconsider automatically protecting / unprotecting groups and leave that as a choice to the group manager

packages/teams/src/remove-team.ts
```diff
@@ -0,0 +1,23 @@
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { removeGroup, unprotectGroup } from "@esri/arcgis-rest-portal";
+
+/**
+ * removeTeam unprotects a group, then deletes it
+ *
+ * @export
+ * @param {string} id Team Id
+ * @param {UserSession} authentication authentication
+ * @return {*}  {Promise<any>}
```
ajturner, line 10: Should methods return an explicit outcome and verification of what was deleted. That would be useful for multiple async calls

packages/teams/src/remove-users-from-team.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  IRemoveGroupUsersResult,
+  removeGroupUsers,
+} from "@esri/arcgis-rest-portal";
+import { IAuthenticationManager } from "@esri/arcgis-rest-request";
+
+/**
+ * Remove N users from team. Acts as a wrapper around removeGroupUsers from arcgis-rest-portal
```
ajturner, line 8: Generally I don't think we need wrappers if the direct call does the exact same thing.

ajturner (changes requested):
Change remove to delete when it's deleting resources. Only use remove when it's in associating resources (e.g. remove user from group, remove team from site)

@dbouwman we should determine if we need to wrap simple Group operations.

could we limit Hub.js to "group to Hub workflows" and then just rely on Groups for all standard operations?

The Hub.js doc / guide can make explicit the operations a Dev would use for standard group ops vs Hub-specific ops.

packages/teams/src/remove-team-from-items.ts
dbouwman: Although this may have been written as a single method in Ember, I agree with Andrew - keep this simpler, and remove the team via the `removeTeam` fn vs here

packages/teams/src/remove-team-from-items.ts
dbouwman: Assuming the Model's are fully populated, you should be checking the `itemControl` and it should be `admin` or `update` in order for the `update` call to succeed.

packages/teams/src/remove-users-from-team.ts
dbouwman: @ajturner  I'd thought we wanted Hub.js to be "complete" and to use Hub Terminology - thus `removeUsersFromTeam` vs expecting a user to know to import some other fn from REST-JS?

packages/teams/src/remove-team.ts
dbouwman: This is a long-standing intentional pattern in Hub to limit unintended item/group deletion in the Home App. Within the Hub UI we typically have a confirmation dialog, and then remove an Item (site etc) along with it's related items/groups. If we want to change this pattern, so it's all user-controller, that's fine, but we need to accommodate that in the UI before we just change these functions.

packages/teams/src/remove-team.ts
dbouwman: Let's handle the `remove` vs `destroy` vs `delete` in a wholistic manner -  for now let's leave this as `remove` since it's consistent w/ rest-js and the rest of hub.js

dbouwman (approved):

Issue #606, opened 2021-08-10
searchContent results should be an array of IHubContent


tomwayson:
Currently `results` is an array of the `attributes` returned for each record when querying the Hub API:

https://github.com/Esri/hub.js/blob/40a59953fe0504a34f81f4c7e0decda75e740c14/packages/search/src/content/helpers/convert-hub-response.ts#L138-L150

And when querying the portal API, it is just the results (`IITem[]`):

https://github.com/Esri/hub.js/blob/40a59953fe0504a34f81f4c7e0decda75e740c14/packages/search/src/content/helpers/convert-portal-response.ts#L27

I propose that we:
- for Hub API we map the records to `datasetToContent()`
- for for the Portal API we map the results to `itemToContent()`

This should not be much of a breaking change, especially in the case of Portal since `IHubContent` is a superset of `IItem`. In the case of Hub, `datasetToContent()` may be slightly different (mostly corrections), and may omit some enrichments.

I understand that `searchContent()` is in use for DCAT, if there are omissions on the Hub side that are needed by that, we can add them to `IHubContent` and `datasetToContent()`.

ajturner:
@tomwayson this is a good suggestion using the façade pattern. 

Since we're also looking forward to a "v4" search API would this pattern work where we might add something like an internal `searchv4ToContent()` method for transformations from that (ideally simplified) response into a Content object?

tomwayson:
Good idea @andrewctate 

I suggest we expose  these lines:

https://github.com/Esri/hub.js/blob/5816e0d0943a26a556717fe9cf21930e0029bbe9/packages/search/src/content/index.ts#L133-L156

as `searchDatasets(request: IContentSearchRequest): { data: DatasetResource[] }`

So it still takes a search request, but returns essentially the raw JSONAPI response from the API.

As you propose `searchContent()` would  call that internally, but then return `response.data.map(datasetToContent)`

ajturner:
> The fate of the current editor is uncertain, but in the short term we may well simply update those paths to match the `attributes` object in the v3 API response. And, again, customers will be able to see exactly what these paths refer to by inspecting the naked responses from the v3 API.

This is an intriguing concept that the API provides the documentation for the Editor. The editor can evolve to in fact be drive by the API through introspection of the response object model or the self-describing JSON-Schema of the API which would provide additional element information such as description, source, requirements, etc. 

[...]

PR #626, opened 2021-09-07
fix(hub-common): hupApiUrl, isPortal, and authentication options shou…

tomwayson:
…ld be optional

`authentication` is only required to be a `UserSession` for certain requests, so those use a new `IHubUserRequestOptions` - analogous to `IUserRequestOptions` instead of having `IHubRequestOptions` extend the latter.

`isPortal` is not needed by any fns that request directly to the API, and since it is a `boolean` no changes were needed to support making it optional.

In addition to making `hubApiUrl` optional, this PR ensures that any place `hubApiUrl` was referenced that we supplied a default.

This puts in place some things that will make it easier to do #606 

affects: @esri/hub-common, @esri/hub-discussions, @esri/hub-sites

1. Description:

1. Instructions for testing:

### DIFF
--- a/packages/common/src/urls/get-hub-api-url-from-portal.ts
+++ b/packages/common/src/urls/get-hub-api-url-from-portal.ts
@@ -6,5 +6,5 @@ import { getHubUrlFromPortal } from "./get-hub-url-from-portal";
  * @param {Object} portal Portal Self
  */
 export function getHubApiUrlFromPortal(portal: IPortal): string {
-  return `${getHubUrlFromPortal(portal)}/api/v3`;
+  return `${getHubUrlFromPortal(portal)}`;
 }
--- a/packages/common/src/urls/get-hub-api-url-from-portal.ts
+++ b/packages/common/src/urls/get-hub-api-url-from-portal.ts
@@ -6,5 +6,5 @@ import { getHubUrlFromPortal } from "./get-hub-url-from-portal";
  * @param {Object} portal Portal Self
  */
 export function getHubApiUrlFromPortal(portal: IPortal): string {
-  return `${getHubUrlFromPortal(portal)}`;
+  return `${getHubUrlFromPortal(portal)}/api/v3`;
 }
--- a/packages/discussions/src/utils/request.ts
+++ b/packages/discussions/src/utils/request.ts
@@ -1,5 +1,8 @@
 import { IHubRequestOptions } from "../types";
-import { RemoteServerError as _RemoteServerError } from "@esri/hub-common";
+import {
+  RemoteServerError as _RemoteServerError,
+  buildUrl,
+} from "@esri/hub-common";
 
 export class RemoteServerError extends _RemoteServerError {
   error: string;
@@ -62,13 +65,15 @@ export function apiRequest<T>(
     method: options.httpMethod || "GET",
     mode: options.mode,
     cache: options.cache,
-    credentials: options.credentials
+    credentials: options.credentials,
   };
 
-  // NOTE: this should default to the prod url once deployed and microservice root URLs
-  // are normalized: https://github.com/Esri/hub.js/pull/479#discussion_r607866561
-  const apiBase =
-    options.hubApiUrl || "https://hub.arcgis.com/api/discussions/v1";
+  const apiBase = buildUrl({
+    // TODO: we _want_ to use getHubApiUrl(),
+    // but have to deal w/ the fact that this package overwrites IHubRequestOptions
+    host: options.hubApiUrl || "https://hub.arcgis.com",
+    path: "/api/discussions/v1",
+  });
 
   if (options.params) {
     if (options.httpMethod === "GET") {
@@ -80,12 +85,12 @@ export function apiRequest<T>(
   }
 
   const url = [apiBase.replace(/\/$/, ""), route.replace(/^\//, "")].join("/");
-  return fetch(url, opts).then(res => {
+  return fetch(url, opts).then((res) => {
     if (res.ok) {
       return res.json();
     } else {
       const { statusText, status } = res;
-      return res.json().then(err => {
+      return res.json().then((err) => {
         throw new RemoteServerError(
           statusText,
           url,
--- a/packages/sites/src/_remove-site-from-index.ts
+++ b/packages/sites/src/_remove-site-from-index.ts
@@ -1,4 +1,4 @@
-import { IModel, IHubRequestOptions } from "@esri/hub-common";
+import { IModel, IHubRequestOptions, getHubApiUrl } from "@esri/hub-common";
 
 /**
  * Remove a Site from the Hub Index system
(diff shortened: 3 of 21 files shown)


packages/sites/src/_remove-site-from-index.ts
```diff
@@ -13,21 +13,21 @@ export function _removeSiteFromIndex(
   if (hubRequestOptions.isPortal) {
     return Promise.resolve();
   } else {
-    const url = `${hubRequestOptions.hubApiUrl}/v2/${siteModel.item.id}`;
+    const url = `${getHubApiUrl(hubRequestOptions)}/v2/${siteModel.item.id}`;
```
tomwayson, line 16: @andrewctate looks like you were the last to edit this, it should probably be `/v3/`, right?

packages/discussions/src/utils/request.ts
```diff
@@ -62,13 +65,15 @@ export function apiRequest<T>(
...
+    credentials: options.credentials,
   };
 
-  // NOTE: this should default to the prod url once deployed and microservice root URLs
-  // are normalized: https://github.com/Esri/hub.js/pull/479#discussion_r607866561
-  const apiBase =
-    options.hubApiUrl || "https://hub.arcgis.com/api/discussions/v1";
+  const apiBase = buildUrl({
+    // TODO: we _want_ to use getHubApiUrl(),
+    // but have to deal w/ the fact that this package overwrites IHubRequestOptions
```
dbouwman, line 73: O_o WAT? Details?

packages/sites/src/create-site-model-from-template.ts
```diff
@@ -114,12 +115,9 @@ export function createSiteModelFromTemplate(
         );
         settings.solution.url = getPortalSiteUrl(uniqueSubdomain, portal);
       } else {
-        settings.solution.defaultHostname = `${uniqueSubdomain}-${
-          portal.urlKey
-        }.${stripProtocol(hubRequestOptions.hubApiUrl)}`;
-        settings.solution.url = `https://${uniqueSubdomain}-${
-          portal.urlKey
-        }.${stripProtocol(hubRequestOptions.hubApiUrl)}`;
+        const base = stripProtocol(getHubApiUrl(hubRequestOptions));
```
dbouwman, line 118: Seems I can't add a comment in the un-changed code - but this fn takes an `IHubRequestOptions` which now has `authorization` as optional... but this fn needs `authorization`...  I guess this is a way to make a non-breaking change, but it also feels like it opens up possible confusion... could we make it `IHubRequestOptions | IHubUserRequestOptions` so we're widening the types, and then at a major we can remove the first one?

packages/sites/src/drafts/save-published-status.ts
```diff
@@ -10,11 +10,11 @@ import { hasUnpublishedChanges } from "./has-unpublished-changes";
  * leaving everything else on the model alone.
  *
  * @param siteOrPageModel
- * @param hubRequestOptions
+ * @param requestOptions
  */
 export function savePublishedStatus(
   siteOrPageModel: IModel,
-  hubRequestOptions: IHubRequestOptions
+  requestOptions: IHubUserRequestOptions
```
dbouwman, line 17: can we do this in `createSiteModelFromTemplate`?

dbouwman (approved):
Really just one question re: createModelFromTemplate - otherwise /me bows

packages/discussions/src/utils/request.ts
tomwayson: https://github.com/Esri/hub.js/pull/479/files/14dfb07efe2e1ff99bb38b203aaf0af237b62564#r595624082

packages/sites/src/create-site-model-from-template.ts
tomwayson: Good catch.

I thought since those clients were already passing in a `IHubRequestOptions` _with_ `authentication` as a `UserSession` that TS would be able to figure out that's the same as a `IHubUserRequstOptions`, but alas, no. I just tried updating the signature of that fn to take only a `IHubUserRequestOptions` and I got a bunch of compile errors when running the tests about not being able to convert `IHubRequestOptions` to `IHubUserRequestOptions`. 

I then tried your suggestion of `hubRequestOptions: IHubUserRequestOptions | IHubRequestOptions` and tests compiled and passed. So we could do that, but I'd have to figure out all the places that require `authentication` as a `UserSession`. I think that's better than backing out the `authentication` related changes, b/c ultimately we'll want to have `authentication` optional for the majority of requests, so we'll have to do this work at some point.

packages/sites/src/_remove-site-from-index.ts
andrewctate: Yeah, looks like I missed bumping that API version. Thanks!

andrewctate (approved):

packages/common/src/urls/get-hub-api-url-from-portal.ts
```diff
@@ -6,5 +6,5 @@ import { getHubUrlFromPortal } from "./get-hub-url-from-portal";
  * @param {Object} portal Portal Self
  */
 export function getHubApiUrlFromPortal(portal: IPortal): string {
-  return `${getHubUrlFromPortal(portal)}/api/v3`;
+  return `${getHubUrlFromPortal(portal)}`;
```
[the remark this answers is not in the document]
tomwayson: good catch.

so, in looking into this PR I realized that the hub app _always_ passes in `hubApiUrl` w/o `/api/v3` like `https//hub.arcgis.com`, which was surprising to me. I would expect that to be called something like `hubRootUrl`, or `hubBaseUrl`.

It also appeared that most places w/in hub.js appended `/api/v3`, so I figured `hubApiUrl` should not have `/api/v3` on it, and I figured this fn was supposed to derive a `hubApiUrl` from an `IPortal`, so it should not append `/api/v3`.

Turns out we don't use this function anywhere in the app nor hub.js outside of the test, which, I have no idea how it's passing b/c [the only thing expects is for the url to contain `/api/v3`](https://github.com/Esri/hub.js/blob/33b30d656413b0a4eda6c5af686761c52724483b/packages/common/test/urls/get-hub-api-url-from-portal.test.ts#L14)????

Given all that, I think I'll revert this change, but I don't think it's great that everywhere we expect `hubApiUrl` to not have `/api/v3`, but then we have a function called `getHubApiUrlFromPortal()` that returns `/api/v3`.

packages/sites/src/create-site-model-from-template.ts
tomwayson: After our chat this AM, I went through Solution.js and learned that it uses these fns from `@esr/hub-sites`:

```
createPageModelFromTemplate,
createPage,
removePage,
updatePage,
_updatePages,

createSiteModelFromTemplate,
createSite,
getSiteById,
removeSite,
convertSiteToTemplate
_shareItemsToSiteGroups,
updateSite
```

I suspect most of those other than `getSiteById()` would need the `IHubUserRequestOptions | IHubRequestOptions` hack in order to prevent TS compile errors in that package if we want to move forward w/ this PR.

I suspect that template.js imports the same fns, but haven't confirmed.

packages/sites/src/drafts/save-published-status.ts
tomwayson: note to self, this also probably needs the `IHubUserRequestOptions | IHubRequestOptions` hack

tomwayson:
I merged this w/o the `IHubUserRequestOptions | IHubRequestOptions` hack, b/c I realized that if a consuming app had an older version of a Hub.js package, that `IHubRequestOptions` would be equivalent to the new `IHubUserRequestOptions`

PR #629, opened 2021-09-12
c/add-link-underline-capability-default

MarvinPerry:
* feat(hub-sites): new sites in any env get underlinedLinks capability by default

affects: @esri/hub-sites

sites created from embedded templates in non-portal environments will have "underlinedLinks" included
by default

1. Instructions for testing:

### DIFF
--- a/packages/sites/src/ensure-required-site-properties.ts
+++ b/packages/sites/src/ensure-required-site-properties.ts
@@ -65,8 +65,8 @@ export function ensureRequiredSiteProperties(
     "document_iframes",
     "items_view",
     "app_page",
-    "globalNav",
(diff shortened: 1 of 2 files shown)


tomwayson (approved):
This LGTM, as long as this is only being required for newly generated sites. `ensureRequiredSiteProperties()` is called on existing items that could have other implications.

MarvinPerry:
> This LGTM, as long as this is only being required for newly generated sites. `ensureRequiredSiteProperties()` is called on existing items that could have other implications.

We do only want it on for newly created sites, while sites that were created before this change will have to opt into it by toggling the capability on. @tomwayson

PR #632, opened 2021-09-15
a deterministic, reliable spatial reference serialization strategy

andrewctate:
1. Description:

When an item export is completed, we tag the exported item with typeKeywords that make it possible to query for it based on item ID, layer ID, and spatial reference.

The spatial reference typeKeyword was being incorrectly applied since `spatialRefId` can be an ID _or_ a serialized `ISpatialReference` object. In the latter case, the object is split into two typeKeywords instead of one since it can contain a comma which the Portal API interprets as a delineator between the two typeKeywords. Here is an example (look at the first and last typeKeywords):

Screen Shot 2021-09-14 at 8 51 36 PM

That makes it difficult to reliably query for the exported items by spatial reference.

This PR
- introduces a deterministic, reliable spatial reference serialization strategy for the `spatialRefId` typeKeyword
- centralizes and deduplicates tagging logic for reuse

2. Instructions for testing:
`npm t`

3. Closes Issues: None


packages/common/src/downloads/build-existing-exports-portal-query.ts
```diff
@@ -41,12 +43,57 @@ export const PORTAL_EXPORT_TYPES = {
...
+export function serializeSpatialReference(
+  spatialReference: ISpatialReference | string
+): string {
+  if (typeof spatialReference === "object") {
+    const { wkid, wkt } = spatialReference;
+    return wkid ? wkid + "" : btoa(wkt);
+  } else {
+    return spatialReference;
+  }
+}
```
andrewctate, line 75: Here's the magical serialization logic. We know that a spatialReference always contains either `wkid` (no commas) or `wkt` (possibly commas). When there are possibly commas, we base64-encode.

packages/common/src/downloads/build-existing-exports-portal-query.ts
```diff
@@ -143,11 +201,37 @@ function buildExportTypesClause(
...
+ * Returns the keyword identifying exports by the layer they originate from
+ * @param layerId
+ * @returns
+ */
+export function getExportLayerTypeKeyword(layerId?: number | string) {
+  // NOTE - Layer Id's need to be padded with "0" so that /search results are predictable. Searches for typeKeywords:"exportLayer:1" don't work.
+  // See https://github.com/Esri/hub.js/pull/472 for more information.
+  // TODO - use `filter` when Enterprise Sites adds support.
+  return layerId ? `exportLayer:0${layerId}` : `exportLayer:null`;
 }
```
andrewctate, line 237: These are the functions that generate appropriate typeKeywords for the exported items. Now it's one set of logic that is used in both tagging and querying.

packages/common/src/downloads/build-existing-exports-portal-query.ts
```diff
@@ -41,12 +43,57 @@ export const PORTAL_EXPORT_TYPES = {
...
+    return spatialReference;
+  }
+}
+
+/**
+ * spatialRefId can currently take the form of either a WKID string or a
+ * serialized ISpatialReference object.
+ *
+ * TODO - we shouldn't need this function. Instead, spatialRefId should
+ * always be consistent, maybe by using serializeSpatialReference
```
andrewctate, line 82: Looking toward the future

rgwozdz (approved):

PR #633, opened 2021-09-16
fix(hub-content): introduces empty array to handle servers without layers mistakenly typed as feature

drspacemanphd:
affects: @esri/hub-content

1. Description:
A bug in the content-upload component of opendata-ui mistakenly characterizes non-feature services as Feature Services, which has downstream effects. In this case, it breaks an assumption that anything characterized as a Feature Service will automatically have layers as part of assembling enrichments. Because these items are now in the open, this change adds a defense mechanism to [prevent 404's from the search page](https://devtopia.esri.com/dc/hub/issues/1908) when clicking on such an item

2. Instructions for testing:
See [issue](https://devtopia.esri.com/dc/hub/issues/1908) for detailed repro. This also happens for **private** ArcGIS Online items. Once released an in UI, 404's should stop happening. **Note that this PR does not address underlying issue in the content upload component**.

3. Closes Issues: #<number> (if appropriate)

4. [x] ran commit script (`npm run c`)

### DIFF
--- a/packages/content/src/enrichments.ts
+++ b/packages/content/src/enrichments.ts
@@ -36,7 +36,7 @@ const getLayer = (content: IHubContent, layerId?: number) => {
       isFeatureService(content.type) && getOnlyQueryLayer(layers);
 };
 
-const getOnlyQueryLayer = (layers: ILayerDefinition[]) => {
+const getOnlyQueryLayer = (layers: ILayerDefinition[] = []) => {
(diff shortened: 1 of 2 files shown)


rweber-esri (approved):
This looks good to me

tomwayson (approved):
Thanks @drspacemanphd, I'll merge this after #635 and include it in that release.

PR #635, opened 2021-09-16
refactor(hub-search): use a new base-64 encoding lib

andrewctate:
affects: @esri/hub-common, @esri/hub-search

1. Description:
`hub-search` was using `base-64` but we're standardizing on `abab`.

1. Instructions for testing:
`npm t`

1. Closes Issues:

### DIFF
--- a/packages/search/src/content/helpers/convert-request-to-portal-params.ts
+++ b/packages/search/src/content/helpers/convert-request-to-portal-params.ts
@@ -1,5 +1,4 @@
-import { decode } from "base-64";
-import { Logger } from "@esri/hub-common";
+import { Logger, atob as decode } from "@esri/hub-common";
 import { IPagingParams, ISearchOptions } from "@esri/arcgis-rest-portal";
 import { UserSession } from "@esri/arcgis-rest-auth";
 import {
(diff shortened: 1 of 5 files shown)


packages/search/src/content/helpers/convert-request-to-portal-params.ts
```diff
@@ -181,6 +181,9 @@ function stringifyFilterValue(
 function decodePage(page: string): IPagingParams {
   try {
     const decodedPage: any = decode(page);
+    if (decodedPage === null) {
+      throw new Error();
```
tomwayson, line 185: Is this for parity w/ `base-64`'s behavior? Should we have a message here. I'm fine to add a message later.

packages/search/test/content/helpers/convert-request-to-portal-params.test.ts
```diff
@@ -170,7 +171,7 @@ describe("Convert Portal Params Function", () => {
     expect(portalParams.bbox).toBeUndefined();
   });
 
-  it("can handle an error occuring upon decoding a page", () => {
+  fit("can handle an error occuring upon decoding a page", () => {
```
tomwayson: I think this needs to be set back to `it()`

tomwayson (changes requested):
Thanks for doing this @andrewctate.

One minor test change.

packages/search/src/content/helpers/convert-request-to-portal-params.ts
andrewctate: Yes, it is. I'm just throwing to get down into the catch block where the message actually gets logged

andrewctate: alternative would be
```js
if (!decodedPage) {
    return undefined;
}
```

tomwayson (approved):
thanks @andrewctate!

PR #637, opened 2021-09-17
Fix family and normalizedType not updated for layers

### LINKED ISSUE #636: getLayerContent() should update family and normalizedType
`getLayerContent()` updates the `type` but that should also update the `family` `normalizedType`

tomwayson:
Resolves #636

Also, moving `itemToContent()` and `hubToContent()` to `@esri/hub-common` is needed for future work in support of #606 

affects:
@esri/hub-common
@esri/hub-content

1. Description:

1. Instructions for testing:


packages/common/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "abab": "^2.0.5",
     "adlib": "^3.0.7",
+    "jsonapi-typescript": "^0.1.3",
```
andrewctate, line 19: This should be a dev dep, right?

packages/content/src/enrichments.ts
```diff
@@ -596,7 +598,10 @@ export const getLayerContent = (
     return;
   }
   const { type, name, description } = layer;
-  const layerContent = { ...content, layer, type };
+  // get family and normalized type based on layer type
+  const family = getFamily(type);
+  const normalizedType = normalizeItemType({ ...content.item, type });
+  const layerContent = { ...content, layer, type, family, normalizedType };
```
tomwayson, line 604: I'm pretty sure this is the only change in this whole diff that's not just shufflin' from -content to -common

andrewctate (approved):
Looks good! Just a dependency that should be in dev dependencies.

sonofflynn89 (approved):

packages/common/package.json
tomwayson: good question. I'm not sure since we export a type that is based on a `jssonapi-typescript` type. I've been in these debates before, and essentially landed at - it doesn't hurt to have it as a dependency.

PR #647, opened 2021-09-23
F/get content from portal takes hubId

tomwayson:
1. Description:

Previously if you passed something like `<itemId>_<layerId>` to `getContent()` in either of the following scenarios you would get the following undesirable outcomes:
- in enterprise `getContent()` would only fetch the item and not the layer info, and return the item's content
- in online when the item is not public `getContent()` would throw a not found error

Now, in either scenario, `getContent()` will attempt to fetch the item from the portal _and_ then the layer info from the item's service and return content for the item, or the layer accordingly.

This should make the Hub application more tolerant of invalid routes and result in less 404s.

1. Instructions for testing:


dbouwman (approved):
Good work! Thanks Tom!

PR #643, opened 2021-09-23
feat(hub-common): add fetchDatasets() function

tomwayson:
affects: @esri/hub-common

1. Description: add fetchDatasets() fn

1. Instructions for testing:

1. another step towards #606 

1. [x] ran commit script (`npm run c`)

_Note_ If you don't run the commit script at least once, the Semantic Pull Request check will fail. Save yourself some time, and run `npm run c` and follow the prompts.

For more information see the README

PR #649, opened 2021-09-24
Revert "feat(hub-common): add fetchDatasets() function"

tomwayson:
Reverts Esri/hub.js#643

the method I added was called `fetchDatasets()`, but it actually called the `/search` end point, which is different than `/datasets`.

I'm going to add a new PR that adds `searchHub()` instead.


drspacemanphd (approved):
Thank you for the clarity

sonofflynn89 (approved):

PR #650, opened 2021-09-24
feat(hub-common): add searchHub() to exectue searches against the Hub…

tomwayson:
1. Description:

Adds public `searchHub()` fn to call Hub API's `/search` end point.

Updates `searchContent()` to use `searchHub()` internally and to expose a private (undocumented)`_searchHubByContentSearchRequest()` fn that addresses https://github.com/Esri/hub.js/issues/606#issuecomment-912107464

1. Instructions for testing:

### DIFF
--- a/packages/search/src/content/index.ts
+++ b/packages/search/src/content/index.ts
@@ -2,11 +2,12 @@ import { UserSession } from "@esri/arcgis-rest-auth";
 import {
   getHubApiUrl,
   getProp,
-  hubApiRequest,
+  searchHub,
   IHubRequestOptions,
   fetchSite,
   ISiteCatalog,
   getPortalApiUrl,
+  DatasetResource,
 } from "@esri/hub-common";
 import { IItem, ISearchOptions, ISearchResult } from "@esri/arcgis-rest-portal";
 import { searchItems } from "@esri/arcgis-rest-portal";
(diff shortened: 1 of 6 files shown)


packages/search/src/content/index.ts
```diff
@@ -322,30 +326,37 @@ function performEnterpriseContentSearch(
...
+ *
+ * NOTE: this returns the Hub API's raw JSONAPI response
+ * and invalid parameters like isPortal will be ignored
+ *
+ * @param request - see searchContent()
+ * @returns Hub API's JSONAPI response
+ *
+ * @private
+ */
+export function _searchHubByContentSearchRequest(
```
tomwayson, line 341: I'm definitely open to renaming this if someone has a better idea.

drspacemanphd: Is `_searchHubContent` no good?

drspacemanphd (approved):
Thanks for doing this.

packages/search/src/content/index.ts
tomwayson: Thanks for the suggestion @drspacemanphd. I thought about `searchHubContent()`, but it doesn't return content. @andrewctate suggested `searchHubDatasets()`, which is kind of inline w/ what I was originally thinking of calling this: `searchDatasets()`.

He also suggested `searchHubApi()`, which to me is more appropriate for the `searchHub()` fn I added in this PR. Since that fn is a thin wrapper over `hubApiRequest()` I'm thinking of renaming it to `hubApiSearch()`, that way we can reserve the `search` prefix for fns that that `IHubSearchRequestOptions`, which brings me back around to naming this `searchDatasets()`.


I think I'm going to go w/

PR #652, opened 2021-09-24
fix(hub-search): searchContent() should return results as IHubContent[]

### LINKED ISSUE #606: searchContent results should be an array of IHubContent

tomwayson:
affects: @esri/hub-search

1. Description:

1. Instructions for testing:

1. Resolves #606

### DIFF
--- a/packages/search/src/content/helpers/convert-hub-response.ts
+++ b/packages/search/src/content/helpers/convert-hub-response.ts
@@ -1,5 +1,5 @@
 import { UserSession } from "@esri/arcgis-rest-auth";
-import { getProp } from "@esri/hub-common";
+import { datasetToContent, getProp } from "@esri/hub-common";
 import { ISearchParams } from "../../ago/params";
 import {
   IContentAggregations,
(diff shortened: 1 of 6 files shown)


drspacemanphd (approved):
Excellent!

andrewctate (approved):
Nice work!

Issue #658, opened 2021-10-06
searchContent() should return content "path" and portal URLs


tomwayson:
We should call `withContentUrls()` on `IHubContent` results to provide fully qualified thumbnail URLs, etc.

We should also move the logic from the hub app that determines the content's route (primarily based on family) to Hub.js and have that that return a path (i.e. for `family: 'datasets'` it would be `/datasets/dataset`). TBD:
- should this include the leading `/`?
- should it include the trailing identifier `/slugOrId`?
- how can we use this property in the Hub app instead of the current logic that returns a route name (i.e. `datasets.dataset`)?

PR #667, opened 2021-10-13
Add relative and absolute site URLs to content

tomwayson:

1. Description:

This is a WIP, still needs tests.

First step in addressing #658 by adding `content.urls.relative` which will be in the form of `/:content-views-route/:slug-or-id`

Includes the (commented out) code for adding `content.urls.site` which would be the fully qualified URL and has the additional logic for determining if a page is in a site. That will get wrapped in a new fn called `setContentSiteUrl(content: IHubContent, site: IModel): IHubContent` that will return a new content w/ the updated `urls` hash.

@dbouwman, before I write the tests and move forward w/ that plan^^^ , I wanted to validate w/ you:
1. the `.urls` approach, I'm planning to move the others here too, example: `content.itemThumbnailUrl` -> `content.urls.thumbnail`
2. the `setContentSomeProperty(content, updatedValue)` pattern (as an alternative to a HubContent class w/ computed getters). I'll call out a few pain points in inline comments

1. Instructions for testing:


tomwayson:
oh yeah, similarly, `itemToContent()` and `datasetToContent()` now return their best guess at `content.identifier`.

Issue #655, opened 2021-09-30
remove any peerDependencies on packages other than hub-common


tomwayson:
Fortunately it looks like it's just sites that has `peerDependences` on initiatives and teams.

That causes:
- us to have to keep those `peerDependencies` up to date manually at release time
- solution.js to depend on teams and initiatives even though it doesn't use them,
- solution.js maintainers to rely on often ignored console messages to know if they need to update teams or initiatives

Proposed solution:
- [ ] move `getInitiativeTemplate()`, `addInitiative()`, `updateInitiativeSiteId()` (and any fns they need) to common
- [ ] consolidate teams into common since it is relatively small and we'll want to do this for most if not all packages

tomwayson:
NOTE: the above is true of teams in template-js too.

Furthermore, the processors package only depends on initiatives for `getUniqueGroupName()`, so maybe that can move to common too at this juncture and then template-js can drop the dependencies on both teams and initiatives.

PR #669, opened 2021-10-14
fix(hub-sites): prevent surveys and their feature services from being…

rweber-esri:
… shared to site teams during si

affects: @esri/hub-sites, @esri/hub-surveys

ISSUES CLOSED: [2325](https://devtopia.esri.com/dc/hub/issues/2325)

1. Description:

1. Instructions for testing:

1. Closes Issues: #[2325](https://devtopia.esri.com/dc/hub/issues/2325)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,35 +5,35 @@
 	"requires": true,
 	"dependencies": {
 		"@babel/code-frame": {
-			"version": "7.15.8",
-			"resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.15.8.tgz",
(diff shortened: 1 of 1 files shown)


rweber-esri:
> Nice work @rweber-esri, but I would like to keep from introducing another peer dep, which would require a breaking change and add to the work in #655

@tomwayson I _believe_ I've addressed your feedback. I:

* Added a `surveys` dir in `hub-common`
* Moved the method/file/tests/fixtures (and their dependencies) that need to be consumed from `hub-sites` to the new dir in `hub-common`
* Renamed the fixtures that moved to `.json` vs `.ts` to be more consistent w/ other fixtures in `hub-common`, updated all refs to these fixtures to account for the change
* Re-exported things that moved to `hub-common` from `hub-surveys` to prevent breakage.

PR #675, opened 2021-10-29
F/2332 discussions updates

rweber-esri:
1. Description:
* Refactors `parseDiscussionURI` to remove `url-parse` dependency
* Adds `isDiscussable` polymorphic method
* Refactors `isItemDiscussable` and `isGroupDiscussable` to call new `isDiscussable` method and flags them for deprecation in next major release.
* Adds `CONTENT = 'content'` to `DiscussionType` enum
* Adds `getHubRequestOptions` method

1. Instructions for testing:

1. Closes Issues: [2332](https://devtopia.esri.com/dc/hub/issues/2332)


packages/discussions/src/types.ts
```diff
@@ -80,7 +80,8 @@ export enum PostStatus {
 export enum DiscussionType {
   DATASET = "dataset",
   ITEM = "item",
-  GROUP = "group"
+  GROUP = "group",
+  CONTENT = "content",
```
rweber-esri, line 84: This enum is what is causing the [Discussions API](https://devtopia.esri.com/dc/hub-discussions/blob/main/src/posts/dto/validators/is-discussion-uri.ts#L19) to serve an error when we try to create a new post w/ a `content` URI, e.g. `hub://content/12345`. I question if we _really_ want to restrict this; we'll need to update this ENUM and release any time we want a new discussion URI pattern. The alternative would be to refactor the API to completely remove this restriction, then we can deprecate this ENUM.

packages/common/src/urls/get-portal-url.ts
```diff
@@ -28,7 +25,11 @@ import { IHubRequestOptions } from "../types";
  * @returns The portal base URL, defaults to `https://www.arcgis.com`
  */
 export function getPortalUrl(
-  urlOrObject?: string | IPortal | IHubRequestOptions | IRequestOptions
+  urlOrObject?:
+    | string
+    | IHubRequestOptionsPortalSelf
```
rweber-esri, line 30: @tomwayson 

`IHubRequestOptionsPortalSelf` extends `IPortal` and only adds an _optional_ `user` property, therefore this change from `string | IPortal | IHubRequestOptions | IRequestOptions` to `string | IHubRequestOptionsPortalSelf | IHubRequestOptions | IRequestOptions` should be transparent to our users and a non-breaking change, but please check me on this. I'm also happy to change this to `string | IPortal | IHubRequestOptionsPortalSelf | IHubRequestOptions | IRequestOptions`, too, if you think that makes more sense.

packages/common/src/urls/get-portal-api-url.ts
```diff
@@ -26,7 +25,11 @@ import { getPortalUrl } from "./get-portal-url";
  * @returns The portal API URL, defaults to `https://www.arcgis.com/sharing/rest`
  */
 export function getPortalApiUrl(
-  urlOrObject?: string | IPortal | IHubRequestOptions | IRequestOptions
+  urlOrObject?:
+    | string
+    | IHubRequestOptionsPortalSelf
```
rweber-esri, line 30: Similar, `IHubRequestOptionsPortalSelf` extends `IPortal` and only adds an _optional_ `user` property, so this should be non-breaking.

packages/common/src/content.ts
```diff
@@ -66,7 +66,7 @@ const cache: { [key: string]: string } = {};
 export function getCategory(itemType: string = ""): string {
   /* tslint:disable no-console */
   console.warn(
-    "DEPRECATED: Use getCollection() instead. getCategory will be removed at v9.0.0"
+    "DEPRECATED: Use getCollection() instead. getCategory will be removed at v10.0.0"
```
rweber-esri, line 69: Simply updating log statements for things we flagged for removal in v9.0.0 but didn't get the axe when that release was published.

packages/common/src/utils/get-hub-request-options.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Apache-2.0 */
+
+import { UserSession } from "@esri/arcgis-rest-auth";
+import { getSelf } from "@esri/arcgis-rest-portal";
+import { getHubApiUrlFromPortal } from "../urls/get-hub-api-url-from-portal";
+import { getPortalApiUrl } from "../urls/get-portal-api-url";
+import { IRequestOptions } from "@esri/arcgis-rest-request";
+import { IHubRequestOptionsPortalSelf } from "../types";
+
+export interface IGetHubRequestOptionsOptions extends IRequestOptions {
```
rweber-esri, line 11: Not a fan of the `...OptionsOptions` interface name, but felt `...OptionsParams` was equally ugly, LMK if you have any strong feelings about this interface name.

packages/common/src/utils/get-hub-request-options.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * A utility method to create an IHubRequestOptions object
+ * @param {Object} [options={}] An optional map of options
+ * @param {UserSession} [options.authentication] An optional UserSession object
+ * @param {string} [options.portal] An optional portal URL
+ * @param {token} [options.token] An optional token
+ * @param {portalSelf} [options.portalSelf] An optional IHubRequestOptionsPortalSelf object
+ * @returns {Promise}
+ */
+export function getHubRequestOptions(
```
rweber-esri, line 25: From [the authentication docs](https://github.com/Esri/hub.js/blob/master/docs/src/guides/authentication.md),

> Note In the future Hub.js will expose a function to streamline construction of IHubRequestOptions

This method takes a stab at that so we can create an IHubRequestOptions from:
1. `token` & `portal` URL combo
1. existing `UserSession` object
1. `token` & `portalSelf` object combo

since we'll need something like this in `hub-components` since we're likely going to be passing tokens around.

packages/discussions/src/types.ts
ajturner: I agree with the concern about having to regularly update this enumeration for validation. 

Instead, what about using this enumeration just for string operations (e.g. devs only use the enumeration to create the URI) to prevent mistypings, but _not_ for actual validation. Similar pattern to the Telemetry Dictionaries as a method to prevent event string name mistypings. 

e.g. ``let subject = `${DiscussionSource.HUB}://${DiscussionType.CONTENT}/${item.id}` ``

Which sets the stage for Post.subject being an object

```ts
let subject = {
  source: DiscussionSource.HUB,
  type: DiscussionType.CONTENT,
  object: item.id
  ...
}
```

packages/common/src/utils/get-hub-request-options.ts
dbouwman: @rweber-esri  we should not be working directly with tokens. Our components should take a `session` object when appropriate, and/or a IHubRequestOptions. We will be building an identity component for use outside the Hub application, and that will expose a `session` property as well as raise authentication events.

packages/common/src/urls/get-portal-url.ts
dbouwman: I'm not 100% sure that will be non-breaking b/c the interface name changed... if you just add an optional prop to an interface, that's non-breaking... this will be fine for JS clients, but I _suspect_ typescript libs will have issues (i.e. Solution.js)

packages/common/src/urls/get-portal-api-url.ts
dbouwman: Since IPortal has `[key: string]: any;`, unless there is a _strong_ need for `.user` to be typed, I think we can skip this and just check for `.user` (which you'd have to to anyhow), and then cast it via `as UserSession`

packages/common/src/utils/get-hub-request-options.ts
rweber-esri: > @rweber-esri we should not be working directly with tokens. Our components should take a `session` object when appropriate, and/or a IHubRequestOptions. We will be building an identity component for use outside the Hub application, and that will expose a `session` property as well as raise authentication events.

@dbouwman
Thanks for the context, I suspected things might change in this department. Would you then say:
1. We do not need this method
2. This method is fine the way it is
3. Refactor this method to accept only a UserSession and optionally an IPortal or IHubRequestOptionsPortalSelf

More of `hub-components` questions, but:
* Do we need to support both `UserSession` and `IHubRequestOptions`? If we pass a `UserSession`, then the component will be responsible for getting an `IHubRequestOptions` for the provided `UserSession`, and we'd likely need to do that in _many_ components. I would argue we should _only_ support passing `IHubRequestOptions` to simplify things unless I'm overlooking (quite possible) a burning need to support both?
* Has there been any discussion surrounding how we plan to support padding `UserSession` or `IHubRequestOptions` from Storybook? We can't pass instances from controls/knobs, out of the box anyways. Maybe a custom decorator or control(s) is in order?

I'm really glad you raised ^, I'm going to update the discussions components _now_ to accept IHubRequestOptions as part of [PR-154](https://github.com/Esri/hub-components/pull/154), as I'm about to add/gate those components in opendata-ui and would prefer to not have to revisit them.

dbouwman: @rweber-esri - I think this function is useful, but should just take a UserSession. FWIW - I ended up creating a [similar function](https://github.com/Esri/solution.js/blob/386c9c60f8e72f325a19a1bf7b77c58df9dd3fa5/packages/common/src/create-hub-request-options.ts#L34) in Solution.js and that takes a UserSession, and some solutions-specific stuff b/c there are scenarios with Solutions where we need the portalSelf of a different org.

As for what to pass into components... 100% agree that we don't want every component re-fetching portalSelf and the user. That said I'm not certain how best to proceed - i.e. support session or some "HubSession" object which would basically contain the IHubRequestOptions info, but implemented as a class... this is part of the whole Hub.js "class-like-wrappers" project that never seems to get attention :)

As for passing session into storybook - at this point, that's still an open question. I'm _thinking_ that we need to wrap the components into a `<arcgis-identity>` in each story, but we need to build that component first.

packages/discussions/src/types.ts
rweber-esri: I updated the A.C. in [2472](https://devtopia.esri.com/dc/hub/issues/2472) to remove the URI validation and flagged the `DiscussionType.ITEM` and `DiscussionType.DATASET` enum pairs for removal when v10.0.0 publishes.

packages/common/src/utils/get-hub-request-options.ts
rweber-esri: Per our side discussion, @dbouwman, I've removed the `getHubRequestOptions` utility method I originally included on this PR since we don't have an immediate need for it, it was out of scope for this story to begin with, and it's implementation likely needed additional consideration.

dbouwman (approved):

PR #691, opened 2021-12-03
fix(hub-common): setContentType() now uses normalizedType to calculate family

sonofflynn89:
affects: @esri/hub-common

Resolves https://devtopia.esri.com/dc/hub/issues/2123.

Using normalizedType to calculate family solves the issue of legacy item types. For example, Hub
Initiative Templates used to be designated as type: 'Hub Initiative' with typeKeywords:
['hubInitiativeTemplate']. Running these old initiative templates through setContentType() used to
set family incorrectly as 'initiative'. Now family is correctly set to 'template'.

### DIFF
--- a/packages/common/src/content.ts
+++ b/packages/common/src/content.ts
@@ -743,7 +743,7 @@ export function datasetToItem(dataset: DatasetResource): IItem {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tomwayson (approved):
Nice work @sonofflynn89!

PR #693, opened 2021-12-14
feat(hub-common): add setContentBoundary() and setContentExtent()

tomwayson:
affects: @esri/hub-common

1. Description:

The goal of this PR is to ensure that Hub.js has the same logic as the server when deriving `content.boundary` for private or enterprise items.

1. addresses part of: [2356](https://devtopia.esri.com/dc/hub/issues/2356)


packages/common/src/content.ts
```diff
@@ -885,6 +891,66 @@ const getSolutionUrl = (content: IHubContent): string => {
...
+};
+
+const getContentBoundary = (content: IHubContent): IHubGeography => {
+  const item = content.item;
+  const extent = item.extent;
+  // user specified provenance is stored in item.properties
+  const provenance: GeographyProvenance =
+    item.properties?.boundary ||
+    // but we default to item if the item has an extent
+    (isExtentCoordinateArray(extent) ? "item" : undefined);
```
tomwayson, line 936: @drspacemanphd I want to confirm w/ you that this is effectively the same logic for deriving the default `boundary.provenance` that the server uses.

drspacemanphd: I believe so yes. If there is no set boundary in the item properties (either because a content editor has not set it or, potentially in the future, if the server has not already set a default one there), then the default boundary type should be `item`, and the boundary should correspond to the item's extent.

For clarity, I believe the server's get-all-boundaries endpoint returns item extents in GeoJSON format as an envelope with a list of coordinates. I assume that the `itemExtentToBoundary` function is doing something similar.

packages/common/src/content.ts
```diff
@@ -885,6 +891,66 @@ const getSolutionUrl = (content: IHubContent): string => {
   return hubUrl;
 };
 
+/**
+ * Create a new content with updated boundary properties
+ * @param content original content
+ * @param boundary boundary provenance
+ * @returns
+ */
+export const setContentBoundary = (
```
drspacemanphd, line 900: Is this being used anywhere aside from the test?

packages/common/src/content.ts
```diff
@@ -885,6 +891,66 @@ const getSolutionUrl = (content: IHubContent): string => {
...
+  return { ...updated, boundary: getContentBoundary(updated) };
+};
+
+/**
+ * Create a new content with updated extent and derived properties like boundary
+ * @param content original content
+ * @param extent new extent
+ * @returns a new content with the updated extent and boundary
+ */
+export const setContentExtent = (
```
drspacemanphd, line 917: Naming might be confusing, because it obtains a (non-extent) boundary and sets it as well.

packages/common/src/content.ts
```diff
@@ -885,6 +891,66 @@ const getSolutionUrl = (content: IHubContent): string => {
...
+  content: IHubContent,
+  extent: IBBox
+): IHubContent => {
+  // update content's item and extent
+  const item = { ...content.item, extent };
+  const updated = { ...content, item, extent };
```
drspacemanphd, line 923: Does this mean that `extent` becomes a property of both `content` and `content.item`?

packages/common/src/content.ts
tomwayson: not yet, but it will be used by opendata-ui

packages/common/src/content.ts
tomwayson: Unfortunately, all `content.item` properties are mirrored on `content` for backwards compatibility reasons, though in retrospect I wish I had chosen not to do that when we moved to Hub.js, even though it would have meant updating hundreds of LOC in the -ui app.

packages/common/src/content.ts
tomwayson: > get-all-boundaries endpoint returns item extents in GeoJSON format as an envelope with a list of coordinates. I assume that the itemExtentToBoundary function is doing something similar.

Good point. It looks like the server returns polygons:

https://hub.arcgis.com/api/v3/datasets/183adc24880b41c4be9fd6a14eb6165f_0

image

but `itemExtentToBoundary()` returns an envelope (xmin, xmax, etc). I should update that to return a polygon.

tomwayson: This re-affirms my conviction that the fns from the server should move into Hub.js and the server should consume them from there.

packages/common/src/content.ts
tomwayson: boundary is a property that depends on extent. and we don't have computed properties b/c this is a POJO. The way we decided to deal w/ this is to have setters for the lower level properties that derived properties depend on (in this case `extent`). Then those setters also update the derived properties (in this case `boundary`).

drspacemanphd (approved):

sonofflynn89 (approved):
Great work!

PR #695, opened 2021-12-22
B/set prop undefined

tomwayson:
1. Description:

fixes `setProp()` bug and refactors recently added `setContnetBoundary()` and `setContentExtent()` to be internal-only (they have not been released yet so that is not a breaking change)

1. Instructions for testing:

if tests pass we good


sonofflynn89 (approved):
🚀🚀🚀

PR #703, opened 2022-01-20
Add ArcGISContext class and Session LocalStorage Utils

dbouwman:
1. Description:
- Add ArcGISContext class and Session LocalStorage Utils

ArcGISContext is pretty similar to AppSettings class in hub-components, but more generically named, and moves localStorage from inside the class to separate utils.

Next step will be to pull this into -ui, and expose an instance on the app-settings service so that it is available for use with hub-components. Then, I'll update the hub-components package to use ArcGISContext instead of AppSettings, and finally update UI with latest of all packages, consuming ArcGISContext from the app-settings service.

1. Instructions for testing:
- run tests

### DIFF
--- a/packages/common/src/ArcGISContext.ts
+++ b/packages/common/src/ArcGISContext.ts
@@ -8,15 +8,19 @@ import { getProp, getWithDefault, IHubRequestOptions } from ".";
 import { IRequestOptions } from "@esri/arcgis-rest-request";
 
 /**
+ * Hash of Hub API end points so updates
+ * are centralized
+ */
+const hubApiEndpoints = {
+  domains: "/api/v3/domains",
+  search: "/api/v3/datasets",
+  discussions: "/api/discussions/v1",
+};
+
+/**
  * Options that can be passed into `ArcGISContext.create`
  */
 export interface IArcGISContextOptions {
-  /**
-   * ClientId for use with oAuth. Although this is optional,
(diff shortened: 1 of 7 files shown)


packages/common/src/ArcGISContext.ts
```diff
@@ -0,0 +1,364 @@
...
+   * ArcGIS Online or ArcGIS Enterprise portal url.
+   * Do not include  `/sharing/rest`
+   * Defaults to `https://www.arcgis.com`
+   * For ArcGIS Enterprise, you must include the webadaptor name.
+   * i.e. https://gis.mytown.gov/portal
+   *
+   * When Authentication is present, the UserSession.portal value is
+   * used instead of this property.
+   */
+  portalUrl?: string;
```
tomwayson, line 40: can't you inline the default value here like ` = 'https://www.arcgis.com';`?

packages/common/src/ArcGISContext.ts
```diff
@@ -0,0 +1,364 @@
...
+  private _currentUser: IUser;
+
+  private _debug = false;
+
+  /**
+   * Private constructor. Use `ArcGISContext.create(...)` to
+   * instantiate an instance
+   * @param opts
+   */
+  private constructor(opts: IArcGISContextOptions) {
```
tomwayson, line 86: TIL about private constructors.

tomwayson (approved):
Nice work @dbouwman

PR #708, opened 2022-01-24
Better naming for Context Classes

dbouwman:
1. Description:

Since we just created these classes, now is the time to refine the naming. As per suggestion from @tomwayson this renaming was done:

`ArcGISContext` -> `ArcGISContextManager`
`IArcGISContextState` -> `IArcGISContext`
`ArcGISContextState` -> `ArcGISContext`

Additionally, to be consistent, `_searchContent` will now add `.thumbnailUrl` to the content objects, appending in the token if the backing item is not public.

This also moved some utils out of the context class module, and exported them.

1. Instructions for testing:
- run tests


packages/common/src/ArcGISContextManager.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  /**
+   * @internal
+   * Log debugging messages to the console
+   * @param message
+   */
+  private log(message: string): void {
+    if (this._debug) {
+      // tslint:disable-next-line:no-console
+      console.info(message);
```
tomwayson, line 190: not a blocker for this PR, but probably should use https://esri.github.io/hub.js/api/common/Logger

tomwayson (approved):
Thanks for doing this @dbouwman!

PR #725, opened 2022-02-15
feat(hub-common): add isProxiedCSV and getProxyUrl utils

sonofflynn89:
affects: @esri/hub-common

1. Description:
For [2408](https://devtopia.esri.com/dc/hub/issues/2408), we needed a way to detect whether an item was a proxied csv. However, all proxied csv logic only exists in opendata-ui. This PR lifts a couple of needed utils into hub-common.

### DIFF
--- a/packages/common/src/content/index.ts
+++ b/packages/common/src/content/index.ts
@@ -2,7 +2,7 @@
  * Apache-2.0 */
 import { ResourceObject } from "jsonapi-typescript";
 import { IItem } from "@esri/arcgis-rest-portal";
-import { HubType, HubFamily, IModel } from "../types";
+import { HubType, HubFamily, IModel, IHubRequestOptions } from "../types";
 import { collections } from "../collections";
 import { categories as allCategories, isDownloadable } from "../categories";
 import { isBBox } from "../extent";
@@ -1016,3 +1016,22 @@ const getSolutionUrl = (content: IHubContent): string => {
 // instead of hardcoding the types here.
 const isPageContent = (content: IHubContent) =>
   includes(["Hub Page", "Site Page"], content.type);
+
+// proxied csv helpers
+const MAX_SIZE = 5000000;
+export const isProxiedCSV = (item: IItem, requestOptions: IHubRequestOptions) =>
+  !requestOptions.isPortal &&
+  item.access === "public" &&
+  item.type === "CSV" &&
+  item.size <= MAX_SIZE;
+
+export const getProxyUrl = (
+  item: IItem,
+  requestOptions: IHubRequestOptions
+) => {
+  let result;
+  if (isProxiedCSV(item, requestOptions)) {
+    result = `${requestOptions.hubApiUrl}/datasets/${item.id}_0/FeatureServer/0`;
+  }
+  return result;
+};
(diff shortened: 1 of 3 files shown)


tomwayson (changes requested):
Thanks for this @sonofflynn89 

I'd like to make it so that we only expose `getProxyUrl()`. You can can keep `isProxiedCSV()`, but move it to `_internal.ts`.

Otherwise, LGTM.

packages/common/src/content/index.ts
```diff
@@ -1016,3 +1016,22 @@ const getSolutionUrl = (content: IHubContent): string => {
...
+  item.type === "CSV" &&
+  item.size <= MAX_SIZE;
+
+export const getProxyUrl = (
+  item: IItem,
+  requestOptions: IHubRequestOptions
+) => {
+  let result;
+  if (isProxiedCSV(item, requestOptions)) {
+    result = `${requestOptions.hubApiUrl}/datasets/${item.id}_0/FeatureServer/0`;
```
sonofflynn89, line 1034: @tomwayson, it looks like opendata-ui has two divergent ways of defining the hubApiUrl:
- Without `/api/v3` ([hub-service mixin](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-opendata-components/addon/mixins/hub-service-mixin.js#L81))
- With `/api/v3` ([Layout Card Decorator](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-opendata-components/addon/decorators/layout-card.js#L16), note the use of `appSettings.hubApiUrl` instead of `appSettings.hubRootDomain`)

It seems like most of the app omits the `/api/v3`, but I wanted to verify the correct url with you.

sonofflynn89:
> Thanks for this @sonofflynn89
> 
> I'd like to make it so that we only expose `getProxyUrl()`. You can can keep `isProxiedCSV()`, but move it to `_internal.ts`.
> 
> Otherwise, LGTM.

@tomwayson thanks for the review! After further investigation, it looks like I will not need the proxied url for dcat distributions. The url for csv distributions uses the same endpoint for all supported item types (e.g. `https://my-site.hub.arcgis.com/datasets/<:id>.csv`). However, I'd `isProxiedCSV()` to quickly determine whether the CSV distribution should be included. Are you ok with that?

Also, I left a quick comment on `getProxyUrl` about a pertinent discrepancy in opendata-ui. Would you mind responding to that as well? Thanks!

packages/common/src/content/index.ts
tomwayson: good catch, I know that hub-service-mixin is widely used, and, based on this https://esri.github.io/hub.js/api/common/getHubApiUrl/ I think _without_ is correct, so it looks like this line needs to be updated to append it.

sonofflynn89: @tomwayson opendata-ui uses the pattern on line 1034 to define proxied csv urls (`${siteHostname}/datasets/${item.id}_0/FeatureServer/0`). I don't think we need to change it unless @drspacemanphd thinks otherwise.

tomwayson (approved):
Thanks @sonofflynn89!

PR #731, opened 2022-02-18
feat(hub-common): add composeContent()

tomwayson:

1. Description:

replaces existing implementations of itemToContent(), datasetToContent(), and getLayerContent()

1. Instructions for testing:

If tests pass w/ complete coverage, we good.

1. Part of : [2870](https://devtopia.esri.com/dc/hub/issues/2870)

### DIFF
--- a/packages/common/src/content/_internal.ts
+++ b/packages/common/src/content/_internal.ts
@@ -8,16 +8,22 @@
  * It's probably a good pattern to add functions here first and then
  * move them to index.ts only when they are needed by a consumer.
  */
-
-import { IItem } from "@esri/arcgis-rest-types/dist/types/item";
+import {
+  ILayerDefinition,
+  IFeatureServiceDefinition,
+} from "@esri/arcgis-rest-feature-layer";
+import { IItem } from "@esri/arcgis-rest-portal";
+import { ISpatialReference } from "@esri/arcgis-rest-types";
+import { IHubContent } from "../core";
 import {
   BBox,
-  IHubContent,
   IHubGeography,
   GeographyProvenance,
   IHubRequestOptions,
-} from "..";
+} from "../types";
 import { bBoxToPolygon, isBBox } from "../extent";
+import { getFamily } from "./get-family";
+import { getProp } from "../objects";
 
 /**
  * Create a new content with updated boundary properties
@@ -33,29 +39,11 @@ export const setContentBoundary = (
   const properties = { ...(content.item.properties || {}), boundary };
   const item = { ...content.item, properties };
   const updated = { ...content, item };
-  return { ...updated, boundary: getContentBoundary(updated) };
+  return { ...updated, boundary: getContentBoundary(item) };
 };
 
-/**
- * Create a new content with updated extent and derived properties like boundary
- * @param content original content
- * @param extent new extent
- * @returns a new content with the updated extent and boundary
- */
-export const setContentExtent = (
-  content: IHubContent,
-  extent: BBox
-): IHubContent => {
-  // update content's item and extent
-  const item = { ...content.item, extent };
-  const updated = { ...content, item, extent };
-  // derive boundary from content properties
-  const boundary = getContentBoundary(updated);
-  return { ...updated, boundary };
-};
-
-const getContentBoundary = (content: IHubContent): IHubGeography => {
-  const item = content.item;
+// TODO: this needs to be able to handle "automatic"
+export const getContentBoundary = (item: IItem): IHubGeography => {
   const extent = item.extent;
   const isValidItemExtent = isBBox(extent);
   // user specified provenance is stored in item.properties
@@ -80,7 +68,12 @@ const getContentBoundary = (content: IHubContent): IHubGeography => {
   };
 };
 
-const MAX_SIZE = 5000000;
+// when no request options are provided, the underlying
+// request functions assume that we are online in production
+// so we only want use enterprise logic if isPortal is explicitly defined
+export const isPortal = (requestOptions?: IHubRequestOptions) => {
+  return requestOptions && requestOptions.isPortal;
+};
 
 /**
  * Returns whether or not an item is a proxied csv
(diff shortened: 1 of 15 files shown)


packages/common/src/content/compose.ts
```diff
@@ -0,0 +1,755 @@
...
+  requestOptions?: IHubRequestOptions;
+}
+
+/**
+ * Compose a new content object out of an item, enrichments, and context
+ * @param item
+ * @param options any enrichments, current state (selected layerId), or context (requestOptions)
+ * @returns new content object
+ */
+export const composeContent = (
```
tomwayson, line 493: most of this PR is shuffling exising fns, aronud, but here's the actual new fn

packages/common/src/content/index.ts
```diff
@@ -433,51 +256,7 @@ export function getItemHubType(itemOrType: IItem | string): HubType {
...
-    createdDate,
-    createdDateSource,
-    publishedDate: createdDate,
-    publishedDateSource: createdDateSource,
-    updatedDate: new Date(item.modified),
-    updatedDateSource: "item.modified",
-    structuredLicense: getStructuredLicense(item.licenseInfo),
-  });
-  return setContentExtent(setContentType(content, item.type), item.extent);
+  return composeContent(item);
```
tomwayson, line 259: `itemToContent()` is now essentially a pass-through to `composeContent()`

packages/common/src/content/index.ts
```diff
@@ -490,95 +269,61 @@ export function itemToContent(item: IItem): IHubContent {
...
+    // setting this to null signals to enrichMetadata to skip this
+    metadata: metadata || null,
+    groupIds,
+    org,
+    server,
+    layers,
+    recordCount,
+    boundary,
+    statistics,
+  });
```
tomwayson, line 319: here's the payoff, `datasetToContent()` is now using the exact same logic as `itemToContent()`, it's just passing in more enrichments that we got form the API

packages/content/src/enrichments.ts
```diff
@@ -573,20 +576,47 @@ export const enrichContent = (
...
+        metadata,
+        groupIds,
+        org,
+        server,
+        layers,
+        recordCount,
+        boundary,
+        statistics,
+        requestOptions,
+      });
```
tomwayson, line 619: more payoff, `enrichContent()` is now using the _exact_ same logic as `datasetToContent()`, this is exactly what we're going to do in the new `fetchContent()` fn that will prioritize fetching fresh enrichments from portal instead over fetching from the Hub API.

packages/common/src/content/_internal.ts
```diff
@@ -89,8 +82,178 @@ const MAX_SIZE = 5000000;
...
+export const getHubRelativeUrl = (
+  type: string,
+  identifier: string,
+  typeKeywords?: string[]
+): string => {
+  // solution types have their own logic
+  let contentUrl = getSolutionUrl(type, identifier, typeKeywords);
+  if (!contentUrl) {
+    const family = getFamily(type);
+    const pluralizedFamilies = [
```
dbouwman, line 103: these don't seem pluralized?

packages/common/src/content/_internal.ts
```diff
@@ -89,8 +82,178 @@ const MAX_SIZE = 5000000;
...
+}
+
+export enum DatePrecision {
+  Year = "year",
+  Month = "month",
+  Day = "day",
+  Time = "time",
+}
+
+// TODO: move this to _internal and copy over tests from hub-content
```
dbouwman, line 191: drop comment?

dbouwman (approved):
All looks good to me - definitely complex so I'm not sure I'm really 👍🏻 -ing the logic but nothing looks O_o 

Only a few very minor comments on naming...

packages/common/src/content/_internal.ts
tomwayson: should be `const familiesWhoseRouteNamesAreThePluralizedFamily`

packages/common/src/content/compose.ts
```diff
@@ -0,0 +1,755 @@
...
+    org,
+    errors: errors || [],
+    // server enrichments
+    server,
+    layers,
+    recordCount,
+    // enrichments from the Hub API (boundary is below)
+    statistics,
+    // derived properties
+    // NOTE: in the getters below you can **not** use this.
```
sonofflynn89, line 586: Ah, it took me a second but now I see you're talking about the `this` keyword.

packages/common/test/content.test.ts
```diff
@@ -699,7 +711,11 @@ describe("setContentType", () => {
       expect(getContentTypeLabel("Application", false)).toEqual("application");
     });
     it("sets content type label to CSV if is proxied", () => {
-      expect(getContentTypeLabel("Application", true)).toEqual("CSV");
+      expect(getContentTypeLabel("PDF", true)).toEqual("CSV");
+    });
+    it("returns an empty string if no args???", () => {
+      // I'm just here for the coverage
```
sonofflynn89, line 717: Hahahaha way to be honest!

packages/content/src/enrichments.ts
```diff
@@ -573,20 +576,47 @@ export const enrichContent = (
...
+        slug,
+        errors,
+        data,
+        metadata,
+        groupIds,
+        orgId,
+        // TODO: org, ownerUser
+        server,
+        layers,
+        // TODO: these should be fetched instead of pulled from the dataset
```
sonofflynn89, line 598: Wait, fetched as in fetched from the item? Because it seems like we are fetching these values from somewhere...

sonofflynn89 (approved):
Heroic work 🎉 
Looks great, just left a few comments about comments (meta, right?) so they read clearer for posterity 👍

packages/content/src/enrichments.ts
tomwayson: in `fetchContent()` I'm going to have to add enrichments for these that fetch them from the Hub API.

sonofflynn89: Ah, got it. I was confused because this line is part of destructuring an object called `fetched`

packages/common/src/content/_internal.ts
```diff
@@ -89,8 +82,177 @@ const MAX_SIZE = 5000000;
...
+  updateFrequency: string;
+  metadataUpdateFrequency: string;
+  metadataUpdatedDate: string;
+  reviseDate: string;
+  pubDate: string;
+  createDate: string;
+}
+
+export function getMetadataPath(identifier: keyof IMetadataPaths) {
+  // NOTE: i have verified that this will work regardless of the "Metadata Style" set on the org
```
ajturner, line 163: This is likely because it's the ArcGIS metadata format - but we should verify and document that.

packages/common/src/content/_internal.ts
```diff
@@ -89,8 +82,177 @@ const MAX_SIZE = 5000000;
...
+ * This is because a) if somone entered 2018, we want to respect that and not treat it as the same as 2018-01-01
+ * and b) you cannot naively call new Date with an ISO 8601 string that does not include time information
+ * For example, when I, here in mountain time, do new Date('2018').getFullYear() I get "2017".
+ * This is because when you do not provide time or timezone info, UTC is assumed, so new Date('2018') is 2018-01-01T00:00:00 in UTC
+ * which is actually 7 hours earlier here in mountain time.
+ *
+ * @param {string} isoString
+ * @return { date: Date, precision: DatePrecision }
+ */
+export function parseISODateString(isoString: string) {
```
ajturner, line 202: Does Date.parse() work?

packages/common/src/content/get-family.ts
```diff
@@ -0,0 +1,44 @@
...
+function collectionToFamily(collection: string): string {
+  const overrides: any = {
+    other: "content",
+    solution: "template",
+  };
+  return overrides[collection] || collection;
+}
+
+/**
+ * return the Hub family given an item's type
```
ajturner, line 14: Please add a little clarity about what family is and how it relates to types

packages/common/src/content/_internal.ts
```diff
@@ -33,29 +39,11 @@ export const setContentBoundary = (
...
-  // derive boundary from content properties
-  const boundary = getContentBoundary(updated);
-  return { ...updated, boundary };
+  return { ...updated, boundary: getContentBoundary(item) };
 };
 
-const getContentBoundary = (content: IHubContent): IHubGeography => {
-  const item = content.item;
+// TODO: this needs to be able to handle "automatic"
+export const getContentBoundary = (item: IItem): IHubGeography => {
```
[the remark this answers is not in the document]
tomwayson: we can decided that if we ever make this a public function. it's currently only used for content.

packages/common/src/content/_internal.ts
mjuniper: I generally try to [avoid](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse) `Date.parse()`. But also see the comment above. We're trying to extract not just a date but also the level of precision the user intended.

packages/common/src/content/_internal.ts
```diff
@@ -80,7 +68,12 @@ const getContentBoundary = (content: IHubContent): IHubGeography => {
   };
 };
 
-const MAX_SIZE = 5000000;
+// when no request options are provided, the underlying
+// request functions assume that we are online in production
+// so we only want use enterprise logic if isPortal is explicitly defined
+export const isPortal = (requestOptions?: IHubRequestOptions) => {
```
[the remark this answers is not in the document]
tomwayson: I happen to know that this is the desired behavior in the current context in which this fn is used. If it turns out to be more generic then I agree that we should move this out of the content folder. I will keep an eye on that in future PR reviews.

packages/common/src/content/get-family.ts
tomwayson: Will do in a future PR. This PR only moves this fn.

PR #732, opened 2022-02-22
feat(hub-common): add removeDomainsBySiteId()

robertvo824:
affects: @esri/hub-common, @esri/hub-sites

rename `removeDomain()` to `removeDomainById()`
add `removeDomainsBySiteId()` to delete all domains associated with a site

BREAKING CHANGE:
`removeDomain()` renamed to `removeDomainById()`

ISSUES CLOSED: [#3230](https://devtopia.esri.com/dc/hub/issues/3230)

1. Description:

- Update `_removeSiteDomains()` to use the new Domains DELETE endpoint `deleteBySiteId()`
  - This allows `_removeSiteDomains()` to make a single request instead of requesting get all domain records by `siteId` and then sending parallel requests to delete each record

1. Instructions for testing:

1. Closes Issues: [#3230](https://devtopia.esri.com/dc/hub/issues/3230)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39250,7 +39250,7 @@
 		},
 		"packages/common": {
 			"name": "@esri/hub-common",
-			"version": "9.17.1",
+			"version": "9.19.0",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"abab": "^2.0.5",
@@ -39281,7 +39281,7 @@
 		},
 		"packages/content": {
 			"name": "@esri/hub-content",
-			"version": "9.17.1",
+			"version": "9.19.0",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"fast-xml-parser": "~3.2.4",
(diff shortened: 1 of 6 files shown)


robertvo824:
> @robertvo824
> 
> I haven't looked at the code yet, but we're not in a position to make a breaking change at this time - we have a large backlog of deprecations that should be made and have been put off for several major releases.
> 
> For now, can you please keep the old function but deprecate it and have it call the new one?

Yup, I can do that. But just to clarify, the old function is still being used elsewhere, but I just changed the name so that it was more descriptive since I'm adding a similar function. Which I don't think would've been a valid reason for a breaking change anyway. But I can just change the old function name back to the original and comment if that works. Thanks

tomwayson:
Thanks @robertvo824 

One thing - I'm not certain, but I think if we were to merge this as is, semantic release would still make a breaking change b/c of your first commit message. I think the best thing to do would be to rebase and squash w/ a new `feat` commit message that will trigger the correct release.

rgwozdz (approved):
LGTM, thanks @robertvo824.

PR #738, opened 2022-02-28
fix(hub-common): fix missing extent in datasetToContent()

tomwayson:
affects: @esri/hub-common

1. Description:

I should not have removed the extent enrichment logic from `datasetToContent()` 

1. Instructions for testing:

### DIFF
--- a/packages/common/src/content/index.ts
+++ b/packages/common/src/content/index.ts
@@ -5,6 +5,7 @@ import { IItem } from "@esri/arcgis-rest-portal";
 import { HubType, IModel } from "../types";
 import { getCollection } from "../collections";
 import { categories as allCategories } from "../categories";
+import { isBBox } from "../extent";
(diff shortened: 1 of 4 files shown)


tomwayson:
@robertvo824 caught this bug when he updated hub.js in opendata-ui and it caused an e2e to fail

packages/common/src/content/index.ts
```diff
@@ -287,6 +288,8 @@ export function datasetToContent(dataset: DatasetResource): IHubContent {
     recordCount,
     statistics,
     // additional attributes needed
+    extent,
+    itemExtent,
```
sonofflynn89, line 292: Do you need `itemExtent`? It seems like you just end up referencing `item.extent` down below.

sonofflynn89 (approved):

PR #740, opened 2022-03-06
[WIP] feat: add cache manager class and tests

dbouwman:
1. Description:

Created a `CacheManager` singleton, that can be used anywhere, with any function (async or not), with any number of parameters.

**Usage**: `await CacheManager.call(getModel, "3ef", {authentication});`

I looked at some memoize functors, but that just seemed overly verbose, and difficult to fold into our existing code style. This approach requires a minor change vs the normal a function call.

**CacheKeys**
The function name is used with the serialized arguments as the cache key, ensuring separation of cache entries across functions which take the same arguments. 

i.e. `getItem("3ef")` and `getItemData("3ef")` have separate cache keys despite having the same arguments

Downside is that it can't work with anonymous functions, and will `throw`. I can't see how that's a problem but comment if you do.

While this can work with non-async functions, `CacheManager.call(...)` *always* returns a Promise. I also don't see this being a big limitation, and could likely be surmounted, but given the chatty nature of our platform, I think this will mostly be used for async functions.

```js
// Normal Call
const itm = getItem("3ef");
// Cached
const itm2 = CacheManager.call(getItem, "3ef");
// subsequent calls to getItem through cache manager will return the same response

// since the cache holds the result of the function (i.e. a Promise) cache hits happen even before
// the first promise resolves. 
// e.g. this code below, will make one call, the other 99 are cache hits
const promises = [];
 for (let index = 0; index < 100; index++) {
    promises.push(CacheManager.call(getServiceInfoById, "3ef");
  }
await Promise.all(promises);
```

1. Instructions for testing:

Run Tests

### DIFF
--- a/packages/common/src/CacheManager.ts
+++ b/packages/common/src/CacheManager.ts
@@ -1,0 +1,95 @@
+/**
+ * Singleton Cache Manager
+ */
+export class CacheManager {
+  private static _instance: CacheManager;
+  public hits: number = 0;
+  public requests: number = 0;
+  private cache: Map<string, unknown>;
+
+  private constructor() {
+    this.cache = new Map<string, unknown>();
+  }
+
+  /**
+   * Convenience static method that directs calls to the instance
+   * @param fn
+   * @param args
+   * @returns
+   */
+  public static async call<Fn extends (...args: any) => any>(
+    fn: (...args: any[]) => ReturnType<Fn>,
+    ...args: any[]
+  ): Promise<ReturnType<Fn>> {
+    return CacheManager.instance.callFn(fn, ...args);
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


ajturner:
Great idea. I'll need to look more on computer, but initial thought is that cache should ideally be hidden inside primary operations. It's an implementation detail that most dev users shouldn't know about.

With this is mind, we can also bust the cache with modifying operations. e.g. updateItem would empty cache for that key. 

So we also need to consider how there are primary keys and optional secondary keys if there are more function parameters

dbouwman:
We have a cache implementation in hub-components -> https://github.com/Esri/hub-components/blob/beta/src/utils/cache.ts we should look to unify these

PR #741, opened 2022-03-09
HubSiteManager & HubSites module

dbouwman:
1. Description:

- Adds the `HubSiteManager` class, and it's backing `HubSites` module
- Adds `.sites` to the `Hub` class.
- DRY-ed up how `.search(..)` works for the Manager classes. They now delegate to `searchContentEntities(...)` which takes a conversion function that handles converting from an item to the entity type. These search functions *only* search against the Portal API - they never search using the Hub API. 
- additionally, Managers have `.fromItem(IItem, requestOptions)`

*Note*: The e2e's are intentionally brittle. They are not intended to be run in CI, but rather, are used during development / debugging to ensure the platform responds as expected. If we want to have ci-runnable e2e's we can add tasks to do that

**Note Part Deux**
FireFox tests failed in CI on this. Took a few hours of digging but the issue was due to rebuilding package-lock.json, which picked up `jasmine-core` 3.99.0` from `3.4.0`. Forcing things back to `3.4.0` resolved the issue, and I created an [issue to resolve](https://github.com/Esri/hub.js/issues/742)

1. Instructions for testing:
- run tests
- run e2e tests to see it work with real items

1. Closes Issues: #[3346](https://devtopia.esri.com/dc/hub/issues/3346)


package.json
```diff
@@ -53,7 +53,7 @@
 		"husky": "^4.3.0",
 		
 		"jasmine": "^3.6.4",
-		"jasmine-core": "^3.4.0",
+		"jasmine-core": "3.4.0",
```
dbouwman, line 56: This was needed to resolve some FireFox test failures. I've opened #742 to look into this

packages/common/src/ArcGISContext.ts
```diff
@@ -305,14 +305,14 @@ export class ArcGISContext implements IArcGISContext {
...
     // We may add more logic around what is returned in some corner cases
-    if (this.isAuthenticated) {
-      return {
-        authentication: this.session,
-        isPortal: this.isPortal,
-        portalSelf: this.portal,
-        hubApiUrl: this.hubUrl,
-      };
-    }
+
```
dbouwman, line 308: We need to be able to create `IHubRequestOptions` when not auth'd

packages/common/src/search/searchContentEntities.ts
```diff
@@ -0,0 +1,105 @@
...
+import { IItem } from "@esri/arcgis-rest-types";
+import {
+  expandContentFilter,
+  Filter,
+  getNextFunction,
+  IHubSearchOptions,
+  serializeContentFilterForPortal,
+} from ".";
+import { ISearchResponse } from "..";
+
```
dbouwman, line 12: Would appreciate input from @tomwayson on this - it all works - even after this is merged

packages/common/src/sites/HubSiteManager.ts
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * Sites are stored as Items in the Sharing API
+   * @param site
+   * @param requestOptions
+   * @returns
+   */
+  async create(
+    site: Partial<IHubSite>,
+    requestOptions?: IHubRequestOptions
+  ): Promise<IHubSite> {
```
dbouwman, line 95: Since this is a multi-step operation, I'm considering  returning the `IHubSite & OperationStack` but mulling how best to encapsulate that w/o it getting gross

packages/common/src/items/index.ts
```diff
@@ -16,3 +16,4 @@ export * from "./get-structured-license";
 export * from "./slugs";
 export * from "./normalize-solution-template-item";
 export * from "./setItemThumbnail";
+export * from "./registerBrowserApp";
```
tomwayson, line 19: does this _need_ to be part of the external API? Is anyone ever going to use this fn on it's own, or just `registerSiteAsApplication()`?

packages/common/src/sites/index.ts
```diff
@@ -1,41 +1,11 @@
...
 export * from "./domains";
 export * from "./drafts";
-export * from "./_ensure-telemetry";
+export * from "./fetchSiteModel";
 export * from "./get-site-by-id";
+export * from "./HubSiteManager";
+export * from "./HubSites";
+export * from "./registerSiteAsApplication";
 export * from "./site-schema-version";
+export * from "./themes";
```
tomwayson, line 10: ditto w/ all these - if we don't have a known place where we are going to call any of them from the Hub app, or hub-components, let's not export them. We can always export them later if needed. If we needlessly export them now it becomes very difficult to modify or remove them and adds to our maintenance costs.

packages/common/src/sites/themes.ts
```diff
@@ -0,0 +1,54 @@
...
+      family: "Avenir Next",
+    },
+  },
+};
+
+/**
+ * Return the default theme, extended with values from the Org's shared theme
+ * @param {Object} portalSelf Org's Portal object
+ */
+export function getOrgDefaultTheme(portalSelf: IPortal) {
```
tomwayson, line 46: why export this fn _and_ `DEFAULT_THEME`? If it's just for testing, then let's make sure that we don't export it from `./index.ts`

packages/common/src/core/helpers/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./PropertyMapper";
```
tomwayson, line 1: does this need to be part of the external API? I don't think so.

packages/common/src/core/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./traits";
 export * from "./types";
+export * from "./helpers";
```
tomwayson, line 3: these are helpers, do they need to be part of the external API?

packages/common/src/search/index.ts
```diff
@@ -7,3 +7,4 @@ export * from "./types";
 export * from "./_searchContent";
 export * from "./_searchGroups";
 export * from "./_searchUsers";
+export * from "./searchContentEntities";
```
tomwayson, line 10: Does this need to be exported, seems like it's just for our internal use, right?

packages/common/src/search/searchContentEntities.ts
tomwayson: I'm not sure what you are expecting not to work, but we definitely should **not** be importing from `"."` and `".."`  b/c of circular deps.

tomwayson (changes requested):
there's a lot of code here, so I only gave a cursory glance, esp w/ the tests. Main thing from my POV is to not export anything we don't need at the time of release in opendata-ui. We can always add exports to index.ts files later.

packages/common/src/sites/themes.ts
dbouwman: DEFAULT_THEME is exported so it can be re-exported from hub-sites package

packages/common/src/items/index.ts
dbouwman: needed b/c it's re-exported in sites package

packages/common/src/core/index.ts
dbouwman: changed to _internal

packages/common/src/sites/index.ts
dbouwman: either needed directly or re-exported in hub-sites

packages/common/src/search/searchContentEntities.ts
dbouwman: was more the  use of `createContentEntitySearchFn` which holds a closure over a `ItemToEntityFunction<T>`. It's really useful but a little complex. That said, the `IHubItemEntity` interface now requires an `fromItem(IItem):Promise<T>` which is the function needed in the closure... so... complex, but not random?

tomwayson (commented):
thanks for making these fns internal-only.

I'm still wary of the `from ".."`. I understand that it can cause mocking issues. The way I got around those w/ content was to move the shared fn out into it's own module, or a "lower level" module that could be safely `import`ed by all consuming locations.

dbouwman:
@tomwayson I'll see what I can do about those... fwiw - I _believe_ that's what vscode does when you move files around / automatically add the imports vs manually typing the import

tomwayson:
yep, that's exactly what it does, which is unfortunate. I have already pulled this, and  started fixing them locally, you want me to push what I have?

PR #744, opened 2022-03-16
feat(hub-common): add viewDefinition and recordCount to fetchContent(…

tomwayson:
…) response

affects: @esri/hub-common

1. Description:

fetchContent now fetches the definition for client-side layer views
and fetches the record count for
layers, tables, and proxied CSVs

1. Instructions for testing:

1. Closes Issues:

#322 (essentially)

There are additional top-level aliases we could add like `geometryType`, `capabilities`, `objectIdField`, however, we've decided for now to update the client-code instead to get those from `layer` or `server`.

### DIFF
--- a/packages/common/src/content/compose.ts
+++ b/packages/common/src/content/compose.ts
@@ -499,6 +499,51 @@ export interface IComposeContentOptions extends IHubContentEnrichments {
 }
 
 /**
+ * get the layer object for
+ * - an item that refers to a specific layer of a service
+ * - a multi-layer services (if a layer id was passed in)
+ * - a single layer feature service
+ * @param item
+ * @param layers the layers and tables returned from the service
+ * @param layerId a specific id
+ * @returns layer definition
+ * @private
+ */
+export const getItemLayer = (
+  item: IItem,
+  layers: ILayerDefinition[],
+  layerId?: number
+) => {
+  // if item refers to a layer we always want to use that layer id
+  // otherwise use the layer id that was passed in (if any)
+  const _layerIdFromUrl = getLayerIdFromUrl(item.url);
+  const _layerId = _layerIdFromUrl ? parseInt(_layerIdFromUrl, 10) : layerId;
+  return (
+    layers &&
+    (!isNil(_layerId)
+      ? // find the explicitly set layer id
+        layers.find((_layer) => _layer.id === _layerId)
+      : // for feature servers with a single layer always show the layer
+        isFeatureService(item.type) && getOnlyQueryLayer(layers))
+  );
+};
+
+// TODO: we should re-define ILayerDefinition
+// in IServerEnrichments.ts to include isView
+interface ILayerViewDefinition extends ILayerDefinition {
+  isView?: boolean;
+}
+
+/**
+ * determine if a layer is a layer view
+ * @param layer
+ * @returns
+ * @private
+ */
+export const isLayerView = (layer: ILayerDefinition) =>
+  (layer as ILayerViewDefinition).isView;
+
+/**
  * Compose a new content object out of an item, enrichments, and context
  * @param item
  * @param options any enrichments, current state (selected layerId), or context (requestOptions)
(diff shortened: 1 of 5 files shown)


tomwayson:
@sonofflynn89 

NOTE: the explore view will expect content to have a top-level `viewFilter`, which we'll now need to handle like:

```js
const getViewFilter = content => content.viewDefinition && `(content.viewDefinition.definitionExpression)`
```

packages/common/src/content/compose.ts
```diff
@@ -498,6 +498,51 @@ export interface IComposeContentOptions extends IHubContentEnrichments {
...
+ * - an item that refers to a specific layer of a service
+ * - a multi-layer services (if a layer id was passed in)
+ * - a single layer feature service
+ * @param item
+ * @param layers the layers and tables returned from the service
+ * @param layerId a specific id
+ * @returns layer definition
+ * @private
+ */
+export const getItemLayer = (
```
tomwayson, line 512: this fn was just hoisted out of `composeContent()` below

packages/common/src/content/fetch.ts
```diff
@@ -24,21 +27,38 @@ export interface IFetchContentOptions extends IFetchItemAndEnrichmentsOptions {
...
 
+const maybeFetchLayerEnrichments = async (
+  itemAndEnrichments: IItemAndEnrichments,
+  options?: IFetchContentOptions
+) => {
+  // determine if this is a client-side feature layer view
+  const { item, data, layers } = itemAndEnrichments;
+  const layer =
+    layers && getItemLayer(item, layers, options && options.layerId);
+  // TODO: add recordCount here too?
```
tomwayson, line 38: I'd prefer to fetch the record count here, where we fetch other enrichments, however, I need a lot of logic that currently only happens in `composeContent()` getters like `isProxied`, `url`, `viewDefinition`, so I'd have to turn all those getters into fns I could call here. So for now, I just fetch `recordCount` after calling `composeContent()`

packages/common/src/content/fetch.ts
```diff
@@ -65,7 +84,7 @@ const fetchContentById = async (
     requestOptions: options,
     ...itemEnrichments,
     slug,
-    layerId: specifiedLayerId || layerId,
+    layerId: isNil(specifiedLayerId) ? layerId : specifiedLayerId,
```
sonofflynn89, line 87: Good catch on the case where `specifiedLayerId` is the number 0!

sonofflynn89 (approved):
Looks good! Just a couple comments about imports. Before you merge I'd like to talk about some of the layerView stuff for my own edification 👍

packages/common/src/content/fetch.ts
```diff
@@ -24,21 +27,42 @@ export interface IFetchContentOptions extends IFetchItemAndEnrichmentsOptions {
...
+  itemAndEnrichments: IItemAndEnrichments,
+  options?: IFetchContentOptions
+) => {
+  // determine if this is a client-side feature layer view
+  const { item, data, layers } = itemAndEnrichments;
+  const layer =
+    layers && getItemLayer(item, layers, options && options.layerId);
+  // TODO: add recordCount here too?
+  const layerEnrichments =
+    layer && isLayerView(layer) && !data
```
tomwayson, line 40: I copied the current logic in isLayerView() from composer.js, which only checks `layer.isView`, but when creating a new layer view to test this I noticed that the item also gets the "View Service" type keyword. We _could_ use that in `fetchItemEnrichments()` to include "data" in the default list of enrichments to fetch for the item, however, there are layer views that don't have that type keyword (for example the item in [#1073](https://devtopia.esri.com/dc/hub/issues/1073)), so we'd still have to do this check _after_ fetching the layer.

packages/common/src/content/fetch.ts
```diff
@@ -24,21 +27,42 @@ export interface IFetchContentOptions extends IFetchItemAndEnrichmentsOptions {
...
+      ? // NOTE: I'm not sure what conditions causes a layer view
+        // to store (at least part of) it's view definition in item data
+        // it seems that most do not, but until we have a reliable signal
+        // we just fetch the item data for all layer views
```
tomwayson, line 44: Furthermore, spot checking a handful of such items from this:  
https://qaext.arcgis.com/sharing/rest/search?num=100&start=0&f=json&q=typekeywords%3A%22View%20Service%22
showed that very few had their definition stored in item data, ex:
https://qaext.arcgis.com/sharing/rest/content/items/864c9eaf76ab47ba93d7308bacf1a34f/data?f=json, so most of the time we're fetching the data for nothing.

PR #745, opened 2022-03-17
fix(hub-common): fix urls for single-layer feature services and fix 0…

tomwayson:
… falsy comparision

affects: @esri/hub-common

1. Description:

1. Instructions for testing:

1. Closes Issues: #<number> (if appropriate)

1. [x] ran commit script (`npm run c`)

PR #746, opened 2022-03-17
fix(hub-common): strip _ from content name and title; content.layer i…

tomwayson:
…s never false

affects: @esri/hub-common

1. Description:

1. Instructions for testing:

### DIFF
--- a/packages/common/src/content/compose.ts
+++ b/packages/common/src/content/compose.ts
@@ -36,8 +36,8 @@ import { getFamily } from "./get-family";
 
 // helper fns - move this to _internal if needed elsewhere
 const getOnlyQueryLayer = (layers: ILayerDefinition[]) => {
-  const layer = layers && layers.length === 1 && layers[0];
-  return layer && layer.capabilities.includes("Query") && layer;
+  const layer = layers && layers.length === 1 ? layers[0] : undefined;
+  return layer && layer.capabilities.includes("Query") ? layer : undefined;
 };
 
 const shouldUseLayerInfo = (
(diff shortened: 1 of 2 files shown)


packages/common/test/compose.test.ts
```diff
@@ -393,6 +393,17 @@ describe("composeContent", () => {
         "should set url"
       );
     });
+    it("public, multi-layer feature service w/o layerId", () => {
+      const layerContent = composeContent(item, { layers });
+      expect(layerContent.layer).toBeUndefined();
+      expect(layerContent.hubId).toBe(item.id);
+      expect(layerContent.type).toBe(item.type);
+      expect(layerContent.family).toBe("map");
+      expect(layerContent.url).toEqual(item.url);
```
tomwayson, line 402: @sonofflynn89 this addresses https://github.com/Esri/hub.js/pull/745#discussion_r829349011

packages/common/src/content/compose.ts
```diff
@@ -54,6 +54,18 @@ const shouldUseLayerInfo = (
...
 
+// this logic adapted from hub-indexer's determineName(), see
+// https://github.com/ArcGIS/hub-indexer/blob/8f4dd6f928124c1f35dd02bc11bd996191ee1160/packages/duke/compose/common.js#L7-L34
+const getContentName = (
+  item: IItem,
+  layer?: Partial<ILayerDefinition>,
+  _shouldUseLayerInfo = false
+) => {
+  return (
+    (_shouldUseLayerInfo ? layer.name : item.title || item.name) || ""
```
sonofflynn89, line 65: The only thing I worry about with this is that if `_shouldUseLayerInfo` is true and `layer.name` doesn't exist, we'll only get an empty string. Is it possible for a layer to NOT have a name assigned?

sonofflynn89 (approved):
Thank you! 🙏

PR #749, opened 2022-03-21
fix(hub-common): only fetch record count for feature layers and tables (e.g. not raster layers)

sonofflynn89:
affects: @esri/hub-common

1. Description:

Part of [2866](https://devtopia.esri.com/dc/hub/issues/2866)
Also fixes a parity issue where slugs passed in to fetch content were not respected (i.e. the slug of the specified layer was populated instead).

### DIFF
--- a/packages/common/src/content/fetch.ts
+++ b/packages/common/src/content/fetch.ts
@@ -126,14 +126,19 @@ const fetchContentBySlug = async (
       options
     );
   }
-  const { slug, boundary, extent, searchDescription, statistics } =
-    hubEnrichments;
+  // Note that we are not extracting the slug for the specified layer.
+  // It seems that the old client composer code always populated the slug
+  // field with the slug that was passed into the function (typically the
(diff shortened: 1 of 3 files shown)


packages/common/src/content/fetch.ts
```diff
@@ -174,9 +179,14 @@ export const fetchContent = async (
...
+  const { layer, type } = content;
   content.recordCount =
-    isProxied || !!layer ? await fetchContentRecordCount(content) : undefined;
+    !!layer && hasFeatures(type)
+      ? await fetchContentRecordCount(content)
+      : undefined;
   return content;
 };
+
+const hasFeatures = (contentType: string) =>
```
tomwayson, line 191: Style suggestion, I prefer to move these helper fns to the top of the file, right under the `import` statements, since they may one day become imports if they end up needing to move out of the file to be used elsewhere.

tomwayson (approved):
Well done @sonofflynn89

PR #752, opened 2022-03-21
fix(hub-common): fetchContentRecordCount did not pass auth nor handle…

tomwayson:
… errors

affects: @esri/hub-common

1. Description:

Pass authentication and other request options when fetching record count. Also implement existing error handling logic from composer (i.e. return infinity)

1. Instructions for testing:

1. Closes Issues: n/a

### DIFF
--- a/packages/common/test/content/fetch.test.ts
+++ b/packages/common/test/content/fetch.test.ts
@@ -494,10 +494,12 @@ describe("fetchContent", () => {
             "fetchItemEnrichments"
           ).and.returnValue(Promise.resolve(itemEnrichments));
           const count = 1000;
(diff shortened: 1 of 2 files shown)


packages/common/test/content/fetch.test.ts
```diff
@@ -494,10 +494,12 @@ describe("fetchContent", () => {
             "fetchItemEnrichments"
           ).and.returnValue(Promise.resolve(itemEnrichments));
           const count = 1000;
```
sonofflynn89, line 496: Looks like we can delete this const since we're no longer using it.

sonofflynn89 (approved):
Thanks! 🙏

PR #758, opened 2022-03-25
fix(hub-common): fix composeContent to prefer layer spatialReference to item spatialReference

sonofflynn89:
affects: @esri/hub-common

1. Description:

Part of [2866](https://devtopia.esri.com/dc/hub/issues/2866).

It turns out that `item.spatialReference` isn't always a wkid. Sometimes, it is the name of a wkt (but not the full wkt itself). This is problematic since we don't currently have a way in our applications to lookup the full wkt that we would need for geospatial operations.

As such, we decided to give preference to the layer spatial reference, since it is a more predictable source of truth.
We also eliminated referencing the server's spatial reference since we are unsure if different layers can have different spatial references.


tomwayson (approved):
Thanks @sonofflynn89!

PR #759, opened 2022-03-30
fix(hub-common): fix additional resource logic to handle metadata props possibly being arrays

sonofflynn89:
affects: @esri/hub-common

1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/3339

### DIFF
--- a/packages/common/src/content/_internal.ts
+++ b/packages/common/src/content/_internal.ts
@@ -347,23 +347,60 @@ export const getAdditionalResources = (
   metadata?: any,
   requestOptions?: IHubRequestOptions
 ): IHubAdditionalResource[] => {
-  let rawResources: IAGOAdditionalResource | IAGOAdditionalResource[] =
-    getProp(metadata, "metadata.distInfo.distTranOps.onLineSrc") || null;
-
-  if (rawResources === null) {
-    return null;
-  }
-
-  // Coerce to array since rawResources will be an object if only 1 resource is available
-  rawResources = Array.isArray(rawResources) ? rawResources : [rawResources];
-
-  return rawResources.map(
-    (resource: IAGOAdditionalResource): IHubAdditionalResource => ({
-      name: resource.orName,
-      url: getAdditionalResourceUrl(resource, item, requestOptions),
-      isDataSource: isDataSourceOfItem(resource, item),
-    })
+  const rawResources: IAGOAdditionalResource[] = extractRawResources(metadata);
+  return (
+    rawResources &&
+    rawResources.map(
+      (resource: IAGOAdditionalResource): IHubAdditionalResource => ({
+        name: resource.orName,
+        url: getAdditionalResourceUrl(resource, item, requestOptions),
+        isDataSource: isDataSourceOfItem(resource, item),
+      })
+    )
   );
+};
+
+/**
+ * Extracts additional resources from formal item metadata.
+ * If none are available, null is returned.
+ *
+ * @param metadata the formal item metadata
+ * @returns an array of all additional resources, or null
+ */
+export const extractRawResources = (
+  metadata?: any
+): IAGOAdditionalResource[] => {
+  const rawResources: IAGOAdditionalResource[] = [];
+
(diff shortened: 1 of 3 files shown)


packages/common/src/content/_internal.ts
```diff
@@ -347,23 +347,60 @@ export const getAdditionalResources = (
...
-    (resource: IAGOAdditionalResource): IHubAdditionalResource => ({
-      name: resource.orName,
-      url: getAdditionalResourceUrl(resource, item, requestOptions),
-      isDataSource: isDataSourceOfItem(resource, item),
-    })
+  // The property path to additional resources should be fairly simple.
+  // In many cases, it's just `metadata.metadata.distInfo.distTranOps.onLineSrc`.
+  // However, since `distInfo`, `distTranOps` and `onLineSrc` can be either
+  // Objects OR Arrays, we have to do all this looping.
+  maybeCastToArray(getProp(metadata, "metadata.distInfo")).forEach(
```
sonofflynn89, line 379: @drspacemanphd this is my interpretation of your indexer commit, without using lodash. Would you mind double checking it?

packages/common/src/content/_internal.ts
```diff
@@ -347,23 +347,60 @@ export const getAdditionalResources = (
...
 
-  // Coerce to array since rawResources will be an object if only 1 resource is available
-  rawResources = Array.isArray(rawResources) ? rawResources : [rawResources];
+/**
+ * Extracts additional resources from formal item metadata.
+ * If none are available, null is returned.
+ *
+ * @param metadata the formal item metadata
+ * @returns an array of all additional resources, or null
+ */
```
tomwayson, line 369: this probably needs a `@private` to keep it out of the docs

tomwayson (commented):
I'll let @drspacemanphd weigh in on the logic.

packages/common/src/content/_internal.ts
drspacemanphd: Yes that is right. "Officially", you can only make `distTranOps` and `onLineSrc` arrays via the metadata editor, but scripting shows that AGO will not invalidate `distInfo` arrays

packages/common/src/content/_internal.ts
```diff
@@ -347,23 +347,60 @@ export const getAdditionalResources = (
...
+
+/**
+ * Arrays are returned as-is.
+ * Objects are wrapped into a 1 element array.
+ * Undefined values return an empty array.
+ *
+ * @param objectOrArray
+ * @returns the casted array
+ */
+const maybeCastToArray = (objectOrArray: any = []) => {
```
drspacemanphd, line 402: just name this as castArray. The "maybe" is confusing because it suggests that sometime is _won't_ when in reality the result is always an array. Also, just an FYI, `null` is not treated as identical to `undefined` when it comes to default parameters. If `null` is provided as `objectOrArray`, the return from this will be an array of length 1 with the only value being `null`.

packages/common/test/content.test.ts
```diff
@@ -961,12 +964,57 @@ describe("getAdditionalResources", () => {
...
+    };
+    const result = extractRawResources(metadata);
+    expect(result.length).toEqual(1);
+    expect(result[0]).toEqual({
+      orName: "Resource Name",
+      linkage: "resource-url",
+    });
+  });
+
+  it("handles when distInfo is an Array", () => {
```
drspacemanphd, line 998: if other combinations are not covered by tests, I would put them here. Namely the matrix of:
- distInfo as object, array of 1, or array of 2
- distTransOps as object, array of 1, or array of 2
- onLineSrc as object, array of 1, array of 2

packages/common/src/content/_internal.ts
drspacemanphd: Also if `getProp` allows, this would be where I would put a default value of an empty array in case `distInfo` does not exist.

drspacemanphd (changes requested):
Looks good for most part but some small changes.

PR #760, opened 2022-03-30
B/search description

tomwayson:
1. Description:

use searchDescription for content.summary, not content.description

also fixes tests for relativeDateToDateRange() that would only fail at the end of months that are longer than the previous month. Not sure if we consider the underlying logic a bug, but wasn't sure what logic we want. For example: what should be the from date of the range that represents the previous 1 month form 3/30? There is no 2/30. So I kept the existing logic under the assumption that if it's good enough for `Date.setDate()` it's good enough for us, and just updated the test expectations.

1. Instructions for testing:

1. Closes Issues: [3507](https://devtopia.esri.com/dc/hub/issues/3507)

### DIFF
--- a/packages/common/src/search/utils.ts
+++ b/packages/common/src/search/utils.ts
@@ -218,6 +218,9 @@ export function relativeDateToDateRange(
       break;
     case "months":
       // get the current month and subtract num
+      // NOTE: when the previous month has fewer days than this month
+      // setMonth() will return a date w/in the current month
+      // example: 3/30 -> 3/2 b/c there is no 2/28
       now.setMonth(now.getMonth() - relative.num);
       result.from = now.getTime();
       break;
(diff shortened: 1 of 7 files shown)


packages/common/src/search/utils.ts
```diff
@@ -218,6 +218,9 @@ export function relativeDateToDateRange(
       break;
     case "months":
       // get the current month and subtract num
+      // NOTE: when the previous month has fewer days than this month
+      // setMonth() will return a date w/in the current month
+      // example: 3/30 -> 3/2 b/c there is no 2/28
```
tomwayson, line 223: bah, there is no 2/30

dbouwman: dates-- thanks for adding this note and fixing the test

dbouwman (approved):
Code looks good to me, but I'm not exactly sure about the expected logic. I assume @sonofflynn89  is covering that

sonofflynn89 (approved):
Good work!

PR #762, opened 2022-03-31
fix(hub-common): defaults q param of fetchItemsBySlug to be slug and adds portal

drspacemanphd:
affects: @esri/hub-common

1. Description - This PR does the following.
  - Ensures any portal is passed on in case no auth is provided.
  - Removes a wrapped error to maintain original error status code and message
  - Handles null pointer when no project matching an identifier is found.

3. Instructions for testing:
Modified code in Hub Project API and saw proper responses.

4. [x] ran commit script (`npm run c`)


dbouwman (approved):
Looks good!

PR #763, opened 2022-04-01
fix(hub-common): re-fetch target layer for legacy services that return sparsely populated layers

sonofflynn89:
affects: @esri/hub-common

1. Description:
Fix for https://devtopia.esri.com/dc/hub/issues/3488
Targeted strike for older services (version < 10.5)

### DIFF
--- a/packages/common/src/content/fetch.ts
+++ b/packages/common/src/content/fetch.ts
@@ -1,4 +1,8 @@
-import { queryFeatures } from "@esri/arcgis-rest-feature-layer";
+import {
+  getLayer,
+  parseServiceUrl,
+  queryFeatures,
+} from "@esri/arcgis-rest-feature-layer";
 import { getItem } from "@esri/arcgis-rest-portal";
 import { IHubContent } from "../core";
 import {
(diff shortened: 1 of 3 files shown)


tomwayson (approved):
Great work @sonofflynn89

packages/common/src/content/fetch.ts
```diff
@@ -36,9 +40,25 @@ const maybeFetchLayerEnrichments = async (
   options?: IFetchContentOptions
 ) => {
   // determine if this is a client-side feature layer view
-  const { item, data, layers } = itemAndEnrichments;
-  const layer =
-    layers && getItemLayer(item, layers, options && options.layerId);
+  const { item, data } = itemAndEnrichments;
+  let { layers } = itemAndEnrichments;
```
tomwayson, line 44: @sonofflynn89 

what I was suggesting we could do to expand support to content where `enrichLayers()` failed b/c `getAllLayersAndTables()` didn't return `layers` and `tables` is that we could  modify this line to:

```ts
let layers = itemAndEnrichments.layers || mergeLayersAndTables(itemAndEnrichments.server)
```

which I think is more in line w/ how `_composeLegacyFeatureLayer()` worked.

PR #764, opened 2022-04-04
fix(hub-common): fix issues composing proxied csvs

sonofflynn89:

1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/3512

Fixes incorrect fetching and composing of Proxied CSVs. These changes include:
- `fetchContent` sets the `item.url` of proxied csvs to point to the proxying feature service
- `fetchContent` grabs layer and server enrichments for proxied csvs
- `composeContent` sets the `hubId` of proxied csvs to the `item.id`
- `composeContent` correctly handles `extent` fallback scenarios and sets `extent` to be a bbox

### DIFF
--- a/packages/common/src/content/compose.ts
+++ b/packages/common/src/content/compose.ts
@@ -8,7 +8,7 @@ import { IHubRequestOptions } from "../types";
 import { getHubApiUrl } from "../api";
 import { isDownloadable } from "../categories";
 import { IHubContentEnrichments, IHubContent } from "../core";
-import { isBBox } from "../extent";
+
 import { getStructuredLicense } from "../items/get-structured-license";
 import { getProp } from "../objects";
 import { getItemThumbnailUrl } from "../resources/get-item-thumbnail-url";
@@ -30,6 +30,7 @@ import {
   isProxiedCSV,
   parseISODateString,
   getAdditionalResources,
+  determineExtent,
 } from "./_internal";
 import { getFamily } from "./get-family";
 
@@ -40,16 +41,18 @@ const getOnlyQueryLayer = (layers: ILayerDefinition[]) => {
 };
 
 const shouldUseLayerInfo = (
+  item: IItem,
   layer: Partial<ILayerDefinition>,
   layers: Array<Partial<ILayerDefinition>>,
-  url: string
+  requestOptions?: IHubRequestOptions
 ) => {
   return (
+    !isProxiedCSV(item, requestOptions) &&
     layer &&
     layers &&
     layers.length > 1 &&
     // we use item info instead of layer info for single layer items
-    !getLayerIdFromUrl(url)
+    !getLayerIdFromUrl(item.url)
   );
 };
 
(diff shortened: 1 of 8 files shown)


packages/common/src/content/compose.ts
```diff
@@ -40,16 +41,18 @@ const getOnlyQueryLayer = (layers: ILayerDefinition[]) => {
...
 
 const shouldUseLayerInfo = (
+  item: IItem,
   layer: Partial<ILayerDefinition>,
   layers: Array<Partial<ILayerDefinition>>,
-  url: string
+  requestOptions?: IHubRequestOptions
 ) => {
   return (
+    !isProxiedCSV(item, requestOptions) &&
```
sonofflynn89, line 50: @tomwayson I made this change here because as far as I can tell, we should never use layer info for proxied csvs (affects `name` and `description`)

packages/common/src/content/compose.ts
```diff
@@ -556,6 +560,24 @@ interface ILayerViewDefinition extends ILayerDefinition {
 export const isLayerView = (layer: ILayerDefinition) =>
   (layer as ILayerViewDefinition).isView;
 
+const determineHubId = (
```
sonofflynn89, line 563: I abstracted this out to a helper to prevent triple nested ternaries, but didn't feel that we need to expose this logic for others to consume. I left it in `compose.ts` because `getItemHubId` is a helper that is defined within this file.

packages/common/src/content/compose.ts
```diff
@@ -694,15 +713,8 @@ export const composeContent = (
         // TODO: groups?
       };
     },
-    // TODO: do we want to add dataset extent logic as a getter here?
```
sonofflynn89, line 697: Is the dataset extent logic referenced in this comment the same checks I put in `determineExtent`?

packages/common/src/content/compose.ts
tomwayson: yeah, I was thinking the same.

packages/content/src/enrichments.ts
```diff
@@ -615,6 +615,7 @@ export const enrichContent = (
         groupIds,
         org,
         server,
+        // TODO: fetch and pass in extent enrichment
```
sonofflynn89, line 618: I think I may have found a bug in `enrichContent` since tests started failing once I made my extent changes. The old tests had incorrect assumptions and were passing because of unintended behavior.

tomwayson (approved):
excellent work @sonofflynn89

packages/common/src/content/compose.ts
tomwayson: thanks for checking, but I'm not sure. I think this was an out dated comment.

packages/content/src/enrichments.ts
tomwayson: well, we're trying to kill this, so maybe we won't have to fix it?

PR #777, opened 2022-04-22
Feat/3227 create content

benstoltz:
1. Description: Moves create item from url/local file logic from ember to hub.js.

1. Instructions for testing:

1. [ ] ran commit script (`npm run c`)


packages/common/src/items/_internal/_wait-for-item-ready.ts
```diff
@@ -0,0 +1,34 @@
...
+      id: itemId,
+      ...requestOptions,
+    });
+
+    if (statusResult.status === "completed") {
+      break;
+    } else if (statusResult.status === "failed") {
+      throw new Error(statusResult.statusMessage);
+    }
+  } while (true);
```
rweber-esri, line 33: The `while` condition should be `statusResult?.status !== "completed"` and then conditionally throw when `statusResult\.status === "failed"`.

packages/common/src/items/_internal/_prepare-upload-requests.ts
```diff
@@ -0,0 +1,46 @@
...
+  requestOptions: IUserRequestOptions
+): any[] {
+  const queue = [];
+  // part number starts from 1
+  let partIndex = 1;
+  let sizeIndex = 0;
+
+  // Slice the file data and create an upload request for each part
+  while (sizeIndex < file.size) {
+    queue.push(() =>
```
rweber-esri, line 31: Can simplify this to return just the objects to use w/ `batch`. Might need to reverse the order of items in array since `_worker` called `pop`?

```js
queue.push({
        file: file.slice(sizeIndex, sizeIndex + sizeLimit, file.type),
        owner,
        id,
        partNum: partIndex,
        ...requestOptions,
});
```

packages/common/src/items/create-content-with-file.ts
```diff
@@ -0,0 +1,78 @@
...
+    await cancelItemUpload({
+      id: itemId,
+      owner: item.owner,
+      ...requestOptions,
+    });
+
+    throw e;
+  }
+
+  return itemId;
```
rweber-esri, line 77: We should return the created item, not its id.

packages/common/src/items/create-content-with-url.ts
```diff
@@ -0,0 +1,30 @@
...
+    owner: item.owner,
+    file: item.file,
+    dataUrl: item.dataUrl,
+    text: item.text,
+    multipart: item.multipart,
+    async: item.async,
+    ...requestOptions,
+  });
+  // return the newly created item id
+  return createResult.id;
```
rweber-esri, line 29: Return the created item, not it's id.

packages/common/src/items/create-content.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (item.access !== "private") {
+    await setItemAccess({
+      id: itemId,
+      owner: item.owner,
+      access: item.access,
+      ...requestOptions,
+    });
+  }
+
+  return itemId;
```
rweber-esri, line 51: Return the created content, not it's id

rweber-esri (changes requested):

packages/common/src/items/create-content-with-url.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Create AGO item from a URL
+ *
+ * @export
+ * @param {IItemAdd} item Item to be uploaded into online.
+ * @param {IUserRequestOptions} requestOptions
+ * @return {*}  {Promise<string>} Newly created item ID
+ */
+export async function createContentWithUrl(
```
rweber-esri, line 13: Similar to above, this may be better named `createItemWithUrl`

packages/common/src/items/create-content-with-file.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Creates an item in online from a local file/item.
+ * Upload is multithreaded as the item is chunked up.
+ *
+ * @export
+ * @param {IItemAdd} item Item to be uploaded into online.
+ * @param {IUserRequestOptions} requestOptions
+ * @return {*}  {Promise<string>} Newly created item id
+ */
+export async function createContentWithFile(
```
rweber-esri, line 23: I'd not considered this before, but the name `createContentWithFile` seems to imply it'll create/resolve an object that implements IHubContent interface, though it resolves `ICreateItemResponse`. Maybe this method would be better named `createItemWithFile` since it lives in the `items` namespace, and it's returning the result of the RESTJS `createItem` function?

packages/common/src/items/create-content.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Creates an item in online from either url or file.
+ * Once created we wait for the item to be ready (or throw an error if creation failed)
+ * If access is not private then we make a call to update that.
+ *
+ * @export
+ * @param {IItemAdd} item Item to be uploaded into online
+ * @param {IUserRequestOptions} requestOptions
+ * @return {*}  {Promise<string>} AGO item id
+ */
+export async function createContent(
```
rweber-esri, line 18: Similar to above. We're resolving the result from a call to RESTJS `createItem` and we're in the `items` namespace, , this may be better named `createItem`.

benstoltz: As discussed went with `createItemFromUrlOrFile`

rweber-esri (approved):
LGTM, thanks for making those changes. Nice work!

PR #783, opened 2022-05-05
refactor(hub-common): remove deprecated hub-content package

tomwayson:
1. Description:

1. Instructions for testing:

1. Part of [3319](https://devtopia.esri.com/dc/hub/issues/3319)


packages/common/src/content/_internal.ts
```diff
@@ -183,6 +183,8 @@ const getSolutionUrl = (
       // deployed solutions go to the generic content route
       hubUrl = `/content/${identifier}`;
     }
+    // TODO: I think this is a bug. The line below should be w/in else
+    // in order to avoid overwriting the above for deployed solutions, right?
```
dbouwman, line 187: I think you are correct @tomwayson

packages/common/src/content/index.ts
```diff
@@ -438,11 +438,12 @@ export const getContentTypeLabel = (
 };
 
 /**
- * returns a new content that has the specified hubId and updated identifier
+ * DEPRECATED returns a new content that has the specified hubId and updated identifier
```
dbouwman, line 441: If this PR removes a package... which I expect is breaking change... can we remove these functions as well?

dbouwman (approved):
This seems ok to me, but the stats on the PR seem weird... this only has 43 files changes, but it's saying 84,000 lines added? 
image
Before merging, double check that it's some O_O diff thing. Also if we're going to do a major with this, let's kill more deprecated fns

packages/common/src/content/index.ts
tomwayson: it's not a breaking change to remove the package from source code - the package still exists on NPM and is unchanged. 😉

packages/common/src/content/_internal.ts
tomwayson: thx. I wonder if we have some code in -ui that is correcting for this bug. Maybe @vivzhang knows?

packages/common/src/content/index.ts
dbouwman: 🎉🎉🎉🎉

tomwayson:
the large diff is due to package-lock.json:

image

I would have expected it to be red, I suspect the reason it's green is that by adding the demo packages back in, package-lock.json includes all their (dev) dependencies like webpack, Ember, etc, which as we know are quite hefty.

PR #787, opened 2022-05-11
fix(hub-common): fix issue where enrichServer actually fetched the layer for reference layer items

sonofflynn89:
affects: @esri/hub-common

1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/3861

`enrichServer` was returning the layer definition instead of the server definition for items that had a reference layer url (i.e. `.../FeatureServer/:layer`). This is because we mistakenly assumed that Rest.js' `getService` function stripped `/:layer` from the url before making the call. It does not, so we have to do it on our own.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61569,7 +61569,7 @@
 		},
 		"packages/common": {
 			"name": "@esri/hub-common",
-			"version": "9.29.0",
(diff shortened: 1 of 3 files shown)


tomwayson (approved):
Thank you for this @sonofflynn89

PR #793, opened 2022-05-17
fix(hub-initiatives): filter group search by title

MarvinPerry:
1. Description:
 - adds title filter to the query used by`checkGroupExists`  function to eliminate false positives due to fuzzy agol search

### DIFF
--- a/packages/initiatives/src/groups.ts
+++ b/packages/initiatives/src/groups.ts
@@ -7,7 +7,7 @@ import {
   searchGroups,
   removeGroup,
   unprotectGroup,
-  IUserGroupOptions
+  IUserGroupOptions,
 } from "@esri/arcgis-rest-portal";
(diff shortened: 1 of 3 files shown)


packages/initiatives/src/groups.ts
```diff
@@ -122,13 +122,14 @@ export function checkGroupExists(
   requestOptions: IRequestOptions
 ): Promise<any> {
   const options = {
-    q: `("${title}") AND orgid: ${orgId}`,
-    ...requestOptions
+    q: `(orgid: ${orgId}`,
+    filter: `title:"${title}"`,
```
mjuniper, line 126: This is the key change.

mjuniper (approved):

tomwayson (approved):

PR #796, opened 2022-05-20
feat(hub-common): add project family to compose flow

sonofflynn89:
affects: @esri/hub-common

### DIFF
--- a/packages/common/src/content/compose.ts
+++ b/packages/common/src/content/compose.ts
@@ -212,7 +212,7 @@ const getMetadataUpdatedDateInfo = (item: IItem, metadata?: any) => {
 
 // public API
 /**
- * DEPRECATED: Compute the content type icon based on the content type
+ * Compute the content type calcite-icon based on the content type
  * @param content type
  * @returns content type icon
  */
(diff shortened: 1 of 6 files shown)


packages/common/src/content/compose.ts
```diff
@@ -212,7 +212,7 @@ const getMetadataUpdatedDateInfo = (item: IItem, metadata?: any) => {
 
 // public API
 /**
- * DEPRECATED: Compute the content type icon based on the content type
+ * Compute the content type calcite-icon based on the content type
```
sonofflynn89, line 215: @tomwayson do we want to un-deprecate this now?

packages/common/src/content/compose.ts
```diff
@@ -255,6 +255,7 @@ export const getContentTypeIcon = (contentType: string) => {
     hubInitiative: "initiative",
     hubInitiativeTemplate: "initiative-template",
     hubPage: "browser",
+    hubProject: "briefcase",
```
sonofflynn89, line 258: we won't be using this in opendata-ui, but since we're in here I thought I'd add it for hub-components

tomwayson (approved):
Perfect!

PR #799, opened 2022-05-24
feat(hub-common): createItemFromUrlOrFile rtrns itm title, access chn…

benstoltz:
…ge resp, and item chnge resp

1. Description: Change up the response object for createItemFromUrlOrFile to show whether or not access level / sharing to groups succeeded.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/items/create-item-from-url-or-file.ts
+++ b/packages/common/src/items/create-item-from-url-or-file.ts
@@ -1,6 +1,7 @@
 import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
 import {
   ICreateItemResponse,
+  ISharingResponse,
   setItemAccess,
   shareItemWithGroup,
 } from "@esri/arcgis-rest-portal";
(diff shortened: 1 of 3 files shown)


packages/common/src/items/create-item-from-url-or-file.ts
```diff
@@ -22,16 +23,28 @@ interface ICreateItemFromUrlOrFileOptions extends IUserRequestOptions {
...
  */
 export async function createItemFromUrlOrFile(
   createItemFromUrlOrFileOptions: ICreateItemFromUrlOrFileOptions
-): Promise<ICreateItemResponse> {
+): Promise<{
+  title: string;
+  createdItem: ICreateItemResponse;
+  itemAccessResponse: ISharingResponse;
+  itemSharingResponse: ISharingResponse[];
+}> {
```
dbouwman, line 40: Returning all the responses is a good idea. Also I'm pretty sure tsdoc will generate the return details

packages/common/src/items/create-item-from-url-or-file.ts
```diff
@@ -74,5 +87,10 @@ export async function createItemFromUrlOrFile(
     );
   }
 
-  return createdItem;
+  return {
```
dbouwman, line 90: I know it's not part of this PR, but do you have some `try/catch` stuff inside the `createItemFrom...` fns? I see a `failSafe(...)` on the share call, but it looks like the others could fail and that could cascade poorly

dbouwman: Like I don't see any `Promise.reject(...)` in the `create-item-...` test below.

dbouwman (approved):
This LGTM,  but give some thought to the comments

packages/common/src/items/create-item-from-url-or-file.ts
benstoltz: Sub funcs do have try/catch.

PR #807, opened 2022-06-14
feat(hub-common): add metadata.contact as a possible source for IHubContent's publisher field and add isExternal flag

sonofflynn89:
affects: @esri/hub-common

1. Description:
Missed AC from https://github.com/ArcGIS/opendata-ui/pull/10731
Also fixes issue where I mixed up metadata.contact with metadata.resource.contact


tomwayson (approved):
This LGTM, but I wonder if we should add the public/external logic at this time?

PR #810, opened 2022-06-21
perf(hub-common): fetchContent returns cached recordCount when it can

tomwayson:
affects: @esri/hub-common

1. Description:

Always fetching the live record count is proving very costly for some feature servers. In the long run, we will remove this from `fetchContent()` so that the fetch can be deferred and not block page load. For now, `fetchContent()` will still return a `recordCount`, but we'll rely on the cached value from the Hub API when it's available, which should be for any non-enterprise public layer.

1. Instructions for testing:

1. Closes Issues:  addresses [3976](https://devtopia.esri.com/dc/hub/issues/3976)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61386,7 +61386,7 @@
 		},
 		"packages/common": {
 			"name": "@esri/hub-common",
-			"version": "9.43.0",
+			"version": "9.43.1",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"abab": "^2.0.5",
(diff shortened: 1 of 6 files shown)


ajturner:
> For now, `fetchContent()` will still return a `recordCount`, but we'll rely on the cached value from the Hub API when it's available, which should be for any non-enterprise public layer.

@tomwayson how will this affect the display of feature count in the various areas of the Dataset metadata view? 
We don't want that number to be out of date - so I'm wondering if this change needs to be paired up with the change for those components to async load the live feature count

tomwayson:
> so I'm wondering if this change needs to be paired up with the change for those components to async load the live feature count

exactly, that is the plan. We will not show cached recordCount - we'll re-fetch it out of band.

mjuniper (approved):
woot woot!

PR #811, opened 2022-06-23
fix(hub-common): enrichOrg was fetching the org of the the authenticated user (not the item) in some instances

sonofflynn89:
affects: @esri/hub-common

1. Description:

To prevent the org of the authenticated user from being fetched as part of `enrichOrg`, we simply do a no-op if `orgId` is not available on the item or on the ownerUser.


tomwayson (approved):
good catch

PR #812, opened 2022-06-24
feat: hubSearch IQuery Implementation

dbouwman:
1. Description:
- add IHubCatalog/IHubCollection which will eventually replace ICatalog/ICollection
- add IQuery. IFilter, IPredicate
- add utils to serialize IQuery into search options for Portal
- adds `hubSearchQuery` which uses IQuery
- adds all needed backing functions for IQuery 
- ensure existing FilterGroup functions continue to operate

This is the first of a series of PRs. When this merges, we can start moving Hub-Components to using the `IQuery` based functions. Once that is completed, we can then remove the `FilterGroup` based functions, rename `hubSearchQuery` to `hubSearch`

1. Instructions for testing:

- run tests

### DIFF
--- a/packages/common/src/search/Collection.ts
+++ b/packages/common/src/search/Collection.ts
@@ -1,0 +1,92 @@
+import { hubSearchQuery, IArcGISContext } from "..";
+import {
+  IHubCatalog,
+  IHubCollection,
+  IHubSearchOptions,
+  IHubSearchResponse,
+  IHubSearchResult,
+  IQuery,
+} from "./types";
+
+/**
+ * Implements Collection behavior
+ */
+/* istanbul ignore next */
+export class Collection implements IHubCollection {
+  private _context: IArcGISContext;
+  private _collection: IHubCollection;
+  private constructor(collection: IHubCollection, context?: IArcGISContext) {
+    this._collection = collection;
+    if (context) {
+      this._context = context;
+    }
+  }
+
+  public static async create(
+    collection: IHubCollection,
+    context?: IArcGISContext
+  ): Promise<Collection> {
+    // Async so we could extend to look up by ref like
+    // https://mysite.com#documents vs passing in a full ICollection
+    const col = new Collection(collection, context);
+    return Promise.resolve(col);
+  }
+
+  // Getters
+
+  public get label(): string {
+    return this._collection.label;
+  }
+  public get key(): string {
+    return this._collection.key;
+  }
+  public get include(): string[] {
+    return this._collection.include;
+  }
+  public get scope(): IQuery {
+    return this._collection.scope;
+  }
+  public get sortField(): string {
+    return this._collection.sortField || "title";
+  }
+  public get sortDirection(): "asc" | "desc" {
+    return this._collection.sortDirection || "asc";
+  }
+
+  public async search(
+    query: string | IQuery,
+    options: IHubSearchOptions = {}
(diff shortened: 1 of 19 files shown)


packages/common/src/search/Collection.ts
```diff
@@ -0,0 +1,92 @@
...
+      this._context = context;
+    }
+  }
+
+  public static async create(
+    collection: IHubCollection,
+    context?: IArcGISContext
+  ): Promise<Collection> {
+    // Async so we could extend to look up by ref like
+    // https://mysite.com#documents vs passing in a full ICollection
```
tomwayson, line 30: I get this, but I wonder how much better: 

`const collection = await Collection.create(url)`

is than: 

`const collection = Collection.create(await fetchCollection(url))`

The latter ensures that the fetching logic is tree-shakable.

packages/common/src/search/Collection.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    // merge the passed in query w/ the scope
+    qry.filters = [...qry.filters, ...this.scope.filters];
+
+    options.requestOptions = this._context.hubRequestOptions;
+    // inject default sort info if not specified
+    options.sortField = options.sortField || this.sortField;
+    options.sortOrder = options.sortOrder || this.sortDirection;
+    // inject default includes if not specified
+    options.include = options.include || this.include;
```
tomwayson, line 87: we should build a new options hash out of the default values from `this` and whatever was passed in rather than mutating the `options` that was passed in.

packages/common/src/search/Collection.ts
```diff
@@ -0,0 +1,92 @@
...
+            ],
+          },
+        ],
+      };
+    } else {
+      qry = query;
+    }
+
+    // merge the passed in query w/ the scope
+    qry.filters = [...qry.filters, ...this.scope.filters];
```
tomwayson, line 80: I would have expected some reaching up into the collection's catalog to merge in it's scopes as well. 😕

packages/common/src/search/_internal/portalSearchGroups.ts
```diff
@@ -1,24 +1,28 @@
 import { ISearchOptions, searchGroups } from "@esri/arcgis-rest-portal";
 import { IGroup } from "@esri/arcgis-rest-types";
+import { HubError } from "../..";
```
tomwayson, line 3: I know this was auto added, but import statements should not end in a `.` in order to avoid circular refs.

packages/common/src/search/index.ts
```diff
@@ -10,4 +10,5 @@ export * from "./types";
 export * from "./collectionState";
 export * from "./convertCatalog";
 export * from "./filter-utils";
+export * from "./ifilter-utils";
```
tomwayson, line 13: do these need to be exported to external users of the library?

packages/common/src/search/Collection.ts
dbouwman: This is definitely a puzzler... b/c many times a collection will exist w/o a Catalog...

packages/common/src/search/index.ts
dbouwman: I'll check on this - I think there are a few fn's that may make sense, but I'll see if I can avoid it

packages/common/src/search/Collection.ts
dbouwman: This class is not exported, so we can revisit and revise as we move forward

PR #819, opened 2022-06-29
fix: remove filtergroups search implementation

dbouwman:
1. Description:

Remove the now un-used FilterGroups implementation, and the related functions

1. Instructions for testing:

Run tests

1. Closes Issues: [#4307](https://devtopia.esri.com/dc/hub/issues/4307)


packages/common/src/search/hubSearch.ts
```diff
@@ -15,77 +13,27 @@ import {
...
-import {
   portalSearchUsers,
-  portalSearchUsersFilterGroups,
-} from "./_internal/portalSearchUsers";
+} from "./_internal";
 
 /**
- * Main Search function for ArcGIS Hub
+ * @private
+ * TEMPORARY - this function only exists to allow incremental
```
tomwayson, line 21: would be good to change this to "DEPRECATED" next time you are in this code, b/c that's how I plan to search for things to cull in the Great Breaking Change to come

tomwayson (approved):
Thanks @dbouwman 

If I understand, this is only deleting internal methods, correct? If so let'r rip

PR #828, opened 2022-07-08
feat: add Catalog and Collection Classes

dbouwman:
1. Description:

To simplify the use of Collections and Catalogs, I've created classes to wrap up some of the complexities.

Now with tests!

## Examples
```js  
// From a IHubCatalog structure
const instance = Catalog.fromJson(catalog, context);
// fetch for a site
const i2 = await Catalog.init("https://opendata.dc.gov", context);
// fetch for an item
const i3 = await Catalog.init("3ef", context);

// search catalog for items
const r1 = await instance.searchItems("schools");
// search catalog for groups
const r2 = await instance.searchGroups("schools");

// get a Collection from a catalog
const coll = instance.getCollection("environment");

// search collection
const r3 = await coll.search("Pine St", {include: ["owner"]});

```
1. Instructions for testing:

Run tests


packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+        ],
+      } as IQuery;
+    }
+    // Now merge in catalog scope level filters
+    if (this._catalog.scopes[targetEntity]) {
+      query.filters = [
+        ...query.filters,
+        ...this._catalog.scopes[targetEntity].filters,
+      ];
+    }
```
tomwayson, line 189: You are mutating the `query` arg here. Better to do something like `const _query = (typeof query === "string") ? { targetEntity, filters: [...] } : cloneObject(query)`

packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    // An instance always uses the context so we remove any
+    delete options.authentication;
+    options.requestOptions = this._context.hubRequestOptions;
```
tomwayson, line 193: ditto here, mutating the `options` arg.

also, if we always use the instance's context, why take in an `options` argument?

packages/common/src/search/types/IHubSearchOptions.ts
```diff
@@ -1,5 +1,5 @@
 import { UserSession } from "@esri/arcgis-rest-auth";
-import { IArcGISContext, IHubRequestOptions, IModel } from "../..";
+import { EntityType, IArcGISContext, IHubRequestOptions, IModel } from "../..";
```
tomwayson, line 2: imports that end in `.` result in circular dependencies

packages/common/src/search/types/IHubSearchOptions.ts
```diff
@@ -47,6 +47,11 @@ export interface IHubSearchOptions {
    * DEPRECATED: Use `include` instead
    */
   enrichments?: Enrichments[];
+
+  /**
+   * Specify what entity to search for; For use with hubSearch
+   */
+  targetEntity?: EntityType;
```
tomwayson, line 54: why is this part of options and query?

packages/common/src/search/upgradeCatalogSchema.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const CATALOG_SCHEMA_VERSION = 1.0;
+
+/**
+ * Apply schema upgrades to Catalog objects
+ * @param catalog
+ * @returns
+ */
+export function upgradeCatalogSchema(catalog: any): IHubCatalog {
+  if (getProp(catalog, "schemaVersion") === CATALOG_SCHEMA_VERSION) {
```
tomwayson, line 12: as mentioned above, this should return a clone

packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+  }
+
+  /**
+   * Execute a search against the Catalog as a whole
+   * @param query
+   * @param targetEntity
+   * @returns
+   */
+  search(
```
tomwayson, line 164: I wonder if instead of `i.search(query, { targetEntity: "groups" })` we should just have `i.searchItems(query)` and `i.searchGroups(query)`?

tomwayson (commented):
Nice work @dbouwman 

A few suggestions and questions

packages/common/src/search/types/IHubSearchOptions.ts
dbouwman: If you pass in a string to `search(...)` we need a means to specify what entity we're searching against. It is duplicative if  you pass in an `IQuery` which is why it's optional

packages/common/src/search/Catalog.ts
dbouwman: The intent here is that we should be able to use this single fn to search for any of those entities. That said, needing to pass in the entityType in the searchOptions seems confusing.

At one point AT had wanted a catalog search to execute a query for each entity that had a scope, so you'd get an array if `IHubSearchResponse<IHubSearchResult>` but in later discussions re: the http api, I understood that it would specify the entity type as a url segment, and thus return a single `IHubSearchResponse<IHubSearchResult>`.

Anyhow - I'm open to changing this, but that's the back story

packages/common/src/search/types/IHubSearchOptions.ts
dbouwman: fixed

packages/common/src/search/Catalog.ts
dbouwman: the options holds paging, aggregation, sort etc etc info. Added clone

packages/common/src/search/Catalog.ts
dbouwman: fixed

packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+    return mapBy("key", this.collections);
+  }
+
+  /**
+   * Get a Collection instance by name
+   * @param name
+   * @returns
+   */
+  getCollection(name: string): Collection {
+    const json = this._catalog.collections.find((entry) => entry.key === name);
```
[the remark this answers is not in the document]
dbouwman: fixed

packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Create a Catalog instance from a Catalog Json object
+   * @param json
+   * @param context
+   * @returns
+   */
+  static fromJson(json: any, context: IArcGISContext): Catalog {
+    // ensure it's in the latest structure
+    json = upgradeCatalogSchema(json);
+    return new Catalog(cloneObject(json), context);
```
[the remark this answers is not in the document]
dbouwman: fixed

packages/common/src/search/Catalog.ts
```diff
@@ -0,0 +1,198 @@
...
+   * Create a Collection instance from a Catalog json object, a site url or itemId
+   * '''js
+   * const catalog = await Catalog.create('https://site-org.hub.arcgis.com', context);
+   * '''
+   *
+   * @param identifier
+   * @param context
+   * @returns
+   */
+  public static async init(
```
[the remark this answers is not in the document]
dbouwman: I was just being consistent with ArcGISContextManager, but I don't have strong feels either way - let's get input from @drspacemanphd as well

docs/src/guides/concepts/catalog-collection.md
```diff
@@ -8,91 +8,246 @@ group: 2-concepts
...
+When executing queries, additional filters can be applied at the Catalog or Collection level.
 
-Catalogs were designed to be associated with any item, but currently Hub Sites utilize a catalog to define what content is available on a site search. Hub Projects and Hub Initiatives will also have Catalogs.
+While a Catalog can reference any set of platform entities, any individual Collection can only reference a single platform entity type (i.e. items vs groups)
 
-All the Hub related items will define the catalog in the data.json, in the `.catalog` property. The exact structure of this configuration will change over time so it's important to either fetch the Site/Project/Initiative via hub.js functions, or use the `convertCatalog(..)` function to ensure you are working with the latest structure.
+Hub Sites utilize a Catalog to define what content is available on a site search. Catalogs are also planned for Hub Projects and Hub Initiatives.
+
+Any hiearchy between Sites, Initiatives and Projects must be manually maintained, and will not be automatic.
```
drspacemanphd, line 19: small misspelling

docs/src/guides/concepts/catalog-collection.md
```diff
@@ -8,91 +8,246 @@ group: 2-concepts
...
+  collections: [
     {
-      "targetEntity": "item",
-      "key": "webmaps",
-      "label": "Web Maps",
-      "scope": {
-        "targetEntity": "item",
-        "filters": [
+      targetEntity: "item",
+      key: "webmaps",
```
drspacemanphd, line 103: See my question about collection keys in the interface definition.

packages/common/e2e/helpers/Artifactory.ts
```diff
@@ -98,7 +101,7 @@ export default class Artifactory {
     return this.getOrg(name).fixtures as Record<string, unknown>;
   }
 
-  getFixtureId(org: string, path: string): string {
-    return getProp(this.getOrg(org).fixtures, path);
-  }
+  // getFixtureId(org: string, path: string): string {
```
drspacemanphd, line 104: Do we plan on re-implementing this? Just want to avoid lots of stray comments

packages/common/e2e/helpers/config.ts
```diff
@@ -72,21 +72,16 @@ const config = {
             },
           },
         },
-        // portal: {
```
drspacemanphd, line 75: Keep comments?

packages/common/src/search/Catalog.ts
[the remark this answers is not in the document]
drspacemanphd: I'm ok with either tbh

packages/common/src/search/Catalog.ts
drspacemanphd: This is a fine implementation now but will definitely get more complex with general searching and "common" filters. For example, if one provides an IQuery with a single `title` filter, and the catalog's scope contains multiple targetEntity types (e.g. `item` and `group`) it's not unreasonable to assume it would be able to search across those multiple scopes simultaneously. The complexity is then determining which filters are "allowed" for each given  type.

packages/common/src/search/Collection.ts
```diff
@@ -1,36 +1,68 @@
...
 
-  public static async create(
+  /**
+   * TBD if we implement this
+   * Fetch a collection from a catalog stored with an item
+   * @param collection
+   * @param context
+   * @returns
+   */
+  // public static async init(
```
drspacemanphd, line 33: It might just be simpler to simply fetch the catalog of the resource (as we do in Catalog) and then filter by identifier.

packages/common/src/search/Collection.ts
```diff
@@ -1,36 +1,68 @@
...
 /**
- * Implements Collection behavior
+ * Collection Class
+ *
+ * Abstracts searching a Collection
+ *
+ * For more information, check out the [Catalog & Collection Guide](/hub.js/guides/concepts/catalog-collection/)
  */
-/* istanbul ignore next */
 export class Collection implements IHubCollection {
```
drspacemanphd, line 18: Are we intending the `key` property to be an identifier unique among the collections of a resource's catalog? If so we might want to include that logic when trying to create a new Collection

packages/common/src/search/upgradeCatalogSchema.ts
```diff
@@ -0,0 +1,71 @@
...
+import { IHubCatalog } from "./types";
+
+const CATALOG_SCHEMA_VERSION = 1.0;
+
+/**
+ * Apply schema upgrades to Catalog objects
+ * @param catalog
+ * @returns
+ */
+export function upgradeCatalogSchema(catalog: any): IHubCatalog {
```
drspacemanphd, line 12: Nice idea. Is the plan that we always invoke this as part of the workflow to save a catalog? Or is it a wrapper for when we simply retrieve a catalog and need it to be structured according the latest schema?

drspacemanphd (approved):

packages/common/src/search/Catalog.ts
dbouwman: @tomwayson  & @drspacemanphd - I'm now leaning towards having separate functions, as that simplifies the API - want to search for Groups? use `myCatalog.searchGroups(...)`. 

Another related question is how to handle catalog level queries for entities that don't have a scope defined. i.e. if a catalog only defines `item` scope, and we search for `group` - what should happen? I'm leaning towards an exception.

As for searching across the entity types, I think that leads to even more complexity (how to handle paging, aggregations etc etc) so my inclination is to say that's just not supported. While it would be "neat" to issue a single query and get back users, groups, events and items related to a single search term, I don't think that's a critical feature, and any UI that wants to show that information could manage it via multiple search calls.

packages/common/e2e/helpers/Artifactory.ts
dbouwman: This is just test harness code - I can remove, but it's non-shipping

packages/common/src/search/Collection.ts
dbouwman: Within the context of a catalog, the key should be unique. But, if you're just creating a collection to pass into a Gallery, separate from a Catalog, the prop  just needs to be present. Agreed that when we add a `addCollection(...)` fn to the Catalog we need to implement this check, but since Collections can exist separate from Catalogs, we can't put that logic into the Collection class itself

packages/common/src/search/upgradeCatalogSchema.ts
dbouwman: The latter. The way we currently handle "migrations" for Sites, Initiatives and a few other things in the system is to always run things through these functions "at read". Then the rest of the system can reliably work with the latest schema version structure. If the object happens to be saved, then it's saved with the latest version.

packages/common/src/search/Collection.ts
dbouwman: Agreed - the tricky bit is more in how to specify the collection key, for Enterprise Site urls, which already have a hash in them. I tried to explore this with some of the Enterprise test servers, but apparently they all have CORS locked down tight so I can't really test permutations. Until I find a server w open CORS, I figured I'd just comment this out. 

Also - since most collections are actually generated dynamically and passed into galleries, it's not super high-value at this point

dbouwman:
Next Steps from discussion with @ajturner 
- implement 
   - `Catalog:searchCollections(...):Promise<CollectionResultHash>` 
   - `Catalog:searchScopes(...):Promise<ScopeResultHash>`
   - leave existing `search{Entity}(...)` functions
   - keep `search(...)` internal

PR #829, opened 2022-07-13
feat: change entity search to use IQuery

dbouwman:
1. Description:

Change the entity search functions (`searchProjects, searchSites, HubProjectManager.search, HubSiteManager.search`) to take `IQuery` instead if `Filter<type>`

This also marks a lot of `/_internal` fns as `@private` b/c the doc generator does not check to see if fn's are exported from the package - just that the fn / type is "exported" at all.

None of these fns are used in hub-components or opendata-ui - so while techincally breaking change, it will not impact anything.

1. Instructions for testing:

- run tests

1. Closes Issues: [4392](https://devtopia.esri.com/dc/hub/issues/4392)


packages/common/e2e/entity-search.e2e.ts
```diff
@@ -18,51 +20,50 @@ describe("Entity Search", () => {
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 200000;
   });
 
-  it("validate createEntitySearchFn", async () => {
+  fit("validate createEntitySearchFn", async () => {
```
tomwayson, line 23:
```suggestion
  it("validate createEntitySearchFn", async () => {
```

packages/common/src/projects/HubProjects.ts
```diff
@@ -246,27 +249,38 @@ export async function destroyProject(
...
  * Search for Projects, and get IHubProject results
  *
- * Different from `searchContent` in that this returns the specific entity type
+ * Different from `hubSearch` in that this returns the IHubProjects
  *
  * @param filter
  * @param options
  * @returns
  */
 export async function searchProjects(
```
tomwayson, line 258: in the other managers, his method name is just `search()`, right? If so we could (and should) introduce that here, and we could leave and deprecate `searchProjects()` - one less breaking change.

packages/common/e2e/hub-sites.e2e.ts
```diff
@@ -26,7 +26,7 @@ describe("Hub Sites", () => {
       // inspect and ensure we have what we expect
       expect(site.typeKeywords).toBeDefined();
       // get the slug keyword
-      const slug = site.typeKeywords.find((e) => e.indexOf("slug|") === 0);
+      const slug = site.typeKeywords?.find((e) => e.indexOf("slug|") === 0);
```
drspacemanphd, line 29: Does this automatically coerce to `[]`?

packages/common/src/sites/HubSites.ts
```diff
@@ -489,18 +489,30 @@ export async function convertItemToSite(
...
-  };
-  // merge filters
-  const sitesFilter = mergeContentFilter([scopingFilter, filter]);
-  // delegate
-  return searchContentEntities(sitesFilter, convertItemToSite, options);
+): Promise<IHubSearchResponse<IHubSite>> {
+  let qry: IQuery;
+
+  if (typeof query === "string") {
+    qry = createQueryFromString(query, "term", "item");
```
drspacemanphd, line 498: Do we need the `term` argument here? As in, are we expecting any other filter to be applied as a string other than a general `term`?

packages/common/test/projects/HubProjects.test.ts
```diff
@@ -71,18 +72,18 @@ const PROJECT_ITEM_ENRICH: portalModule.IItem = {
   ],
   description: "Test Project with data",
   tags: ["hubproject"],
-  snippet: null,
+  // snippet: null,
```
drspacemanphd, line 75: why are these commented out?

tomwayson (approved):
one small change and a suggestion

drspacemanphd (approved):
Good work, I think the builds nicely in that we are trying to generalize search a bit more.

packages/common/src/projects/HubProjects.ts
dbouwman: In the Manager's the function is `search` b/c that's scoped into the Manager class... the backing module that just exports fn's needs to qualify the fn name - thus `searchProject`

tomwayson: oops - I thought that was still part of the class, carry on.

packages/common/src/sites/HubSites.ts
dbouwman: When searching for users, the generic thing may be `name`. I also considering enabling a query like `tag: water` which we could split and the use this same fn - or sniff for guid's and then use `id`. Viv built something like that in the early versions of the gallery and it was handy

packages/common/test/sites/HubSites.test.ts
```diff
@@ -106,14 +107,14 @@ const SITE_ITEM_ENRICH: portalModule.IItem = {
     "Web Map",
     "Registered App",
   ],
-  description: null,
+  // description: null,
```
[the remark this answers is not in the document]
dbouwman: see above

PR #845, opened 2022-08-02
fix: predicates need to be wrapped in parens

mjuniper:
1. Description: Fixes an issue with search. What I was seeing was that when I sent in a query that contained a filter like:

```js
{
  operation: "OR",
  predicates: ['$app', '$webmap']
};
```
The well known type filters would get combined in such a way that they could negate each other or otherwise produce incorrect results. What needed to happen was for each predicate to be wrapped in parens so they would be handled as a block when they were combined with others with an `AND' or an 'OR'.

1. Instructions for testing: run the tests

### DIFF
--- a/packages/common/test/projects/HubProjects.test.ts
+++ b/packages/common/test/projects/HubProjects.test.ts
@@ -359,7 +359,7 @@ describe("HubProjects:", () => {
       // verify the query
       const searchOpts = searchSpy.calls.argsFor(0)[0];
 
-      expect(searchOpts.q).toBe(`type:"Hub Project" AND water`);
+      expect(searchOpts.q).toBe(`(type:"Hub Project") AND (water)`);
     });
     it("accepts an IQuery", async () => {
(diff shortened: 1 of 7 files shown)


packages/common/test/projects/HubProjects.test.ts
```diff
@@ -359,7 +359,7 @@ describe("HubProjects:", () => {
       // verify the query
       const searchOpts = searchSpy.calls.argsFor(0)[0];
 
-      expect(searchOpts.q).toBe(`type:"Hub Project" AND water`);
+      expect(searchOpts.q).toBe(`(type:"Hub Project") AND (water)`);
```
mjuniper, line 362: This looks kinda weird but I did some manual tests agains the portal api and verified that it handles it ok with the term (water) wrapped in parens.

dbouwman (approved):
glad this seems to fix what would have been really gnar otherwise!

PR #848, opened 2022-08-03
fix: re-export survey utils

dbouwman:
1. Description:

Some exports were mistakenly removed > year ago, but we never noticed b/c opendata-ui was not picking up later versions of `@esri/hub-surveys`. Now that we are doing a breaking chance and bumping all packages, this became a problem

1. Instructions for testing:

- run tests


tomwayson (approved):
:shipit:

PR #858, opened 2022-08-08
feat(hub-discussions): adding acl support for channels

jeromesteve202:
1. Description:
Adding support for acl when it comes to creating channels

1. Instructions for testing:
Create a channel with ACL paramater

https://devtopia.esri.com/dc/hub/issues/4526

### DIFF
--- a/packages/discussions/src/types.ts
+++ b/packages/discussions/src/types.ts
@@ -738,9 +738,9 @@ export interface IPermission {
  *
  * @export
  * @interface IWithACL
- * @extends {IWithSettings}
+ * @extends {ICreateChannelWithACL}
  */
-export interface IWithACL extends IWithSettings {
+export interface ICreateChannelWithACL extends IWithSettings {
   anonymous?: IPermission;
   authenticated?: IPermission;
   groups?: {
@@ -762,7 +762,7 @@ export interface IWithACL extends IWithSettings {
  * @extends {IHubRequestOptions}
  */
 export interface ICreateChannelOptions extends IHubRequestOptions {
-  params: ICreateChannel | IWithACL;
+  params: ICreateChannel | ICreateChannelWithACL;
 }
 
 /**
--- a/packages/discussions/src/types.ts
+++ b/packages/discussions/src/types.ts
@@ -706,7 +706,7 @@ export interface ISearchChannelsOptions extends IHubRequestOptions {
 }
 
 /**
- * permissino object that will populate ACL interface
+ * permission object that will populate ACL interface
  *
  * @export
  * @interface IPermission
--- a/packages/discussions/src/types.ts
+++ b/packages/discussions/src/types.ts
@@ -706,13 +706,28 @@ export interface ISearchChannelsOptions extends IHubRequestOptions {
 }
 
 /**
+ * Role types
+ *
+ * @export
+ * @enum {string}
+ */
+export enum Role {
+  READ = "read",
+  WRITE = "write",
+  READWRITE = "readWrite",
+  MODERATE = "moderate",
+  MANAGE = "manage",
+  OWNER = "owner",
+}
+
+/**
  * permission object that will populate ACL interface
  *
  * @export
  * @interface IPermission
  */
 export interface IPermission {
-  role: string;
+  role: Role;
   createdAt: string;
   updatedAt: string;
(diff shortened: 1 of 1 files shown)


packages/discussions/src/types.ts
```diff
@@ -705,6 +705,40 @@ export interface ISearchChannelsOptions extends IHubRequestOptions {
...
 }
 
+/**
+ * permission object that will populate ACL interface
+ *
+ * @export
+ * @interface IPermission
+ */
+export interface IPermission {
+  role: string;
```
rgwozdz, line 715: Should the role just be a primitive string?  What about a distinct set of string values or enums?

jeromesteve202: Yea, good idea to strictly constrain that.

packages/discussions/src/types.ts
```diff
@@ -705,6 +705,55 @@ export interface ISearchChannelsOptions extends IHubRequestOptions {
...
+}
+
+/**
+ * request options for creating channel with ACL
+ *
+ * @export
+ * @interface IWithACL
+ * @extends {IWithSettings}
+ */
+export interface IWithACL extends IWithSettings {
```
rgwozdz, line 743: Would the name `ICreateChannelWithACL` be a more accurate? i.e., it extends IWithSettings, which looks specific to Channels, so if this interface is only used for channels, let's include that in the name.

jeromesteve202: Yea, I agree

rgwozdz (approved):

packages/discussions/src/types.ts
```diff
@@ -705,6 +705,55 @@ export interface ISearchChannelsOptions extends IHubRequestOptions {
...
+
+/**
+ * request options for creating channel with ACL
+ *
+ * @export
+ * @interface IWithACL
+ * @extends {ICreateChannelWithACL}
+ */
+export interface ICreateChannelWithACL extends IWithSettings {
+  anonymous?: IPermission;
```
rweber-esri, line 744: Is there a scenario where the client would call `createChannel` and provide `createdAt` and `updatedAt` values for the `authentication`, `groups`, `orgs`, and `users` permission? Seems like that's something that would be included in the output, i.e. `IChannel`, but not necessarily `ICreateChannelWithACL`. If that's true, we could use the [omit utility](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys) to define another type, e.g.

```
export interface IPermissionOptions extends Omit<IPermission, "createdAt", "updatedAt">

export interface ICreateChannelWithACL extends IWithSettings {
  anonymous?: IPermissionOptions;
  // ...
}
```

I believe we also need to make sure that `IChannel` reflects the channel schema returned when calling the `getById` and `search` channel endpoints. When I pull the release for this PR merge, accessing `channel.acl` results in the typescript error `Property 'acl' does not exist on type 'IChannel'.`

jeromesteve202: @rweber-esri So a couple of things here:

1. The scenario at the moment does involve the client to provide `createdAt` and `updatedAt` for those entities that you mentioned. In the Discussions API, I believe we need to add a way for those fields to be populated with a date/time of current value in some way rather than making the client enter that in. 

2. I see your concern with the `IChannel` needing to reflect the schema of a channel. I believe it is related to this interface
```
export interface IChannel
  extends IWithSettings,
    IWithSharing,
    IWithAuthor,
    IWithEditor,
    IWithTimestamps {
  id: string;
}
```
I didn't change this during my PR and I see now that `acl` is not covered here. I wonder if I need to add an acl object in here like the `id` but make it the type of the interface `ICreateChannelWithACL`

rgwozdz: > 1. The scenario at the moment does involve the client to provide `createdAt` and `updatedAt` for those entities that you mentioned. In the Discussions API, I believe we need to add a way for those fields to be populated with a date/time of current value in some way rather than making the client enter that in.

ha, that's really remarkable...

rgwozdz: I said this before, but a lot of the confusion here is that we are trying to use the same interfaces for requesting a channel to be created _and_ the shape of a channel that gets returned.  They share some common properties, but they are different things.

PR #860, opened 2022-08-15
feat(hub-common): introduces a getFamilyTypes function

drspacemanphd:
affects: @esri/hub-common

1. Description:
[4533](https://devtopia.esri.com/dc/hub/issues/4533). Hub Site OGC APIs currently can return (and search using) user-defined collections that are scoped to the site's catalog. However, Hub also currently has "default" collections on the search route, with each collection as a bucket of different types. The story represents the effort to introduce these collections as default collections that are automatically returned from Site OGC APIs.

While the terminology of this story refers to collections (which are defined and recognized resources within the OGC spec), the _actual_ collections we plan to include are MUCH more akin to non-configurable Hub Families than configurable collections. They are more aligned with resource types that will have their own defined API (and UI) resource path.

In order to get the types associated with a given Hub Family, represented as an OGC collection, this PR introduces a function that returns the types associated with a given family. It is the converse of `getFamily`, and takes the same overrides into account.

2. Instructions for testing:
Unit tests have been included

3. Closes Issues: #<number> (if appropriate)

4. [x] used semantic commit messages
  
5. [x] PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

### DIFF
--- a/packages/common/src/collections.ts
+++ b/packages/common/src/collections.ts
@@ -142,6 +142,20 @@ export const getCollection = (type?: string) => {
   });
 };
 
+/**
+ * The converse of getCollection, returns associated types of provided collection
+ * @param collection The Hub collection
+ * @returns An array of types or undefined if collection is not found
+ * @private
+ */
+export const getCollectionTypes = (collection?: string) => {
+  if (!collection) {
+    return;
+  }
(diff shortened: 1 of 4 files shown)


tomwayson (approved):
Great work @drspacemanphd!

PR #863, opened 2022-08-24
fix: rename permission, catalog props and simplify adding permission

dbouwman:
1. Description:

- since we are not exposing the underlying entity props on the classes, we can store the catalog and permissions as `.catalog` and `.permissions`
- remove the redundant `HubPermission` param on `addPermission` method

1. Instructions for testing:

- run tests


juliannemarik (approved):
These look like nice simplifications to me 👍

PR #876, opened 2022-09-09
fix(hub-common): getFamilyTypes() now correctly returns "Hub Initiative Template" as a type of the "template" family

sonofflynn89:
affects: @esri/hub-common

1. Description:
Discovered while working on https://devtopia.esri.com/dc/hub/issues/3904
Just returns a missing type for the `template` family


tomwayson (approved):
good catch @sonofflynn89

PR #879, opened 2022-09-13
feat: add editor config static method to HubProject class

dbouwman:
1. Description:

Expose `.getEditorConfig(...)` as static method on Hub entity classes.

This is a rough first pass implementation to get feedback. Will polish once assumptions have been validated.

Some things to consider:
- adding schemas to hub.js required importing Ajv and json-schema-typed. Since we're just using those for types, I *hope* nothing else gets bundled into the build - @tomwayson should/could we make these devDeps?
- another solution here is to put all this into a util fn in hub-components, and fiddle with adding static methods to the classes. While nominally less ergonomic, I'm not sure it's ZOMG worse, and keeps UI layer concerns in the UI layer.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/behaviors/IWithEditorBehavior.ts
+++ b/packages/common/src/core/behaviors/IWithEditorBehavior.ts
@@ -1,0 +1,88 @@
+import { JSONSchema } from "json-schema-typed";
+import Ajv from "ajv";
+
+export enum UiSchemaRuleEffects {
+  SHOW = "SHOW",
+  HIDE = "HIDE",
+  DISABLE = "DISABLE",
+  NONE = "",
+}
+
+export enum UiSchemaElementTypes {
+  section = "Section",
+  step = "Step",
+  control = "Control",
+  layout = "Layout",
+  slot = "Slot",
+}
+
+export enum UiSchemaSectionTypes {
+  accordion = "accordion",
+  stepper = "stepper",
+}
+export interface IConfigurationSchema extends JSONSchema {
+  type?: "object";
+}
+
+export interface IConfigurationValues {
+  [key: string]: unknown;
+}
+
+export interface IChangeEventDetail {
+  valid: boolean;
+  values?: {
+    [key: string]: any;
+  };
+}
+
+export interface IValidationResult {
+  valid: boolean;
+  errors?: Ajv.ErrorObject[];
+}
+
+export interface IUiSchemaRule {
+  effect: UiSchemaRuleEffects;
+  condition: {
+    scope: string;
+    schema: IConfigurationSchema;
+  };
+}
+
+export interface IUiSchemaElement {
+  type: string;
+  labelKey?: string;
+  label?: string;
+  options?: {
+    [key: string]: any;
+  };
+  scope?: string;
+  rule?: IUiSchemaRule;
+  elements?: IUiSchemaElement[];
+  tooltip?: string;
+}
+export interface IUiSchema extends IUiSchemaElement {
+  elements?: IUiSchemaElement[];
+}
+
+export type EditorConfigType = "complete" | "minimal";
+
+export interface IEditorConfig {
+  schema: IConfigurationSchema;
+  uiSchema: IUiSchema;
+}
+
+/**
+ *
+ */
+export type SchemaElementOptions = Pick<IUiSchemaElement, "scope" | "options">;
+
+/**
+ *
+ */
+export interface IWithEditorBehavior {
+  getEditorConfig(
+    i18nScope: string,
+    type: EditorConfigType,
+    options: SchemaElementOptions[]
+  ): Promise<IEditorConfig>;
+}
--- a/packages/common/src/core/behaviors/IWithEditorBehavior.ts
+++ b/packages/common/src/core/behaviors/IWithEditorBehavior.ts
@@ -1,70 +1,10 @@
-import { JSONSchema } from "json-schema-typed";
-import Ajv from "ajv";
+import {
+  IConfigurationSchema,
+  IUiSchema,
+  UiSchemaElementOptions,
+} from "../schemas";
 
-export enum UiSchemaRuleEffects {
-  SHOW = "SHOW",
-  HIDE = "HIDE",
-  DISABLE = "DISABLE",
-  NONE = "",
-}
-
-export enum UiSchemaElementTypes {
-  section = "Section",
-  step = "Step",
-  control = "Control",
-  layout = "Layout",
-  slot = "Slot",
-}
-
-export enum UiSchemaSectionTypes {
-  accordion = "accordion",
-  stepper = "stepper",
-}
-export interface IConfigurationSchema extends JSONSchema {
-  type?: "object";
-}
-
-export interface IConfigurationValues {
-  [key: string]: unknown;
-}
-
-export interface IChangeEventDetail {
-  valid: boolean;
-  values?: {
-    [key: string]: any;
-  };
-}
-
-export interface IValidationResult {
-  valid: boolean;
-  errors?: Ajv.ErrorObject[];
-}
-
-export interface IUiSchemaRule {
-  effect: UiSchemaRuleEffects;
-  condition: {
-    scope: string;
-    schema: IConfigurationSchema;
-  };
-}
-
-export interface IUiSchemaElement {
-  type: string;
-  labelKey?: string;
-  label?: string;
-  options?: {
-    [key: string]: any;
-  };
-  scope?: string;
-  rule?: IUiSchemaRule;
-  elements?: IUiSchemaElement[];
-  tooltip?: string;
-}
-export interface IUiSchema extends IUiSchemaElement {
-  elements?: IUiSchemaElement[];
-}
-
-export type EditorConfigType = "complete" | "minimal";
+export type EditorConfigType = "create" | "edit";
 
 export interface IEditorConfig {
   schema: IConfigurationSchema;
@@ -74,15 +14,10 @@ export interface IEditorConfig {
 /**
  *
  */
-export type SchemaElementOptions = Pick<IUiSchemaElement, "scope" | "options">;
-
-/**
- *
- */
 export interface IWithEditorBehavior {
   getEditorConfig(
     i18nScope: string,
     type: EditorConfigType,
-    options: SchemaElementOptions[]
+    options: UiSchemaElementOptions[]
   ): Promise<IEditorConfig>;
 }
--- a/packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
+++ b/packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
@@ -25,7 +25,7 @@ export function applyUiSchemaElementOptions(
       // we dont want to apply this to rules
       // Note: If this logic becomes more complex, we may need to
       // write a function that's more specific to the uiSchema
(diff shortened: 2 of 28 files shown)


packages/common/src/core/behaviors/IWithEditorBehavior.ts
```diff
@@ -0,0 +1,88 @@
+import { JSONSchema } from "json-schema-typed";
+import Ajv from "ajv";
+
```
dbouwman, line 3: Basically hoisted all the types/interfaces from hub-components into Hub.js

packages/common/src/core/behaviors/IWithEditorBehavior.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ *
+ */
+export type SchemaElementOptions = Pick<IUiSchemaElement, "scope" | "options">;
+
+/**
+ *
+ */
```
dbouwman, line 81: While we have this interface, we don't actually use it because typescript can't use an interface to define static class methods. :-/

packages/common/src/projects/HubProject.ts
```diff
@@ -126,6 +133,24 @@ export class HubProject
     }
   }
 
+  /**
```
dbouwman, line 136: This is the function to really look at initially. I've left lots of comments and I _think_ I've made some correct assumptions, but pls add comments if I'm wrong. Once we're agreed this is the logic we want, I'll optimize it and hoist out some helper fns.

packages/common/src/projects/HubProjects.ts
```diff
@@ -280,3 +289,68 @@ export async function enrichProjectSearchResult(
 
   return result;
 }
+
+/**
```
dbouwman, line 293: This is the function to really look at initially. I've left lots of comments and I think I've made some correct assumptions, but pls add comments if I'm wrong. Once we're agreed this is the logic we want, I'll optimize it and hoist out some helper fns.

packages/common/src/projects/schemas.ts
```diff
@@ -0,0 +1,73 @@
...
+      // Issue https://devtopia.esri.com/dc/hub/issues/3725
+      subtype: "long-text",
+    },
+  },
+} as unknown as IConfigurationSchema;
+
+/**
+ * Minimal UI Schema for Hub Project
+ */
+export const HubProjectMinimalUiSchema: IUiSchema = {
```
dbouwman, line 30: This is intentionally incorrect - I just wanted something that's smaller so I can validate the function that subset's the json schema to the set of props defined in a ui schema.

tsconfig.json
```diff
@@ -54,8 +54,8 @@
     "types": [
       "node",
       "jasmine"
-    ]                           /* Type declaration files to be included in compilation. */
-    // "allowSyntheticDefaultImports": true  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
+    ],                           /* Type declaration files to be included in compilation. */
+    "allowSyntheticDefaultImports": true  /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */
```
dbouwman, line 58: apparently requires for Ajv. Was also in hub-components

packages/common/src/core/behaviors/IWithEditorBehavior.ts
```diff
@@ -0,0 +1,88 @@
...
+  scope?: string;
+  rule?: IUiSchemaRule;
+  elements?: IUiSchemaElement[];
+  tooltip?: string;
+}
+export interface IUiSchema extends IUiSchemaElement {
+  elements?: IUiSchemaElement[];
+}
+
+export type EditorConfigType = "complete" | "minimal";
```
juliannemarik, line 67: I don't have super strong feelings about this, but just wondering why we decided to go with "minimal" | "complete" vs. the more explicit "create" | "edit" types?

Technically the "create" form will be more "minimal" in that it will have less fields than the complete "edit" form, but I feel like it might also be a bit confusing to just call the "create" form "minimal" because there's also an implicit layout (i.e. stepped) that's associated with it. Not sure if this makes sense, but just a thought.

packages/common/src/projects/HubProjects.ts
```diff
@@ -280,3 +289,68 @@ export async function enrichProjectSearchResult(
...
+    case "minimal":
+      uiSchema = cloneObject(HubProjectMinimalUiSchema);
+      filterSchema = true;
+      break;
+  }
+
+  if (filterSchema) {
+    // TODO: Hoist into a utility function as all other classes will need to do this
+    // filter out the schema elements that are not needed for the requested UI Schema
+    const propsToKeep = uiSchema.elements.reduce((acc, entry) => {
```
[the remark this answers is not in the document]
dbouwman: Thanks! I'll use these in some tests to ensure I'm grabbing the right things!

packages/common/src/core/behaviors/IWithEditorBehavior.ts
dbouwman: Love it. Many times when I'm 🤔  about naming etc, I'll just slap in something and worry about better names later - this is exactly one of those cases. I'll change to `create` and `edit` :)

packages/common/src/initiatives/HubInitiative.ts
```diff
@@ -135,14 +135,14 @@ export class HubInitiative
...
    */
-  static async getEditorConfig(
-    i18nScope: string,
-    type: EditorConfigType,
-    options: SchemaElementOptions[] = []
-  ): Promise<IEditorConfig> {
-    // Delegate to module fn
-    throw new Error("getEditorConfig Not Implemented for Initiatives");
-  }
+  // static async getEditorConfig(
```
dbouwman, line 138: Commented out for now... once we're happy w/ how this works for projects, we can author the schemas for initiative

packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Apply a set of run-time configurations into the UI schema so
+ * the controls can be configured for the current environment.
+ * TODO: Add Example
+ * @param schema
+ * @param options
+ * @returns
+ */
+export function applyUiSchemaElementOptions(
```
juliannemarik, line 13: I found another edge case we need to account for here 😬 In the `uiSchema`, we can define "rules" to conditionally show/hide/disable fields based on the value of another field. A rule looks something like:

```ts
const uiSchema = {
   type: "Layout",
   elements: [
      ...,
      {
         type: "Control",
         labelKey: "some.label.key",
         scope: "/properties/description",
         rule: {
            effect: "DISABLE",
            condition: {
               scope: "/properties/name",
               schema: { enum: [ "hello" ] }
            }
         }
      }
   ]
}
```
This essentially says, if the value of the "name" field is "hello", disable the "description" field. 

This poses a problem though because we don't want this function to match `elementOptions.scope` with a rule's condition scope and spread the options in there.

packages/common/src/core/schemas/internal/filterSchemaToUiSchema.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IConfigurationSchema, IUiSchema } from "../types";
+import { subsetSchema } from "./subsetSchema";
+
+/**
+ * Subset a json schema to only those properties needed by the uiSchema
+ * @param schema
+ * @param uiSchema
+ * @returns
+ */
+export function filterSchemaToUiSchema(
```
juliannemarik, line 11: The "rule" issue described above also appears to create the same issue here

packages/common/src/core/schemas/internal/getUiSchemaProps.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Get all the property names from a UI schema.
+ *
+ * Used to subset a json schema to only those properties used in a UI Schema
+ * @param uiSchema
+ * @returns
+ */
+export function getUiSchemaProps(uiSchema: IUiSchema): string[] {
+  return deepGetPropValues(uiSchema, "scope").map((scope: string) => {
+    return scope.split("/").pop();
```
juliannemarik, line 13: I think we're going to want to use the logic you'd previously written to grab the 2nd entry:

```ts
scope.split("/")[2]
```

because once we get into schema properties of type `object`, `uiSchema` scopes can look like `/properties/link/properties/href`. For example, given the following schema: 

<details>
  <summary>Schema</summary>

```ts
const schema = {
  required: ['title'],
  type: 'object',
  properties: {
    ...,
    link: {
      type: 'object',
      properties: {
        href: {
          type: 'string',
          format: 'url'
        },
        title: {
          type: 'string'
        }
      }
    }
  }
};
```
</details>

The `uiSchema` would look something like: 
<details>
  <summary>uiSchema</summary>

```ts
const uiSchema = {
  type: 'Layout',
  elements: [
    ...,
    {
      labelKey: 'link.href',
      scope: '/properties/link/properties/href',
      type: 'Control'
    },
    {
      labelKey: 'link.title',
      scope: '/properties/link/properties/title',
      type: 'Control'
    }
  ]
}
```
</details>

and from this function, I believe we'd want to return the "link" property from the scope as opposed to "title" and "href"

juliannemarik (commented):
sorry for not calling out these edge cases before 😬

packages/common/src/core/schemas/internal/filterSchemaToUiSchema.ts
dbouwman: Resolved in lower-level fn.

packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
dbouwman: I've addressed this, but it's specific to this case. If we run into more cases, I'll write some more specific functions that are specific to the schema structures.

packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+  // merge the options into the uiSchema
+  // hoist into applyElementOptions fn
+  options.forEach((elementOptions) => {
+    // find the entry in the uiSchema for the elementOptions.scope,
+    // and extend the elementOptions.options into the .options property
+    const elConfig = deepFind(cloneSchema, (entry) => {
+      // we dont want to apply this to rules
+      // Note: If this logic becomes more complex, we may need to
+      // write a function that's more specific to the uiSchema
+      return !entry.schema && entry.scope === elementOptions.scope;
```
juliannemarik, line 28: one very minor thing, but I think this will need to be `!entry?.schema && entry?.scope === elementOptions.scope` because you can have a rule that looks like: 

```ts
rule: {
   effect: "DISABLE",
   condition: {
      scope: "/properties/name",
      schema: { enum: [ undefined, "" ] }
   }
}
```
and the `undefined` and "" will mess up the `deepFind` function when it tries to apply the predicate on these values

juliannemarik (approved):
left one more minor comment, but otherwise, I think this looks good!

packages/common/src/core/schemas/internal/applyUiSchemaElementOptions.ts
dbouwman: good catch! I'll add test case and make sure it works

PR #881, opened 2022-09-15
fix(): update discussion auth methods

rgwozdz:
1. Description:
* fix(): canModifyPost should match creator to username.
* feat(): canModifyPostStatus - a specific auth method for modifying a post's status
* refactor(): canPostToChannel to partially leverage channel ACL
* BREAKING CHANGE: changed signature of canModifyPost, because it no longer requires the `channel` param
* 
1. Instructions for testing:
Run unit tests

### DIFF
--- a/packages/discussions/src/types.ts
+++ b/packages/discussions/src/types.ts
@@ -2,6 +2,7 @@ import { IHubRequestOptions as _IHubRequestOptions } from "@esri/hub-common";
 import {
   IPagedResponse as IRestPagedResponse,
   IPagingParams,
+  IUser,
 } from "@esri/arcgis-rest-types";
 import { Geometry } from "geojson";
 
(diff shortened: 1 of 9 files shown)


packages/discussions/src/types.ts
```diff
@@ -836,3 +852,7 @@ export interface IDiscussionsMentionMeta {
   postId: string;
   replyId?: string;
 }
+
+export interface IDiscussionsUser extends IUser {
+  username: string | undefined | null;
```
rweber-esri, line 857: I believe the way this is written, it requires a `username` property be defined, but it can have values of type `string` `undefined` or `null`. The [IUser](https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-request/src/types/user.ts#L17) that we extend defines this property as `username?: string;` which I believe means the `username` property may not be defined. Maybe we should consider rewriting this as `username?: string | null;`?

rweber-esri (approved):
LGTM, one minor comment to consider, nice work!

packages/discussions/src/types.ts
rgwozdz: Changed in https://github.com/Esri/hub.js/pull/881/commits/b4beb938c6245ebb33f9225a627b4339ecafa499.

PR #882, opened 2022-09-15
feat: project extents are converted to strings on create/update

benstoltz:
1. Description: project create/update call the addItem and update endpoints which both expect extent as a string, not a bbox.

1. Instructions for testing:


dbouwman (approved):
Thanks!

PR #888, opened 2022-09-22
feat: support serializing categories into separate param

dbouwman:
1. Description:

sending a predicate with `categories` was adding that into the `q` which the portal does not respect, and thus ignores. This correctly serializes it into a separate property which is sent as a separate form field in the search request

1. Instructions for testing:

run rests


vivzhang (approved):
LGTM!

PR #892, opened 2022-09-26
feat(hub-discussions): check groups for cannotDiscuss

rgwozdz:
1. Description:
Alter the canPostToDiscussion util so that during "authorization by group" (i.e. user's group matches channel group), we check that the group has not been tagged with typeKeyword `cannotDiscuss` which is a flag that users of this group are not allowed to participate in discussions.

1. Instructions for testing:
Added units tests, so we are covered.

Doesn't fully close, but contributes to: https://zentopia.esri.com/workspaces/engagement-workspace-61508ae50946c40014ef574f/issues/dc/hub/4751

### DIFF
--- a/packages/discussions/src/utils/channels/can-post-to-channel.ts
+++ b/packages/discussions/src/utils/channels/can-post-to-channel.ts
@@ -7,6 +7,8 @@ import {
   SharingAccess,
 } from "../../types";
 
+const CANNOT_DISCUSS = "cannotDiscuss";
+
 const ALLOWED_GROUP_ROLES = Object.freeze(["owner", "admin", "member"]);
 
 const ALLOWED_ROLES_FOR_POSTING = Object.freeze([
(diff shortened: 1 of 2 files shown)


packages/discussions/src/utils/channels/can-post-to-channel.ts
```diff
@@ -99,15 +99,19 @@ function channelAllowsPostsByThisUsersGroups(
...
+      typeKeywords,
     } = userGroup;
 
     const channelGroupPermission = acl.groups[userGroupId];
 
-    if (!channelGroupPermission) {
+    if (
+      !channelGroupPermission ||
+      !ALLOWED_GROUP_ROLES.includes(userMemberType) ||
+      typeKeywords.includes("cannotDiscuss")
```
rweber-esri, line 110: Optional, maybe make `cannotDiscuss` a `const` and reference here and below, not sure if we'll be using this typekeyword value in the future for content.

Could just return the result of the expression if we invert its conditions.

```ts
return (
  channelGroupPermission &&
  ALLOWED_GROUP_ROLES.includes(userMemberType) &&
  !typeKeywords.includes("cannotDiscuss")
);
```

rweber-esri (approved):
LGTM, let a couple optional comments.

packages/discussions/src/utils/channels/can-post-to-channel.ts
rgwozdz: Changed in https://github.com/Esri/hub.js/pull/892/commits/5b262ef91e8eaa66b798c2a6c5a28fbb2368754c.

PR #896, opened 2022-09-27
feat: add location to project schemas

benstoltz:
Adds schema entries for location to create/edit project. Also takes edit project schema and breaks it into sections.


packages/common/src/core/schemas/hubproject.ts
```diff
@@ -118,39 +129,58 @@ export const HubProjectEditUiSchema: IUiSchema = {
...
+          type: "Control",
+          options: {
+            helperText: {
+              labelKey: "{{i18nScope}}.description.helperText",
+            },
+          },
         },
-      },
+        {
+          labelKey: "{{i18nScope}}.status.label",
```
juliannemarik, line 162: So, looking back at my issue for status, I think it's actually supposed to be its own section. Sorry I didn't do that myself, but what I'd do here is create a new section with label "Status", and then not include a label on the actual field, i.e: 

```ts
{
  type: "Section",
  labelKey: "{{i18nScope}}.status.label",
  elements: [
    {
      scope: "/properties/status",
      type: "Control",
      options: {
        enum: {
           i18nScope: "{{i18nScope}}.status.enum"
        }
      }
    }
  ]
}
```

I also think this is supposed to go _after_ the location section

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -118,39 +129,58 @@ export const HubProjectEditUiSchema: IUiSchema = {
...
-      labelKey: "{{i18nScope}}.status.label",
-      scope: "/properties/status",
-      type: "Control",
-      options: {
-        enum: {
-          i18nScope: "{{i18nScope}}.status.enum",
+      type: "Section",
+      options: { headerTag: "h4" },
+      labelKey: "{{i18nScope}}.location.label",
+      elements: [
```
juliannemarik, line 177: Looking at the AC of your issue, it sounds like there's supposed to be another (toggle) field here to optionally hide the map on a project page. Is that still relevant? Or has that been removed from the criteria here?

juliannemarik (approved):

packages/common/src/core/schemas/hubproject.ts
juliannemarik: talked to Ben - he'll follow up on this. Getting this merged in for now so others can keep iterating

PR #898, opened 2022-09-28
feat: implement deepContains function

dbouwman:
1. Description:

Addresses [4868](https://devtopia.esri.com/dc/hub/issues/4868), which is part of [4802](https://devtopia.esri.com/dc/hub/issues/4802).

Note: `deepContains` is specifically _not_ exported. It will be surfaced through `HubSite.contains(...)`, which will be in a subsequent PR

1. Instructions for testing:

- run test


tomwayson (approved):
LGTM other than the duplicate line

packages/common/src/core/_internal/deepContains.ts
```diff
@@ -0,0 +1,80 @@
...
+ * - Check if `00b` is contained in `00a`'s catalog
+ * If all three are true, then `00d` is considered contained in `00a`'s catalog.
+ *
+ * @param identifier
+ * @param options
+ * @returns
+ */
+export async function deepContains(
+  identifier: string,
+  hiearchy: IDeepCatalogInfo[],
```
drspacemanphd, line 30: minor but I believe this should be _hierarchy_ in all references.

drspacemanphd (approved):

packages/common/src/core/_internal/deepContains.ts
dbouwman: As if `initiative` was not bad enough, I'm adding `hierarchy` 🤦🏼‍♂️  Open to other terms... but thanks for catching this

packages/common/src/index.ts
```diff
@@ -14,26 +14,25 @@ export * from "./downloads";
...
+export * from "./pages";
 export * from "./projects";
-export * from "./initiatives";
 export * from "./request";
 export * from "./resources";
 export * from "./search";
 export * from "./sites";
 export * from "./surveys";
 export * from "./types";
+export * from "./types";
```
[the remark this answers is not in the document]
dbouwman: ha! What's funny is that all I did in that file was sort the lines... so the dupe was already there, but not visible b/c it was not sorted. Will remove.

PR #901, opened 2022-10-03
fix: refactor so conversion of model to ihubsite is separate fn

dbouwman:
1. Description:

This extracts part of the `HubSites::fetch` fn into a separate fn that will just convert a IModel into an IHubSite. This greatly streamlines usage in OpenData-ui

1. Instructions for testing:

### DIFF
--- a/packages/common/src/sites/HubSites.ts
+++ b/packages/common/src/sites/HubSites.ts
@@ -461,6 +461,19 @@ export async function fetchSite(
   // get the model
   const model = await fetchSiteModel(identifier, requestOptions);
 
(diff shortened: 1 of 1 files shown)


packages/common/src/sites/HubSites.ts
```diff
@@ -461,6 +461,19 @@ export async function fetchSite(
   // get the model
   const model = await fetchSiteModel(identifier, requestOptions);
 
+  return convertModelToSite(model, requestOptions);
```
dbouwman, line 464: Simply refactored this into a separate fn, callable when the app already has an IModel

PR #904, opened 2022-10-10
feat: show toggle to show/hide project map

benstoltz:
Expands projects to have data.view and also adds toggle to show/hide map in project edit.

### DIFF
--- a/packages/common/src/core/types/IHubProject.ts
+++ b/packages/common/src/core/types/IHubProject.ts
@@ -1,4 +1,4 @@
-import { IHubTimeline, IHubItemEntity } from "./index";
+import { IHubTimeline, IHubItemEntity, IHubProjectView } from "./index";
 import {
   IWithLayout,
   IWithPermissions,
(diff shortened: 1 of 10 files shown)


packages/common/src/core/types/IHubProjectView.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Hub Project view: defines the display properties of a project
+ */
+
```
dbouwman, line 4: I think this extra line will prevent the jsdoc from associating the comment with the interface

packages/common/src/core/types/IHubProjectView.ts
```diff
@@ -0,0 +1,7 @@
...
+export interface IHubProjectView {
+  showMap: boolean;
```
dbouwman, line 5: Add comment on prop

packages/common/src/core/types/IHubProject.ts
```diff
@@ -23,6 +23,10 @@ export interface IHubProject
    * Project Status
    */
   status: PROJECT_STATUSES;
+  /**
+   * Project display properties
+   */
+  view?: IHubProjectView;
```
dbouwman, line 29: Since this will be a common property w/ (at least) initiatives... could we generalize this?

dbouwman (approved):

PR #914, opened 2022-10-21
fix: allows for project fields to be cleared out

benstoltz:
1. Description: Adds `clearEmptyFields` to `getBasePropertyMap` so that empty fields can be cleared.

1. Instructions for testing:


packages/common/test/models/models.test.ts
```diff
@@ -192,6 +193,8 @@ describe("model utils:", () => {
       )[0] as unknown as portalModule.IUpdateItemOptions;
 
       expect(opts.authentication).toBe(MOCK_AUTH);
+      expect(opts.params).toBeDefined();
```
dbouwman, line 196: add something that checks that description is still passed through to the http request

dbouwman (approved):
Pls add the specified test. Otherwise looks great! Thanks!

PR #915, opened 2022-10-21
feat(hub-discussions): create methods for notifications opt out

juliannaeapicella:

Description: Adds methods for channel opt out and activity routes.

- [x] used semantic commit messages
- [x] PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

### DIFF
--- a/packages/discussions/src/channels.ts
+++ b/packages/discussions/src/channels.ts
@@ -8,13 +8,13 @@ import {
   IChannel,
   IPagedResponse,
   IRemoveChannelResponse,
-  IFetchOptOutOptions,
-  ICreateOptOutOptions,
-  IRemoveOptOutOptions,
-  IRemoveActivityOptions,
-  IDeleteChannelNotificationOptOutResponse,
-  IDeleteChannelActivityResponse,
-  IChannelNotificationOptOutResponse,
+  IFetchChannelNotificationOptOutOptions,
+  ICreateChannelNotificationOptOutOptions,
+  IRemoveChannelNotificationOptOutOptions,
+  IRemoveChannelActivityOptions,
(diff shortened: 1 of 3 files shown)


packages/discussions/src/channels.ts
```diff
@@ -7,7 +7,14 @@ import {
   IRemoveChannelOptions,
   IChannel,
   IPagedResponse,
-  IRemoveChannelResponse
+  IRemoveChannelResponse,
+  IFetchOptOutOptions,
```
rweber-esri, line 11: We should probably make these interfaces names a bit more consistent. I'd recommend the following:

```
  IFetchChannelNotificationOptOutOptions,
  ICreateChannelNotificationOptOutOptions,
  IRemoveChannelNotificationOptOutOptions,
  IRemoveChannelNotificationOptOutResult,
  IRemoveChannelActivityOptions,
  IRemoveChannelActivityResult,
  IChannelNotificationOptOut,
```

packages/discussions/src/channels.ts
```diff
@@ -79,3 +86,68 @@ export function removeChannel(
...
 }
+
+/**
+ * get channel opt out status
+ *
+ * @export
+ * @param {IFetchOptOutOptions} options
+ * @return {*}
+ */
+export function fetchOptOut(
```
rweber-esri, line 97: Similar, these method names are a little generic. Lets be more explicit to avoid future collisions since.

```
fetchChannelNotifcationOptOut
createChannelNotificationOptOut
removeChannelNotificationOptOut
removeChannelActivity
```

rweber-esri (approved):
Overall this looks pretty good. See my comments about naming conventions. Once addressed this is approved.

PR #918, opened 2022-10-28
feat(hub-discussions): refactor/add/rename interfaces; update methods

rgwozdz:

BREAKING CHANGE:
Interface names have changed and been refactored

1. Description: Add/update ACL interfaces.  Refactor and reorganization channel, post, reaction interfaces into (1) separate files and (2) with reword inheritance and mixins.  Update the canPostToChannel method to reflect the new intra-group membership-type mapping.


rweber-esri (approved):
This looks reasonable to me, though I did not link this to `opendata-ui` `hub-discussions` or `hub-components` projects, so unsure the effort to update them to account for the interface changes. Are any of these changes critical to be updated in the `api` or either frontend project in the short-term? Will the frontend break if we don't update the client apps immediately after updating the api?

rgwozdz:
> This looks reasonable to me, though I did not link this to `opendata-ui` `hub-discussions` or `hub-components` projects, so unsure the effort to update them to account for the interface changes. Are any of these changes critical to be updated in the `api` or either frontend project in the short-term? Will the frontend break if we don't update the client apps immediately after updating the api?

Thanks for asking.  I don't believe implementing these changes on the discussions-service requires any coordinated change with FE.  While some of the interface names have changed, their shape has remained intact.  Others have had a shape change but only with additional optional parameters (e.g., `acl?`).

rgwozdz:
> Are any of these changes critical to be updated in the `api`?

The change to `canPostToChannel` should be used in the service ASAP.

rweber-esri:
@rgwozdz 

Thanks for the clarification. I verified we are not using the `canPostToChannel` method from the client.

PR #922, opened 2022-11-03
feat: project edit schema supports featured image

benstoltz:
1. Description: expands project edit schema with featuredImage support. Adds featuredImageUrl to getBasePropertyMap bc I forgot about it.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/hubproject.ts
+++ b/packages/common/src/core/schemas/hubproject.ts
@@ -30,6 +30,9 @@ export const HubProjectSchema: IConfigurationSchema = {
     timeline: {
       type: "object",
     },
+    featuredImage: {
+      type: "object",
+    },
   },
 } as unknown as IConfigurationSchema;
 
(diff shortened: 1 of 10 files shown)


dbouwman (approved):
pls change map to store featuredImageUrl in data.view.featuredImageUrl, otherwise gtg

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -174,6 +177,17 @@ export const HubProjectEditUiSchema: IUiSchema = {
             },
           },
         },
+        {
+          labelKey: "{{i18nScope}}.featuredImage.label",
+          scope: "/properties/featuredImage",
+          type: "Control",
+          options: {
+            control: "hub-field-input-image-picker",
+            sizeDescription: "{{i18nScope}}.featuredImage.sizeDescription",
```
juliannemarik, line 186: Maybe you already know this, but I don't think think this will work currently. You'll need to add logic to the `image-picker` field to translate the `uiSchema.options.sizeDescription`. 

I'm also wondering if we should pass it in as an option similar to `helperText` where we do: 
```ts
{
  ...,
  options: {
    sizeDescription: {
      label?: "...", // un-translated label
      labelKey?: "..." // translated label key
    }
  }
}
```

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -30,6 +30,9 @@ export const HubProjectSchema: IConfigurationSchema = {
     timeline: {
       type: "object",
     },
+    featuredImage: {
```
dbouwman, line 33: Is this still a top-level thing in the schema? tbh I'm not sure how nesting is handled in here...

benstoltz: I was handling things nested in view with logic in the controller upstream, but you raise a good point. I'll discuss w/ @juliannemarik

PR #924, opened 2022-11-04
chore(hub-common): hubProject lazy-loads CUD code

tomwayson:
affects: @esri/hub-common

1. Description:

The code to create, update, or delete a project will now be lazily loaded by the `HubProject` class. That way, users that are just viewing projects won't end up loading that code in their browser.

I've also split the `HubProjects` module into 3 modules:
- fetch - code used to fetch or search for projects
- edit - code used to edit a project
- _internal - internal fns that are shared by both of the above that are not exposed by the library

Splitting the code up into these modules/files is not strictly necessary, but I suggest this is a patter than we should adopt for all our entities in order to help our development processes and mental models reflect the division of fetch vs CUD code.

1. Instructions for testing:

`npm run test`


tomwayson:
@dbouwman I'm testing this locally via yalc in -ui but I need a valid site/project. cc @benstoltz

tomwayson:
@dbouwman I now have confirmed locally that doesn't break anything. I say let's merge and I'll PR the release to -ui.

dbouwman (approved):
Lazy all the things! Thanks!

PR #926, opened 2022-11-07
fix(hub-discussions): refactor discussion interfaces

velveetachef:

BREAKING CHANGE:
interface changes for channel, post, and reaction

1. Description:

1. Instructions for testing:


packages/discussions/src/channels/types.ts
```diff
@@ -13,10 +13,12 @@ import {
   Role,
   SharingAccess,
 } from "../shared/types";
+import { IPost } from "../types";
```
rgwozdz, line 16: I think it might be safer to import this directly from  `../posts/types` in order to protect against a circular reference.

packages/discussions/src/channels/types.ts
```diff
@@ -13,10 +13,12 @@ import {
...
 
 /**
  * Channel sorting fields
  *
+ * @export
```
rgwozdz, line 21: I think this isn't supposed to be here...?

rgwozdz (commented):
Just the two or three small comments/questions.

.nvmrc
```diff
@@ -1 +1 @@
-v12.16.1
\ No newline at end of file
+v16.13.0
```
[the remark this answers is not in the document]
velveetachef: I thought this should match the node version in the main package.json, but can change it back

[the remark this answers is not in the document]
rgwozdz: all good then.

packages/discussions/src/channels/types.ts
velveetachef: 👍

rgwozdz (approved):
Awesome! ship it.

PR #939, opened 2022-12-05
fix(hub-discussions): prevent post edit when user group is non-discus…

velveetachef:
…sable

1. Description: add check to function `canModifyPost` that prevents update if user is authorized by channel group, but the group is marked `cannotDiscuss`
  - add parameter `channel`, and only check channel groups if included (for backwards compatibility)
  - move post utils from `utils/posts` => `/utils/posts/index` and separate out function `canModifyPost`
  - adjust test location in the same manner, and separate out test for `canModifyPost`

1. Instructions for testing:

### DIFF
--- a/packages/discussions/src/utils/posts/can-modify-post.ts
+++ b/packages/discussions/src/utils/posts/can-modify-post.ts
@@ -22,6 +22,7 @@ export function canModifyPost(
   const { access, groups, orgs, allowAnonymous } = channel;
 
   if (!channel) {
+    // backwards compatibility
     return isPostCreator(post, user);
   }
 
--- a/packages/discussions/src/utils/posts/can-modify-post.ts
+++ b/packages/discussions/src/utils/posts/can-modify-post.ts
@@ -52,15 +52,11 @@ function isAuthorizedToModifyByLegacyPermissions(
     return true;
   }
 
-  if (access === SharingAccess.PRIVATE) {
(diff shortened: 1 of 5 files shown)


packages/discussions/src/utils/posts/can-modify-post.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (access === SharingAccess.PRIVATE) {
+    return isAuthorizedToModifyPostByLegacyGroup(channelGroups, userGroups);
+  }
+
+  if (access === SharingAccess.ORG) {
+    return orgs.includes(userOrgId);
+  }
+
+  return false;
+}
```
rgwozdz, line 64: you could slightly simplify this:

```ts
  if (access === SharingAccess.PUBLIC) {
    return true;
  }

  if (access === SharingAccess.ORG) {
    return orgs.includes(userOrgId);
  }

  return isAuthorizedToModifyPostByLegacyGroup(channelGroups, userGroups);
```

rgwozdz (approved):

packages/discussions/src/utils/posts/can-modify-post.ts
velveetachef: 👍

rgwozdz:
Appears to be some missing test coverage.

PR #940, opened 2022-12-07
feat: add optional assertions to IPermissionPolicy

dbouwman:
# ALPHA BRANCH

1. Description:

To address requirements for discussions, `IPermissionPolicy` supports optional `assertions` array, which has more flexible semantics for creating assertions.

# Notes
This is in a "tech preview" state - there are lots of ts-doc comments still needed and I expect some refactoring of function internals (if you see something particularly O_o pls leave a comment w/ ideas to streamline).

1. Instructions for testing:
- run tests


packages/common/src/permissions/types/PolicyResponse.ts
```diff
@@ -16,10 +16,23 @@ export type PolicyResponse =
...
+  | "array-contains-invalid-value" // assertion specifies a value not be included
+  | "array-missing-required-value" // assertion specifies a value not be included
+  | "property-mismatch"
+  | "user-not-group-member"
+  | "user-not-group-manager"
+  | "user-not-group-owner"
+  | "assertion-property-not-found"
+  | "assertion-failed" // assertion condition was not met
+  | "assertion-requires-numeric-values" // assertion requires numeric values
+  | "property-match";
```
dbouwman, line 38: This... is an aspect of things I'm not wild about... like maybe this should just be a string

packages/common/src/permissions/_internal/checkAssertion.ts
```diff
@@ -0,0 +1,232 @@
...
+): PolicyResponse {
+  let response: PolicyResponse = "granted";
+  const userGroups = context.currentUser.groups || [];
+
+  // Default the groups to all groups the user is a member of
+  let groups = mapBy("id", userGroups);
+  let failResponse: PolicyResponse = "user-not-group-member";
+
+  // if they are manage, filter to those groups
+  if (assertion.assertion === "is-group-admin") {
```
rgwozdz, line 93: nit comment: `assertion.assertion` makes me stop and get a little confused.

packages/common/src/permissions/_internal/checkAssertion.ts
```diff
@@ -0,0 +1,232 @@
...
+      if (grp.userMembership.memberType === "admin") {
+        acc.push(grp.id);
+      }
+      // if they are an owner, they are also a manager
+      if (grp.userMembership.memberType === "owner") {
+        acc.push(grp.id);
+      }
+      return acc;
+    }, []);
+  }
```
rgwozdz, line 105: nit: it would be nice if these reduce functions were packaged up into a well-named function in order to hide implementation details.

packages/common/src/permissions/_internal/checkAssertion.ts
```diff
@@ -0,0 +1,232 @@
...
+    groups = userGroups.reduce((acc, grp) => {
+      if (grp.userMembership.memberType === "admin") {
+        acc.push(grp.id);
+      }
+      // if they are an owner, they are also a manager
+      if (grp.userMembership.memberType === "owner") {
+        acc.push(grp.id);
+      }
+      return acc;
+    }, []);
```
rgwozdz, line 104: I think the upshot of our conversation last week was that there's not a clear path to making these kind of assertions on the group's properties without hard-wiring them into the code base(?).  For example, say we suddenly need to check if the user has membership-type `member`.  We need to PR and add an `is-group-member` assertion check?

packages/common/src/permissions/_internal/checkEntityPolicy.ts
```diff
@@ -51,13 +50,26 @@ export function checkEntityPolicy(
 
   if (type === "group-admin") {
```
rgwozdz, line 51: maybe this should be `group-owner-or-admin` since you check for either?

rgwozdz (approved):
Nice @dbouwman.  This is a ton of work!  Not too much to add except for the few minor comments, and clarification.

packages/common/src/permissions/_internal/checkAssertion.ts
dbouwman: Correct. We can use the assertion `type` as an extension point to add more complex logic while keeping the actual assertion structure (relatively) simple

packages/common/src/permissions/_internal/checkEntityPolicy.ts
dbouwman: I think it depends what we're checking b/c from the platform perspective, an owner has all the powers of an admin, and a member, but they do not have separate entries as `member` and `admin`. 

So - if we want to allow "owner" users to access things gated to "at least" admin, then we want the inclusive approach. 

Locally I've added `is-strict-group-admin` and `is-strict-group-member`, which check for that strict value, vs the platform transitive model. Lets discuss over a call

dbouwman: For now, we will stick with the platform semantics; if needed we can add the strict semantics.

PR #943, opened 2022-12-12
fix(hub-discussions): change restrictedBefore type to string

velveetachef:

1. Description: change type for `restrictedBefore` from Date to string

1. Instructions for testing:

### DIFF
--- a/packages/discussions/src/types.ts
+++ b/packages/discussions/src/types.ts
@@ -810,7 +810,7 @@ export interface IChannelAclPermissionDefinition {
   subCategory?: AclSubCategory;
   key?: string;
   role: Role;
-  restrictedBefore?: Date;
+  restrictedBefore?: string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/discussions/src/types.ts
```diff
@@ -837,7 +837,7 @@ export interface IChannelAclPermission
     IWithEditor,
     IWithTimestamps {
   id: string;
-  restrictedBefore: Date;
+  restrictedBefore: string;
```
rgwozdz, line 840: When it comes out of DB as a field the result of a typeorm query, is it a string?  The DB column is a timestamp, so it's possible that it will be a Date object.

velveetachef: Is the Date not converted to a string for the http response from the service?

rgwozdz: yeah, nvm.  What I was thinking of would be covered the ChannelAclEntity interface.

rgwozdz (approved):

packages/discussions/src/types.ts
velveetachef: 👍

PR #944, opened 2022-12-14
fix(hub-common): adds case for Web Mapping Application with hubSite typeKeyword in portalSearchItems

jordantsanz:
1. Description:
Fixes a bug where Web Mapping Applications with hubSite typeKeyword (basically old hub sites) would return incorrectly from hubSearch

1. Instructions for testing:
Can yalc into hub-components to test with -- I tested the query and options from an old hub site id in the hubSearch portalSearchItems tests and looked at output to verify as well -- note: hub-components may still need an update as this may not fully solve the bug, but this does now return the correct result

1. Closes Issues: [#5348](https://zentopia.esri.com/workspaces/polaris-sprint-board-614a18b38a61a3001196c48d/issues/dc/hub/5348)


mjuniper (approved):
Nice work!

PR #949, opened 2023-01-03
docs: access control docs

dbouwman:
1. Description:

Add docs for Capabilities and Permissions

1. Instructions for testing:


juliannemarik (approved):
These are really helpful docs! I think it will all make even more sense once we actually start implementing/using this logic for projects. Once we do, I'll keep note of what might need more detail in the docs, but for now I think this looks good 👍

PR #960, opened 2023-01-17
feat: add relatedItemIds to IWithViewSettings and relatedItems to project schema to support storing content on hub project

vivzhang:
affects: @esri/hub-common

ISSUES CLOSED: 5497

1. Description:

- add relatedItemIds to IWithViewSettings, add relatedItems to HubProjectSchema and HubProjectEditUiSchema

1. Instructions for testing:

Part of [5497](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/5497)

### DIFF
--- a/packages/common/src/core/schemas/hubproject.ts
+++ b/packages/common/src/core/schemas/hubproject.ts
@@ -255,7 +255,6 @@ export const HubProjectEditUiSchema: IUiSchema = {
       type: "Section",
       labelKey: "{{i18nScope}}.relatedItems.label",
       options: {
-        control: "hub-field-input-gallery-picker",
         helperText: {
           labelKey: "{{i18nScope}}.relatedItems.helperText",
         },
--- a/packages/common/src/core/schemas/hubproject.ts
+++ b/packages/common/src/core/schemas/hubproject.ts
@@ -36,7 +36,7 @@ export const HubProjectSchema: IConfigurationSchema = {
         featuredImage: {
           type: "object",
         },
-        relatedItems: {
+        relatedItemIds: {
           type: "array",
           items: {
             type: "string",
@@ -254,9 +254,15 @@ export const HubProjectEditUiSchema: IUiSchema = {
     {
       type: "Section",
       labelKey: "{{i18nScope}}.relatedItems.label",
+      options: {
+        control: "hub-field-input-gallery-picker",
+        helperText: {
+          labelKey: "{{i18nScope}}.relatedItems.helperText",
+        },
+      },
       elements: [
         {
-          scope: "/properties/relatedItems",
+          scope: "/properties/view/properties/relatedItemIds",
           type: "Control",
           options: {
             control: "hub-field-input-gallery-picker",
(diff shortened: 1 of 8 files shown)


packages/common/src/search/types/types.ts
```diff
@@ -131,3 +131,10 @@ export interface IContainsResponse {
    */
   duration?: number;
 }
+
```
vivzhang, line 134: Not sure if this is the best place to store it, or `common/src/core/types`?

juliannemarik (approved):
This looks good to me, but I might double check with Dave, especially on where to store that `IGallerySelection`

dbouwman:
The implementation is well done on @vivzhang... but I fear this naming + structure will be confusing down the line.

I think we should re-consider the naming (sorry I did not catch this last week @jul11252 ) - `IWithContent` is confusing because we refer to "every item that's not site/page/initiative/project" as "Content" (thus `IHubContent`), and that interface somewhat suggests this is how we'd associate searchable content with the entity, but really, we do that via the Catalog /`IWithCatalog` interface

This also seems like something specific to the "View" for an item... could we put this into `IWithViewSettings` as `.highlightItems: string[]`? I'm not sure we'll use the whole `IGallerySelection`, so if we can, let's strip out the other array props that are not needed. (If this is a big pain, we can probably skip this, but it's worth considering what would happen if the `.groups` array had values in it?)

dbouwman (changes requested):
see my comment on the PR

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -245,5 +251,20 @@ export const HubProjectEditUiSchema: IUiSchema = {
         },
       ],
     },
+    {
+      type: "Section",
+      labelKey: "{{i18nScope}}.relatedItems.label",
```
juliannemarik, line 256: Looks like the AC calls out that the section should have helper text, so you'll want to add that to the section's `options`. Something like: 
```ts
{
  type: "Section",
  options: {
    helperText: {
      labelKey: "{{i18nScope}}.relatedItems.helperText"
    }
  },
  elements: [ ... ]
}
```

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -36,6 +36,12 @@ export const HubProjectSchema: IConfigurationSchema = {
         featuredImage: {
           type: "object",
         },
+        relatedItems: {
```
juliannemarik, line 39: Sorry I think I might have been incorrect earlier when I said you could call this `relatedItems`. We'll actually need the property defined on the schema to match the property we're storing on the item, so we'll want to make this `relatedItemIds`

packages/common/src/core/schemas/hubproject.ts
```diff
@@ -245,5 +251,20 @@ export const HubProjectEditUiSchema: IUiSchema = {
         },
       ],
     },
+    {
+      type: "Section",
+      labelKey: "{{i18nScope}}.relatedItems.label",
+      elements: [
+        {
+          scope: "/properties/relatedItems",
```
juliannemarik, line 259: since this is nested under `view`, the scope will need to be `"/properties/view/properties/relatedItemIds"`

juliannemarik (approved):

PR #962, opened 2023-01-18
fix: resolve issue with node imports

dbouwman:
1. Description:

This was so gnarly, it's hard to describe - but here goes:

In node, `hubSearch`, `Catalog`, `Collection` were `undefined`, but worked just fine in a browser. Symptom was jest tests failing.

Built a quick `ts-node` harness, was able to debug in vs-code, and found that (again, specifics are O_o) although the undefined classes / functions were present in node_modules, and they were loading... when the environment was fully loaded, those were not exported. Its like they were exported into some other dimension.

If I removed the `hubSearch(...)` calls from `Catalog` and `Collection` then those classes and `hubSearch` function would be defined in node. O_o

After a _lot_ of work, I ended up fixing all the `import ... from "../.."` and `import ... from ".."` and `import ... from "../index"` sort of things (first commit) and when that did not fix the issue, I started re-implementing the Collection class attempting to find what was wrong. I'd completely re-implemented that in another file (as "Canary"), and `hubSearch` was working just fine in the new class but not in `Collection`. WAT.

I then did the same thing w/ `Catalog` as a new class `Wombat`... got ~90% of that done... and the same pattern was happening - `Wombat` was coming in clean, and calling through to `hubSearch` in `Wombat` worked just fine, but `Catalog` was still borked.

Then, I swapped the class names... so Wombat.ts exported `Catalog` and Catalog.ts exported `Wombat`. In doing this, I ended up changing the places where `Catalog` was imported elsewhere in the project from `../search/Catalog` to `../search`. And then everything worked.

So - for whatever reason, the path from which `Catalog` was imported into other classes, was causing `Catalog` to be `undefined` when node resolved it.

The second commit has all the import path changes for Catalog. Once that was done, then I removed the `Canary` and `Wombat` classes, renamed things back, ran all tests in node + browser and triple checked in the harness app.

I'm not even sure what/how/?? to recommend to others... beyond this description

1. Instructions for testing:
- run tests

### DIFF
--- a/packages/common/src/initiatives/HubInitiative.ts
+++ b/packages/common/src/initiatives/HubInitiative.ts
@@ -16,7 +16,7 @@ import {
   updateInitiative,
 } from "./HubInitiatives";
 
-import { Catalog } from "../search/Catalog";
+import { Catalog } from "../search";
 import { IArcGISContext } from "../ArcGISContext";
 import { HubItemEntity } from "../core/HubItemEntity";
 
--- a/packages/common/e2e/catalog-collection.e2e.ts
+++ b/packages/common/e2e/catalog-collection.e2e.ts
@@ -1,5 +1,5 @@
 import { cloneObject, IHubCatalog, IHubCollection } from "../src";
-import { Catalog } from "../src/search/Catalog";
+import { Catalog } from "../src/search";
 import { Collection } from "../src/search/Collection";
 import Artifactory from "./helpers/Artifactory";
 import config from "./helpers/config";
--- a/packages/common/e2e/helpers/Artifactory.ts
+++ b/packages/common/e2e/helpers/Artifactory.ts
@@ -4,8 +4,12 @@
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { UserSession } from "@esri/arcgis-rest-auth";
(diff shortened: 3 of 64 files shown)


mjuniper (approved):
Heroic work, Dave, many thanks!

dbouwman:
I have yalc-d this into `hub-components` and the previously failing tests now pass. Merging

packages/common/src/initiatives/HubInitiative.ts
```diff
@@ -16,7 +16,7 @@ import {
   updateInitiative,
 } from "./HubInitiatives";
 
-import { Catalog } from "../search/Catalog";
+import { Catalog } from "../search";
```
tomwayson, line 19: I can' t for the life of me figure out why this works and why `../search/Catalog` doesn't.

I wonder if you removed the re-export from index.ts if `../search/Catalog` would have worked. If so, one idea for avoiding this going forward would be to remove the  index.ts files that only re-export. This would also help avoid circular dependencies caused by `import`ing from `../index` or `../.` or any of those variants. Before we consider something like that, should think through the folder structure and what [package entrypoints](https://nodejs.org/api/packages.html#package-entry-points) we want to support.

Until then, my only suggestion is to add comments to this line and the others like it to indicate that it _has_ to be done this way otherwise we will break node.

PR #965, opened 2023-01-26
feat: hubitementity can add/clear resources

benstoltz:
1. Description: this adds add/clear resources to the IHubItemEntity.

1. Instructions for testing:


packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+export async function createResource(
+  id: string,
+  owner: string,
+  resource: any,
+  resourceName: string,
+  ro: IUserRequestOptions,
+  prefix: string = ""
+): Promise<string> {
+  try {
+    // Convert resource if needed
```
juliannemarik, line 28: I'd make this comment a bit more descriptive - why do we need to sometimes convert the resource?

packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+    const extension = resourceName.split(".").pop();
+    const resourceToUpload =
+      // If a json object, convert to blob
+      extension === "json"
```
juliannemarik, line 32: Maybe it's just me, but I find this double ternary a little difficult to parse. I find something like this to be a bit more readable:

```ts
let resourceToUpload = resource;
const extension = resourceName.split(".").pop();

if(extension === "json") {
  resourceToUpload = objectToJsonBlob(resource)
}
if(extension === "txt") {
  resourceToUpload = stringToBlob(resource)
}
```

packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param {string} resourceName
+ * @param {IUserRequestOptions} ro
+ * @param {string} [prefix=""]
+ * @return {*}  {Promise<string>}
+ */
+export async function createResource(
+  id: string,
+  owner: string,
+  resource: any,
+  resourceName: string,
```
juliannemarik, line 23: This is really minor and up to you, but maybe rename this arg to be `name` for consistency with the `arcgis-rest-portal` function args

packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+      owner,
+      resource: resourceToUpload,
+      name: resourceName,
+      prefix,
+      ...ro,
+    });
+    // if err throw
+    if (!response.success) {
+      throw new HubError(
+        "Set Item Featured Image",
```
juliannemarik, line 51: Let's make this error more generic

packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @export
+ * @param {string} id
+ * @param {string} owner
+ * @param {*} resource
+ * @param {string} resourceName
+ * @param {IUserRequestOptions} ro
+ * @param {string} [prefix=""]
+ * @return {*}  {Promise<string>}
+ */
+export async function createResource(
```
juliannemarik, line 19: I'd name this `addResource` for consistency with the `arcgis-rest-portal` function that this wraps

packages/common/src/items/createResource.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    // return url
+    const portalRestUrl = getPortalApiUrl(ro.portal);
+    if (prefix) {
+      prefix = `${prefix}/`;
+    }
+    return `${portalRestUrl}/content/items/${id}/resources/${prefix}${resourceName}`;
+  } catch (err) {
+    if (err instanceof Error) {
+      throw new HubError("Set Item Featured Image", err.message, err);
```
juliannemarik, line 63: again, let's make all these errors more generic and not specific to featured image

packages/common/src/items/removeResource.ts
```diff
@@ -0,0 +1,51 @@
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
```
juliannemarik, line 1: Now that I'm looking at this function, I see that the only thing we really get out of wrapping the `removeItemResource` function from `arcgis-rest-portal` is that we can return our own error message if there's a failure. I guess I'm now reconsidering if we really need this wrapper function

packages/common/src/items/updateResource.ts
```diff
@@ -0,0 +1,71 @@
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
```
juliannemarik, line 1: I know we'd discussed having `updateResource` as a separate function, but the code is essentially identical to `addResource` that I'm wondering how we can DRY it up - maybe adding an optional flag argument (`update`) to `addResource` (set to `false` by default) which conditionally changes which `arcgis-rest-portal` function is called. Then from this function we'd simply call `addResource` with the `update` flag set to true.

packages/common/src/resources/upsertResource.ts
```diff
@@ -0,0 +1,89 @@
...
+  owner: string,
+  resource: any,
+  name: string,
+  ro: IUserRequestOptions,
+  prefix: string = ""
+): Promise<string> {
+  try {
+    const extension = name.split(".").pop();
+    // Search against the item resources to see if the resource exists
+    const doesResourceExist: boolean = await getItemResources(id, ro).then(
```
juliannemarik, line 35: I could be reading this wrong, but I don't actually think this is returning a boolean right now. The `.find` will return the first element in the resources array that has the resource name (if any)...so I think it's either returning a resource object or undefined

packages/common/src/core/HubItemEntity.ts
```diff
@@ -1,4 +1,5 @@
 import {
+  addItemResource,
```
juliannemarik, line 2: Is there a reason these imports are added here (e.g. `addItemResource`, `objectToJsonBlob`, and `stringToBlob`? I don't see them being used

packages/common/src/resources/upsertResource.ts
```diff
@@ -0,0 +1,89 @@
...
+    // if err throw
+    if (!response.success) {
+      throw new HubError(
+        "Add Item Resource",
+        `Error adding resource ${name} to item ${id}.`
+      );
+    }
+    // return url
+    const portalRestUrl = getPortalApiUrl(ro.portal);
+    if (prefix) {
```
juliannemarik, line 75: minor, but instead of overwriting the `prefix` arg, I'd make this a new variable: 

```ts
const _prefix = prefix ? `${prefix}/` : '';
```

juliannemarik (commented):
Left a couple minor comments. Otherwise I think this looks good. 

Also, I'm not sure if you want to do it as part of this PR, but we should go back and clean up the featured image implementation:
1. remove the `uploadImageResource` util
2. leverage the `upsertResource` and `removeResource` utils in the `setFeaturedImage` and `clearFeaturedImage` methods on the `HubItemEntity` class

juliannemarik (approved):

PR #968, opened 2023-02-01
refactor(hub-common): add comment re: tiled image service family

tomwayson:
affects: @esri/hub-common

1. Description:

we should have done this as part of https://github.com/ArcGIS/opendata-ui/pull/10869

1. Instructions for testing:


tomwayson:
> we should have made this determination in hub.js. 

Well we'd need typekeywords to do that, but `getFamily()` currently only takes `type`, so we couldn't w/o making it a bigger change.

I believe that @sonofflynn89 has an upcoming issue where he faces the same issue. We can make an enhancement where `getFamily()` takes `itemOrType: IItem | string`. If we end up doing that, we can make the change for tiled image services at that time and remove the override in opendata-ui, which is one of the reasons to add this comment.

PR #971, opened 2023-02-07
feat: add wellKnownCatalog

vivzhang:
1. Description:

Add a `wellKnownCatalog` file that contains `WellKnownCatalog` and `WellKnownCollection` interfaces, `getWellKnownCatalog` and `getWellknownCollections` functions to get the well-known catalog and a list of well-known collections as well as custom catalog and collections.

**[here](https://github.com/ArcGIS/opendata-ui/pull/11514) is one of the use cases in Opendata.**

1. Instructions for testing:
Yalc hub-common with another package, import and call `getWellKnownCatalog` or `getWellknownCollections` to get the catalog or collections 

[5660](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/5660).


vivzhang:
One thing I'm reconsidering is since the Hub Gallery takes a single `IHubCollection`, maybe we should create one more function called `getWellKnownCollection` (in addition to `getWellKnownCollections`) so it can be used in components that only takes a single collection.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,268 @@
...
+  | "content"
+  | "dataset"
+  | "document"
+  | "event"
+  | "feedback"
+  | "initiative"
+  | "project"
+  | "site"
+  | "solution"
+  | "template";
```
sonofflynn89, line 23: I'm curious what other devs think, should we pluralize these? For example, I'm much more used to saying the `documents` collection rather than the `document` collection. And since we don't get auto complete in opendata-ui, I could see my mind adding the `s` in accidentally and wonder why it's not working.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,268 @@
...
+  if (i18nScope !== "") {
+    i18nScope = `${i18nScope}.`;
+  }
+  let catalog;
+  switch (name) {
+    case "myContent":
+      if (user) {
+        catalog = {
+          schemaVersion: 1,
+          title: `{{${i18nScope}catalog.myContent:translate}}`,
```
sonofflynn89, line 74: My only problem with this is that it is specific to ember interpolation. Is there any reason that we can't just have a function in ember for obtaining the translated name of a given catalog?

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,268 @@
...
+ * @param user Owner of the entity, optional but required for certain catalogs
+ * @param collectionNames A list of collection names, optional, if passed in,
+ * only those collections in the catalog will be returned
+ */
+export function getWellKnownCatalog(
+  i18nScope: string,
+  name: WellKnownCatalog,
+  entityType: EntityType,
+  user?: IUser,
+  collectionNames?: WellKnownCollection[]
```
sonofflynn89, line 39: I'm worried that as we continue to expand this function out to other entity types, we might need other arguments. Already the two optionals can make it cumbersome if somebody doesn't have the user but wants certain collections returned. What if we created another interface like `IGetWellKnownCatalogOptions`? (name just suggested)

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,268 @@
...
+        catalog = {
+          schemaVersion: 1,
+          title: `{{${i18nScope}catalog.myContent:translate}}`,
+          scopes: {
+            item: {
+              targetEntity: "item" as EntityType,
+              filters: [{ predicates: [{ owner: user.username }] }],
+            },
+          },
+          collections: getWellknownCollections(
```
sonofflynn89, line 81: If you don't want to repeat the `getWellKnownCollections` logic every single time, you could bring the initialization outside the switch statement.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,268 @@
...
+            i18nScope,
+            "item",
+            collectionNames
+          ),
+        };
+      } else {
+        throw new Error(`User needed to get "${name}" catalog`);
```
sonofflynn89, line 109: The if statement nested within a switch is a little cumbersome to read. What if we had a `validateArgs` function that checks for these incomplete scenarios and centralizes the error throwing logic?

packages/common/test/test-helpers/fake-user.ts
```diff
@@ -0,0 +1,13 @@
+import { IUser } from "@esri/arcgis-rest-types";
+
+export const mockUser = {
+  username: "vader",
+  fullName: "Darth Vader",
```
sonofflynn89, line 5: Excellent choice 😂

sonofflynn89 (commented):
Awesome work! I'm super excited that we are getting the ball rolling on this and can't wait to use it throughout the app. I left some comments for you another devs to take a look at 👌

packages/common/src/search/wellKnownCatalog.ts
vivzhang: We can, but that adds more complexity IMO. With the strings wrapped in the double curlies and `:translate` at the end, it automatically gets picked up by `interpolate` and we can use this `interpolateTranslations` function [interpolateTranslations](https://github.com/Esri/hub-components/pull/974) in both hub-components and opendata.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
+import { IUser } from "@esri/arcgis-rest-types";
+import { getFamilyTypes } from "../content";
+import { EntityType, IHubCatalog, IHubCollection } from "./types";
+
+export type WellKnownCatalog =
```
dbouwman, line 5: Please add JSDoc Comments describing the reason for this type and where it's used. Same for `WellKnownCollection`

packages/common/src/search/wellKnownCatalog.ts
dbouwman: I don't have strong preferences... I checked that this does line up with the `HubFamily` type, which uses singular... I wonder if we can / should define `WellKnownCollection` in terms of `HubFamily`? 
e.g. `export type WellKnownCollection = Exclude<HubFamily, "app"> | "appAndMap";` which results in this
image 

I think this is a good means to ensure consistency - but be sure to double check the entries in the two types... I think you will also need to add `solution` to `WellCknownCollection`

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
...
+  | "dataset"
+  | "document"
+  | "event"
+  | "feedback"
+  | "initiative"
+  | "project"
+  | "site"
+  | "solution"
+  | "template";
+
```
dbouwman, line 24: add JSDoc comment

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
...
+function getDefaultCollectionNames(): string[] {
+  return ["appAndMap", "dataset", "document", "feedback", "site"];
+}
+
+/**
+ * Check if i18nScope is defined, if so add a `.` at the end
+ */
+function validateI18nScope(i18nScope: string): string {
+  return i18nScope ? `${i18nScope}.` : `${i18nScope}`;
```
dbouwman, line 122: no need for the interpolation on the second case b/c nothing is being added...

packages/common/src/search/wellKnownCatalog.ts
dbouwman: Agree with @sonofflynn89 - group all the optional props into a single interface arg, which itself is optional.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
...
+    case "item":
+      return getWellknownItemCatalog(i18nScope, name, user, collectionNames);
+    /* Add other entity handlers here, e.g. getWellknownEventCatalog */
+    default:
+      throw new Error(`Wellknown catalog not implemented for "${entityType}"`);
+  }
+}
+
+/**
+ * @param i18nScope Translation scope to be interpolated into the catalog
```
dbouwman, line 151: Add description to jsdoc comment

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
...
+      }
+      break;
+    case "organization":
+      catalog = {
+        schemaVersion: 1,
+        title: `{{${i18nScope}catalog.organization:translate}}`,
+        scopes: {
+          item: {
+            targetEntity: "item" as EntityType,
+            filters: [{ predicates: [{ access: "org" }] }],
```
dbouwman, line 214: Maybe I'm wrong, but I _think_ this just returns items with access set to the Organization... vs all items the current user can access, in their the org, _regardless of the sharing level_. I'd have expected to see `orgid: user.orgid` or something like that... if this is how we did it before, then continue with it - otherwise check with Jay to confirm expected behavior

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,266 @@
...
+        },
+        collections,
+      };
+      break;
+  }
+  return catalog;
+}
+
+/**
+ * @param i18nScope Translation scope to be interpolated into the collections
```
dbouwman, line 238: add JsDoc explanation

dbouwman (changes requested):
This is really close - please address comments and ping me for clarifications if needed. Thanks!

packages/common/src/search/wellKnownCatalog.ts
dbouwman: Another option came to mind re: @sonofflynn89's comment about not knowing why something failed b/c of the lack of typing in Ember... for Capabilities I defined the valid values in an array and created the type from that...

image

This allows for the creation of `isCapability` fn, which I use in the fn's which accept a `Capability` as an argument.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,260 @@
...
+      item: {
+        targetEntity: "item" as EntityType,
+        filters: [{ predicates }],
+      },
+    },
+    collections,
+  };
+}
+
+/**
```
dbouwman, line 76: Add description to JsDoc

packages/common/src/search/wellKnownCatalog.ts
vivzhang: You are right, we do need to add `orgid: user.orgId` for this filter

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,310 @@
...
+ * @param args arguments to validate
+ */
+function validateArgs(type: string, name: WellKnownCatalog, args: any): any {
+  switch (type) {
+    case "user":
+      if (!args?.user) {
+        throw new Error(`User needed to get "${name}" catalog`);
+      }
+  }
+}
```
vivzhang, line 111: @sonofflynn89 took your advice, I'm not sure if I overcomplicated the logic though, we could make it simpler by just checking the user, but I thought this is more expandable.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,324 @@
...
+ * @returns An IHubCollection definition JSON object
+ */
+export function getWellknownCollection(
+  i18nScope: string,
+  collectionName: WellKnownCollection,
+  entityType: EntityType
+): IHubCollection {
+  i18nScope = validateI18nScope(i18nScope);
+  return getAllCollectionsMap(i18nScope, entityType)[collectionName];
+}
```
vivzhang, line 324: @sonofflynn89

packages/common/src/search/wellKnownCatalog.ts
dbouwman: typo - `determine` :)

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,324 @@
...
+ * collectionNames is a list of names of the collections, only those collections
+ * in the catalog will be returned
+ */
+interface IGetWellKnownCatalogOptions {
+  user?: IUser;
+  collectionNames?: WellKnownCollection[];
+}
+
+/**
+ * Check if i18nScope is defined, if so add a "." at the end
```
dbouwman, line 39: please add the "Why"... why are we adding a `.` to the end if it's defined?

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -0,0 +1,324 @@
...
+ * @param i18nScope
+ * @returns i18nScope with a "." at the end if it is defined so it stays
+ * as a valid i18nScope
+ * e.g. "project.edit.catalog" vs "project.editcatalog"
+ */
+function validateI18nScope(i18nScope: string): string {
+  return i18nScope ? `${i18nScope}.` : "";
+}
+
+/** Get a single catalog based on the name, entity type and optional requests
```
dbouwman, line 49: Please run `npm run docs:serve` and verify that this comment will be included into the generated docs... I'm not sure how sensitive the generator is to the exact format of the comments

packages/common/src/search/wellKnownCatalog.ts
dbouwman: To me this is confusing... the args you are validating are typed as `IGetWellKnownCatalogOptions` so I think you should take that in vs `any`... I'm also not wild about having a "validation" fn that throws... I'd rather this returns a structure like `{valid: false, message: "User needed to get "user" catalog"}` and then have the caller throw or decide how to handle this. It will also make debugging easier b/c the call stack will be less confusing.

packages/common/src/search/wellKnownCatalog.ts
dbouwman: nit - this is not actually "validating" the i18nScope - rather it's just adding a `.` - not exactly sure what to call this fn - dotifyString? Also - I suggest adding a guard so that if the last char of the i18nScope is a `.`, don't add another `.`

packages/common/src/search/wellKnownCatalog.ts
vivzhang: Agreed about keeping the validation function pure, since we are only checking if the passed options have certain props we need, it's really unnecessary to have the validation function anymore, I'm gonna move the check and throw back to the main function for now, if we have more conditions to check in the future, we will come back and re-evaluate. Since I've done some other cleanups, the main function is less cumbersome now even with the 'if's.

vivzhang: Or, have a simpler function called `checkUser` just for checking the user and throw, to keep it DIY

packages/common/src/search/wellKnownCatalog.ts
vivzhang: Confirmed it's shown in the doc

image

PR #972, opened 2023-02-07
feat(hub-common): add support for passing an item to getFamily()

tomwayson:
1. Description:

This fixes the peer and dev dependencies between packages by making them ranges, and configures multi-semantic-release to only bump dependent packages if the new version of the dependency does not satisfy the dependent's specified range.

**IMPORTANT**:

This means that we will need to _manually_ update `peerDependencies` in dependent packages when those packages need to use features introduced by a minor release of a dependency. **I'm not aware of a way to automate or enforce that.**

It also adds a commit to support `getFamily(item)`, which is really just a way of forcing a release to verify the dependency and semantic configuration changes are working.

TODO: use item properties like type keywords to refine the family returned, for example to avoid doing stuff like [this](https://github.com/ArcGIS/opendata-ui/pull/10869/files#diff-ecc6e1e20fe4ffcee5a21b8ac9ba3aed134074232476fda8857a059a39e05320R39-R45) I'll leave that as an excercise for @tannerjt and/or @sonofflynn89 for their individual use cases.

1. Instructions for testing:

~~let CI do it's thing~~

1. pull
2. check out master (release only runs on master)
3. `git merge f/get-item-family`
4. `npm run release:dry`

You should see output like this that indicates that only hub-common will be bumped:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62622,14 +62622,14 @@
 				"@esri/arcgis-rest-auth": "^3.1.1",
 				"@esri/arcgis-rest-request": "^3.1.1",
 				"@esri/arcgis-rest-types": "^3.1.1",
-				"@esri/hub-common": "12.4.0",
+				"@esri/hub-common": "*",
 				"@types/geojson": "^7946.0.7",
 				"typescript": "^3.8.1"
 			},
 			"peerDependencies": {
 				"@esri/arcgis-rest-auth": "^2.14.0 || 3",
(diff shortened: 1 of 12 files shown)


dbouwman (approved):
This makes sense to me... hopefully it solves the current peerDep rodeo

sonofflynn89 (approved):

tomwayson:
closing in favor of #975

PR #975, opened 2023-02-08
fix(hub-sites): fix peerDependencies should not be fixed to a specifi…

tomwayson:
…c version

affects: @esri/hub-discussions, @esri/hub-downloads, @esri/hub-events, @esri/hub-initiatives, @esri/hub-search, @esri/hub-sites, @esri/hub-surveys, @esri/hub-teams

1. Description:

like #972 but w/o introducing the feature that was just there to prove that the changes to semantic config worked.

1. Instructions for testing:

1. pull
2. check out master (release only runs on master)
3. git merge f/fix-dependency-ranges
4. npm run release:dry

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,9 +140,9 @@
 		"tsc:v": "lerna run tsc:v",
 		"y:publish": "lerna run y:publish",
 		"y:push": "lerna run y:push",
-		"release:dry": "multi-semantic-release --dry-run  --deps.release=inherit --ignore-private-packages --debug",
+		"release:dry": "multi-semantic-release --dry-run --deps.prefix=^ --deps.bump=satisfy --deps.release=inherit --ignore-private-packages --debug",
     "prerelease": "npm config set workspaces-update false",
-		"release": "multi-semantic-release --deps.release=inherit --ignore-private-packages"
+		"release": "multi-semantic-release --deps.prefix=^ --deps.bump=satisfy --deps.release=inherit --ignore-private-packages"
 	},
 	"husky": {
 		"hooks": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -140,9 +140,9 @@
 		"tsc:v": "lerna run tsc:v",
 		"y:publish": "lerna run y:publish",
 		"y:push": "lerna run y:push",
-		"release:dry": "multi-semantic-release --dry-run  --deps.release=inherit --ignore-private-packages --debug",
+		"release:dry": "multi-semantic-release --dry-run --deps.prefix=^ --deps.bump=satisfy --deps.release=inherit --ignore-private-packages --debug",
     "prerelease": "npm config set workspaces-update false",
-		"release": "multi-semantic-release --deps.release=inherit --ignore-private-packages"
+		"release": "multi-semantic-release --deps.prefix=^ --deps.bump=satisfy --deps.release=inherit --ignore-private-packages"
```
tomwayson, line 145: I've tested this locally w/ dry runs of minor bumps and `--deps.bump=satisfy` seems to only update the versions that are not w/in the expected range, which is good. Unfortunately, I can't see _how_ it will update them. Will it use a `^`, or will it wipe out the existing range w/ a fixed version?  the docs imply that `--deps.prefix=` only works w/ `--deps.bump=override`.

so we may still have some manual peer dependency management to do, but I'll have to wait and see once we publish.


For example, if i add a fn to hub-common w/a `feat` commit, it tells me that hub-common will be bumped from `12.4.0` to `12.5.0` and no other packages will be changed. If I add a `feat` commit that uses that fn in hub-downloads, then it will be bumped from `12.4.0` to `12.5.0`, but _I can't te

dbouwman: let's huck and pray... hard to see how it gets "worse" so let's roll with this

dbouwman (approved):
Give'er!

PR #977, opened 2023-02-09
chore: fix imports so e2e tests pass 

dbouwman:
1. Description:

The e2e tests stopped working, complaining about functions not existing. After a lot of digging around, it was determined that the use if `from "../../index"` type imports were the cause.

This PR resolves the imports which were causing the e2es to fail, and also addressed some snarly imports related to projects and sites.

It also adds `madge` and a script for generating dependency tree svgs. This is really rough tooling at this point, and requires  installing graphviz on the machine running this. More info -> https://github.com/pahen/madge

Madge was used to locate a few circular deps - I resolved two of them, and assigned the third to @sonofflynn89 

At this point, all tests pass in node & browser, and all e2e's pass.

1. Instructions for testing:

run all tests

No issue was created - @mjuniper had problems with an e2e and I jumped in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,7 @@ test.html
 
 ## ignore ultra build cache info
 /**/*/.ultra.cache.json
+
+# Madge graphs
+/graph.svg
+/**/*/graph.svg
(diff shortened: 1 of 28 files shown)


tomwayson (approved):
thanks @dbouwman 

only suggestion is that this is more of a chore if it's really only fixing tests, but if we think there's a chance that the current release is broken in external consumers then we can leave it as a fix and publish a release.

mjuniper (approved):

PR #978, opened 2023-02-09
fix(hub-common): fix what dotifyString returns when i18nScope is defined and ending with '.'

vivzhang:
affects: @esri/hub-common

1. Description:
In `getWellKnownCatalog`, when the `i18nScope` is defined and ending with '.', we are returning "" now, but for cases like when the `i18nScope` is `catalog.`, we will want to return that instead of "". 

1. Instructions for testing:


juliannemarik (approved):

juliannemarik (commented):
Approved, but this might something you want to throw a test in for

PR #986, opened 2023-03-03
feat(hub-common): add migration for event list card

juliannaeapicella:
1. Description: Adds a new function for migrating event-list-card site schema.

1. Instructions for testing:

1. Closes Issues: #5935


packages/common/src/sites/_internal/_ensure-event-list-card.ts
```diff
@@ -0,0 +1,41 @@
...
+  setProp,
+} from "../..";
+
+/**
+ * Add telemetry config object
+ * @private
+ * @param {object} model Site Model
+ * @returns {object}
+ */
+export function _ensureEventListCard<T extends IModel | IDraft>(model: T): T {
```
rweber-esri, line 16: Minor, but this migration name suggests event-list-card configs don't exist but we'll make one/more exist, which isn't quite what we're doing. It may make more sense to rename this function to something like `_migrateEventListCardConfigs`.

packages/common/src/sites/_internal/_ensure-event-list-card.ts
```diff
@@ -0,0 +1,41 @@
...
+  const clone = cloneObject(model);
+  clone.data.values.layout.sections.forEach(
+    (section: any, sectionIndex: number) => {
+      section.rows.forEach((row: any, rowIndex: number) => {
+        row.cards.forEach((card: any, cardIndex: number) => {
```
rweber-esri, line 21: We only want this to run if the migration has not already been applied.

Additionally, we could simplify this a bit using `Array.prototype.map`, then we wouldn't need to leverage indexes and traversing long object paths which improves readability.

If we apply the migration, we need to bump the `schemaVersion` to `1.6`
```ts
export function _migrateEventListCardConfigs<T extends IModel | IDraft> (model: T): T {
  // do nothing if migration already applied
  if (model.item.properties.schemaVersion >= 1.6) return model;
  
  // apply migration
  const clone = cloneObject(model);
  clone.data.values.layout.sections.map(section => ({
    ...section,
    rows: section.rows.map(row => ({
      ...row,
      cards: row.cards.map(card => {
        if (card.component.name === 'event-list-card') {
          card.component.settings.displayMode = card.component.settings.calendarEnabled
            ? 'calendar'
            : 'list';
          delete card.component.settings.calendarEnabled;
        }
        return card;
      }),
    })),
  }));

  // increment schemaVersion
  clone.item.properties.schemaVersion = 1.6;

  return clone;
}
```

rweber-esri (changes requested):
Some quick changes to consider then I believe this should be good. Might be worth asking @dbouwman for a a quick sanity check too.

packages/common/src/sites/_internal/_ensure-event-list-card.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  IModel,
+  IDraft,
+  getProp,
+  cloneObject,
+  deleteProp,
+  setProp,
+} from "../..";
```
dbouwman, line 8: please import from files - these type of paths cause issues with the node resolver (sometimes!)

dbouwman (approved):
This looks good to me - good to have this in place as an example - we'll soon need another migration for this card that swaps the `initiativeIds` to something like `parentIds` as we separate this card from requiring an initiative :) Thanks!

rweber-esri (approved):
Thanks for making those tweaks. This looks good to me now too.

PR #987, opened 2023-03-03
feat(hub-common): populate 'geometry' for item IHubSearchResults coming from portal

sonofflynn89:
affects: @esri/hub-common

**Description**: populates the results of `portalSearchItems` with geometry if an extent is available. The need for this arose as I was creating a map search prototype and figured we didn't want to lose the work. Tests have not yet been added and we will want to validate that I have correctly filled out the property.

For reference, the interface for `result.geometry` ([IHubGeography](https://github.com/Esri/hub.js/blob/master/packages/common/src/types.ts#L208)) was created by @tomwayson 3 years ago and the interface was included in `IHubSearchResult` by @dbouwman almost a year ago. We may want to evaluate if we want any additional changes before we start using the interface.

### DIFF
--- a/packages/common/src/content/HubContent.ts
+++ b/packages/common/src/content/HubContent.ts
@@ -1,5 +1,5 @@
 import { IItem } from "@esri/arcgis-rest-types";
-import { isMaster } from "cluster";
+import { extentToPolygon, bBoxToExtent, isBBox } from "../extent";
 import { fetchItemEnrichments } from "../items/_enrichments";
 import { getProp } from "../objects";
 import { getItemThumbnailUrl } from "../resources";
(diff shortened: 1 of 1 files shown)


tomwayson:
@sonofflynn89 

> We may want to evaluate if we want any additional changes before we start using the interface.

I agree. We are using `content.boundary` is `IHubGeography`, but we wanted to get away from `provenance`, so maybe now is the time to do that. Also, we should consider the work that is being done w/ the locaion pciker. We may (also?) want to have `.location: IHubLocation`....

If we do go this route, is there a reason that we wouldn't just use `getContentBoundary()`?

https://github.com/Esri/hub.js/blob/master/packages/common/src/content/_internal.ts#L59

dbouwman:
Closing for now - we can revisit after Workspaces are released

PR #989, opened 2023-03-07
feat(): draft schema migration update

juliannaeapicella:
1. Description: Updates upgradeDraftSchema to use most recent migration.

1. Instructions for testing:

1. Closes Issues: #5935

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63969,7 +63969,7 @@
 		},
 		"packages/common": {
 			"name": "@esri/hub-common",
-			"version": "12.7.0",
+			"version": "12.8.0",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"abab": "^2.0.5",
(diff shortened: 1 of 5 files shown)


juliannaeapicella:
@sonofflynn89 @mjuniper migrateFeedConfig was also not being run on drafts. Should we add it to this migration function?

sonofflynn89:
@juliannaeapicella the migrate feed config function was introduced over a year ago as we switched to a new format for storing the DCAT configs. I don't know that we would have any drafts out in the wild that _would_ have the old config, but I guess we could call the function in here just to be safe.

rweber-esri (approved):
One comment about verifying the new migration spies are called. Otherwise looks good to me.

rweber-esri:
> @juliannaeapicella the migrate feed config function was introduced over a year ago as we switched to a new format for storing the DCAT configs. I don't know that we would have any drafts out in the wild that _would_ have the old config, but I guess we could call the function in here just to be safe.

Thanks for verifying @sonofflynn89

PR #991, opened 2023-03-08
feat: dynamic values types and functions

dbouwman:
1. Description:

Adds types for "Dynamic Values", functions for resolving dynamic values and adds props to IHubInitiative and IHubProject for storing both the dynamic value definitions (how to resolve the values) and for storing values that other entities can resolve from the project/initiative.

While the dynamic values system is generic, it's first concrete useage will be for Initiative metrics, fetched from associated Projects.

1. Instructions for testing:

At this point, really looking for input on the types and naming


mjuniper:
I gave this a good look and it all looks great. I did find the naming a little confusing but I think that is just because you are trying to be very generic - I don't have any better ideas.

packages/common/src/utils/resolveDynamicValues.ts
```diff
@@ -0,0 +1,182 @@
...
+import { memoize } from "./memoize";
+
+/**
+ * Create memoized versions of the searchItems and getSelf functions
+ */
+const memoizedItemSearch = memoize(searchItems);
+const memoizedPortalSelf = memoize(getSelf);
+
+/**
+ * Resolve a set of dynamic values. If the result is a DynamicValueDefinition, it will be resolved recursively
```
ajturner, line 24: Please describe the 'why' of the different source types (item-query, portal, service-query)

packages/common/src/utils/resolveDynamicValues.ts
```diff
@@ -0,0 +1,182 @@
...
+ * @param dynamicValues
```
ajturner, line 25: please add some docs on params - purpose, caveats, required/optional. We should set a pattern here like

`@param dynamicValues Array of dynamic value definitions that are used ... [required]`

packages/common/src/utils/resolveDynamicValues.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  return result;
+}
+
+function assertNever(x: never, source: string): never {
+  throw new Error(`Unexpected source ${source}.`);
+}
+
+/**
+ * Resolve a set of dynamic values by extracting properties
```
ajturner, line 71: when would this function be used? Are there any caveats for Online /license?

packages/common/src/utils/resolveDynamicValues.ts
```diff
@@ -0,0 +1,182 @@
...
+          // vals needs just the value - the number or string but vResult is an object
+          // so we need to get it out of the object
+          vals.push(getProp(vResult, v.outPath));
+        }
+
+        return vals;
+      },
+      Promise.resolve([] as any[])
+    );
+    const values = await Promise.all(promises);
```
ajturner, line 135: is any error handling needed for this and similar functions that may have unavailable items / layer queries, etc.

packages/common/src/utils/resolveDynamicValues.ts
```diff
@@ -0,0 +1,182 @@
...
+  const result: DynamicValues = {};
+
+  // Construct the stats definition
+  const statsDef: IStatisticDefinition = {
+    onStatisticField: valueDef.options.field,
+    statisticType: valueDef.options.statisticType,
+    outStatisticFieldName: valueDef.options.field,
+  };
+  // Execute the query
+  const response = await queryFeatures({
```
ajturner, line 165: per other comment if we need error handling. The layer could be not available (unshared or just down) or the query not valid due to changed schema, outStats not working (this happens with DC all the time)

ajturner (changes requested):

packages/common/e2e/dynamic-value-resolver.e2e.ts
```diff
@@ -0,0 +1,226 @@
...
+    orgUrlKey: context.portal.urlKey,
+  };
+  const instance = await HubInitiative.create(init, context, true);
+  // Construct typed dynamic value definitions
+  const treesDynamicValue: IDynamicItemQueryDefinition = {
+    source: "item-query",
+    sourcePath: "properties.values.treesPlanted",
+    outPath: "treesPlanted",
+    options: {
+      q: `type: "Hub Project" AND typekeywords: initiative|${instance.id}`,
```
ajturner, line 120: I'm worried if we do this type of string query in application code. It's exposing the implementation detail and could be prone to bugs due to mis-spelling and getting the quotes spacing wrong. 

would it be possible to have entity functions that would encapsulate this so instead a dev would do

`{options: ...getProjectSearchOptions( instance.id )}`

packages/common/e2e/dynamic-value-resolver.e2e.ts
```diff
@@ -0,0 +1,226 @@
...
+  const init: Partial<IHubInitiative> = {
+    name: "Test Initiative",
+    description: "This is a test initiative",
+    orgUrlKey: context.portal.urlKey,
+  };
+  const instance = await HubInitiative.create(init, context, true);
+  // Construct typed dynamic value definitions
+  const treesDynamicValue: IDynamicItemQueryDefinition = {
+    source: "item-query",
+    sourcePath: "properties.values.treesPlanted",
```
ajturner, line 117: similar to my comment on functions for search options - can we encapsulate the `sourcePath` with a function like `getDynamicQuerySourcePath( 'treesPlanted' )` which removes developer needing to know if we use `properties.value` or if we changed later to storing another way

packages/common/src/core/types/DynamicValues.ts
```diff
@@ -0,0 +1,102 @@
+import { ISearchOptions } from "@esri/arcgis-rest-portal";
```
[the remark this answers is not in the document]
dbouwman: Since that's from rest-js, VSCode will show some minimal info inline, and F12 will open the type definition file. A comment on this line won't go into the hub.js docs. I just added a link to the rest-js doc from the `ItemSearchOptions` type defintion doc string

packages/common/e2e/dynamic-value-resolver.e2e.ts
```diff
@@ -0,0 +1,226 @@
...
+  // assign values to tne metrics for each project
+  for (let i = 0; i < childCount; i++) {
+    const prj: Partial<IHubProject> = {
+      name: `Test Project ${i}`,
+      description: "This is a test project",
+      orgUrlKey: context.portal.urlKey,
+      typeKeywords: [connectingKeyword, "HubProject"],
+      values: {
+        treesPlanted: i * 10,
+        status: i % 2 === 0 ? "planned" : "started",
```
ajturner, line 185: should this enumeration be a function? or how would this be handled when there are multiple status enuemrations?

packages/common/src/utils/resolveDynamicValues.ts
dbouwman: 100% agreed - this is a draft PR just to get input on...
- does this make sense? 
- can it handle the various scenarios we thing it should? 
- can it be extended to work with a metrics api down the line 
- etc etc.

packages/common/src/core/types/DynamicValues.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Subset of the ISearchOptions interface, used for item queries
+ */
+export type ItemSearchOptions = Partial<Pick<ISearchOptions, "q" | "filter">>;
+
+/**
+ * Options used to defined a feature service query
+ */
+export interface IServiceSearchOptions {
```
tomwayson, line 17:
```suggestion
export interface IServiceQueryOptions {
```

also, I wonder if we want to use [IStatisticDefinition](https://github.com/Esri/arcgis-rest-js/blob/3.x/packages/arcgis-rest-types/src/statisticDefinition.ts) here, or at least sync up on the aggregation types.

packages/common/src/core/types/DynamicValues.ts
```diff
@@ -0,0 +1,102 @@
+import { ISearchOptions } from "@esri/arcgis-rest-portal";
+import { IConfigurationSchema } from "../schemas";
+
+/**
+ * The source for a dyanmic value. Used to determine how to resolve the value
+ */
+export type DynamicValueSource = "item-query" | "service-query" | "portal";
```
tomwayson, line 7:
```suggestion
export type DynamicValueSource = "item-search" | "service-query" | "portal";
```

packages/common/e2e/dynamic-value-resolver.e2e.ts
dbouwman: this line is just switching up values in the fake data that's generated for the tests. For actual categorical metric fields, we need to decide if we want the Parent to be able to declare the valid values for the DataValue from which the Child entity must choose (we can do this via the UI Schema - more a question of how we store that in the definition structure)

packages/common/src/core/types/IHubInitiative.ts
```diff
@@ -15,4 +11,13 @@ export interface IHubInitiative
...
-    IWithPermissions {}
+    IWithPermissions {
+  /**
+   * Definition of any dynamic values that can be resolved
+   */
+  dynamicValues?: DynamicValueDefinition[];
+  /**
+   * Container for values that can be the source for other dynamic values
+   */
+  values?: DynamicValues;
```
tomwayson, line 22: I don't understand this

packages/common/src/utils/memoize.ts
```diff
@@ -0,0 +1,32 @@
...
+    ""
+  );
+
+/**
+ * Wrap a function into a memoized version of itself
+ * `const memoizedItemSearch = memoize(searchItems);`
+ * @param fn
+ * @returns
+ */
+export const memoize = <ARGS extends unknown[], RET>(
```
tomwayson, line 16: ❤️

tomwayson (commented):
mainly suggest you get consistent w/ item _search_ and service _query_.

also, is the idea that one entity's `values` might be used as the source of another entity's `dynamicValues`? That's what I understood, but wasn't certain about from the current doc comments.

packages/common/e2e/dynamic-value-resolver.e2e.ts
dbouwman: let's discuss this - I'm leery of adding constraints at this level - there may be other scenarios where we want to pass more than just the id... storing this as a `q` string leaves things open.

packages/common/src/core/types/DynamicValues.ts
dbouwman: I like the consistency on "query". 

I looked at using IStatsDefinition, and went back and forth on it. The aggs are a subset, and in rest-js they are not their own type... it's just defined inline so I can't derive other types directly from it. 

I left out a few though - we can def add them in but they require more props to be passed, and return a different structure - def something to discuss if it's needed for MVP

packages/common/src/utils/memoize.ts
dbouwman: yeah - this is a nice function - kinda wondering if we need a means to decache after a specific amount of time though... right now, it's a forever-cache (until you reload the app), but it's good enough for MVP, and can be optimized later

packages/common/src/core/types/IHubInitiative.ts
dbouwman: This is one of the things that's hard to name - it's the prop, on the entity, which holds the dynamic values. In the Initiative/Project scenario, this would be `myProject.values.funding`.

packages/common/src/core/types/DynamicValues.ts
dbouwman: I'll make thi change and push up vs accepting the commit (I need to also change it in the e2e harness)

packages/common/src/utils/resolveDynamicValues.ts
dbouwman: added

packages/common/src/utils/resolveDynamicValues.ts
dbouwman: see next comment - tldr; this is draft to get feedback / refactor before adding error trapping / tests etc

docs/src/guides/advanced/dynamic-values.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Dynamic Values
+navTitle: Dynamic Values
+description: Defining and Resolving run-time aggregations
+order: 55
+group: 4-advanced
+---
+
+** DRAFT / WIP / DRAFT / WIP / DRAFT / WIP / DRAFT / WIP / DRAFT / WIP / DRAFT / WIP / DRAFT **
```
ajturner, line 9: It's ok to acknowledge the status of a guide or functions. However, as a dev I wouldn't know what to expect when this would not be 'draft' - and is that referring to the Guide or the functionality itself?

I also wonder if we need this much caveat as this guide is actually implemented in code. 

Perhaps there can be a standard 'Status' section that provides few statements on what is draft (the guide, the functions,...), what the feature can be used for now, and next steps to "v1"

dbouwman:
@mjuniper / @tomwayson  / @ajturner - Thanks for all the comments & feeback. I'm going to create some additional diagrams that we can discuss because I think I've not done a great job explaining that what's in this draft PR is types and functions that act as "internals" to support Initiative/Project metrics, as well as Solution Template "deploy-time values" and "Feature Page" type run-time values. 

The examples currently have things like `initiative.dynamicValues`, but we'd actually want to have that be something like `initiative.metricDefinitions`, that is typed as `DynamicValueDefinition` or even `InitiativeMetricDefinition` which is a more opinionated / distilled version of `DynamicValueDefinition`...  anyhow - I'm going to whip up a diagram and we can discuss more when I'm back - thanks again for the close review - really appreciate it :)

packages/common/src/utils/resolveDynamicValues.ts
ajturner: ok - That's fine, and as noted in Guide we can have some simple, clear status / next action statements in the guide and likely the files too.

PR #993, opened 2023-03-10
feat(hub-common): add min viable HubContent class and supporting func…

tomwayson:
…tions

affects: @esri/hub-common

1. Description:

1. Instructions for testing:

1. Closes Issues: [5984](https://devtopia.esri.com/dc/hub/issues/5984)

1. [x] updated `peerDependencies` as needed: n/a

### DIFF
--- a/packages/common/src/content/HubContent.ts
+++ b/packages/common/src/content/HubContent.ts
@@ -1,90 +1,85 @@
-import { IItem } from "@esri/arcgis-rest-types";
-import { isMaster } from "cluster";
-import { fetchItemEnrichments } from "../items/_enrichments";
-import { getProp } from "../objects";
-import { getItemThumbnailUrl } from "../resources";
-import { IHubSearchResult } from "../search";
-import { parseInclude } from "../search/_internal/parseInclude";
-import { IHubRequestOptions } from "../types";
-import { getItemHomeUrl } from "../urls";
-import { unique } from "../util";
-import { mapBy } from "../utils";
-import { getFamily } from "./get-family";
-import { getHubRelativeUrl } from "./_internal";
+import { IArcGISContext } from "../ArcGISContext";
+import { HubItemEntity } from "../core/HubItemEntity";
+import { IHubEditableContent, IWithStoreBehavior } from "../core";
 
-/**
- * Enrich a generic search result
- * @param item
- * @param includes
- * @param requestOptions
- * @returns
- */
-export async function enrichContentSearchResult(
-  item: IItem,
-  include: string[],
-  requestOptions: IHubRequestOptions
-): Promise<IHubSearchResult> {
-  // Create the basic structure
-  const result: IHubSearchResult = {
-    access: item.access,
-    id: item.id,
-    type: item.type,
-    name: item.title,
-    owner: item.owner,
-    tags: item.tags,
-    typeKeywords: item.typeKeywords,
-    categories: item.categories,
-    summary: item.snippet || item.description,
-    createdDate: new Date(item.created),
-    createdDateSource: "item.created",
-    updatedDate: new Date(item.modified),
-    updatedDateSource: "item.modified",
-    family: getFamily(item.type),
-    links: {
-      self: "not-implemented",
-      siteRelative: "not-implemented",
-      thumbnail: "not-implemented",
-    },
-  };
-
-  // default includes
-  const DEFAULTS: string[] = [];
-
-  // Add any type-specific defaults here
-  // if (["Map Service", "Feature Service"].includes(item.type)) {
-  //   DEFAULTS = ["server.layers.length AS layerCount"];
-  // }
-
-  // if (item.type === "Web Map") {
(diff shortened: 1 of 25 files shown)


packages/common/src/content/_internal/getPropertyMap.ts
```diff
@@ -0,0 +1,22 @@
...
+ * that define the projection of properties from a IModel to an IHubProject
+ * @returns
+ * @private
+ */
+
+export function getPropertyMap(): IPropertyMap[] {
+  const map = getBasePropertyMap();
+
+  /**
+   * content-specific mappings. Note: we do not need to explicitly map
```
juliannemarik, line 15: In all of our other entity-specific `getPropertyMap` functions, we are mapping the following for `permissions` and `capabilities`: 

```ts
map.push({ objectKey: "permissions", modelKey: "data.permissions" });
map.push({ objectKey: "capabilities", modelKey: "data.settings.capabilities" });
```

so that `permissions` and `capabilities` are at the top-level of the entity object. So I'm wondering: 

1. Should you add that here? 
2. Should we move that to the `getBasePropertyMap` since it appears to be consistent between entities that extend the `IHubItemEntity`?

packages/common/src/content/HubContent.ts
```diff
@@ -1,90 +1,85 @@
...
-    family: getFamily(item.type),
-    links: {
-      self: "not-implemented",
-      siteRelative: "not-implemented",
-      thumbnail: "not-implemented",
-    },
-  };
+export class HubContent
+  extends HubItemEntity<IHubEditableContent>
+  implements IWithStoreBehavior<IHubEditableContent>
```
juliannemarik, line 7: maybe this isn't necessary or relevant for content right now, but should this also implement `IWithCatalogBehavior` and `IWithSharingBehavior`?

juliannemarik (approved):
left a few small comments/questions, but otherwise, I think this looks like a good first pass at setting up the content manager class/permissions to test out with workspaces

packages/common/src/content/HubContent.ts
```diff
@@ -1,90 +1,85 @@
...
-  const DEFAULTS: string[] = [];
+  /**
+   * Create an instance from an IHubEditableContent object
+   * @param json - JSON object to create a HubContent from
+   * @param context - ArcGIS context
+   * @returns
+   */
+  static fromJson(
+    json: Partial<IHubEditableContent>,
+    context: IArcGISContext
```
vivzhang, line 21: Should `context` be optional? If we are not fetching for content with an id or if we are accessing public content, context isn't required I believe.

packages/common/src/content/HubContent.ts
```diff
@@ -1,90 +1,85 @@
...
-  // if (["Map Service", "Feature Service"].includes(item.type)) {
-  //   DEFAULTS = ["server.layers.length AS layerCount"];
-  // }
+  /**
+   * Save the HubContent to the backing store. Currently Projects are stored as Items in Portal
+   * @returns
+   */
+  async save(): Promise<void> {
+    this._checkDestroyed();
+    const { createContent, updateContent } = await import("./edit");
```
vivzhang, line 35: I think it's cleaner if we just import these at the top of the file since this await and the two below are redundant.

packages/common/src/content/HubContent.ts
```diff
@@ -1,90 +1,85 @@
...
+   * Delete the HubContent from the store
+   * set a flag to indicate that it is destroyed
+   * @returns
+   */
+  async delete(): Promise<void> {
+    this._checkDestroyed();
+    this.isDestroyed = true;
+    const { deleteContent } = await import("./edit");
+    // Delegate to module fn
+    await deleteContent(this.entity.id, this.context.userRequestOptions);
```
vivzhang, line 76: Same here

vivzhang (approved):
Everything looks good to me, just a couple minor suggestions.

packages/common/src/content/HubContent.ts
tomwayson: We do this so that the edit code is lazily loaded. Most users will only be viewing content (or any other entity). They should not have to wait for edit code to be downloaded and parsed before being able to interact w/ the application.

Another way to handle this would be to require that the entity classes are only `import`ed from other modules that are themselves lazily loaded (i.e. Stencil components), but that is not currently the case (we `import` the classes in Ember routes) and would be hard to enforce.

So for now at least, all entity classes should use dynamic `import()` for the edit code.

vivzhang: Oh okay, that makes sense

packages/common/src/content/HubContent.ts
tomwayson: content doesn't have a catalog, but yes, it should implement `IWithSharingBehavior`

packages/common/src/content/HubContent.ts
tomwayson: resolved in #1007

PR #995, opened 2023-03-14
feat: add arcgis-js-api typings to repo & change IHubLocation interfa…

benstoltz:
…ce to match decided footprint

Adds arcgis-js-api typings to repo && changes up IHubLocation interface to match decided structure in confluence.

### DIFF
--- a/packages/common/src/core/types/IHubLocation.ts
+++ b/packages/common/src/core/types/IHubLocation.ts
@@ -18,4 +18,11 @@ export interface IHubLocation {
 
   // An esri geometry: __esri.Geometry
   geometries?: __esri.Geometry[];
+
+  // DEPRECATED: the following will be removed at next breaking version
+  provenance?: "none" | "custom" | "existing";
+  center?: number[];
+  orgSpatialReference?: ISpatialReference;
+  graphic?: any;
+  geoJson?: any;
 }
--- a/packages/common/src/core/types/IHubLocation.ts
+++ b/packages/common/src/core/types/IHubLocation.ts
@@ -1,4 +1,5 @@
 import { ISpatialReference } from "@esri/arcgis-rest-types";
+import { IHubLocationType } from "./types";
 
 /**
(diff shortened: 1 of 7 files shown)


packages/common/src/core/types/types.ts
```diff
@@ -17,3 +17,5 @@ export const EntityResourceMap: {
 } = {
   location: "location.json",
 };
+
+export type IHubLocationType = "none" | "custom" | "default" | "org-extent";
```
tomwayson, line 21: so for content our design was `"none" | "item" | "automatic"` and I think automatic was the convex hull of the features (if available) falling back to the item's extent. We need to coordinate about what are the right values that work for both content and and other entities. cc @thomas-hervey

tomwayson (changes requested):

packages/common/src/core/types/types.ts
benstoltz: My apologies, re-looked over the confluence docs and realized where I missed that. I'll be adding `"item" | "automatic"` on in.

packages/common/src/core/types/IHubLocation.ts
```diff
@@ -7,23 +8,14 @@ export interface IHubLocation {
...
-  orgSpatialReference?: ISpatialReference;
-
   // the extent of the location
   extent?: number[][];
 
-  // An esri graphics layer: __esri.GraphicsLayer
-  graphic?: any;
-
-  // GeoJSON, for hub api search purposes
-  geoJson?: any;
```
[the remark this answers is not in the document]
benstoltz: Updated with a DEPRECATED: section.

packages/common/src/core/types/types.ts
benstoltz: After talking over with Tom we are on the same page I believe, and this type has been updated.

tomwayson (approved):
thanks @benstoltz

PR #997, opened 2023-03-17
feat(): add support for Discussions entity

rweber-esri:
1. Description:
- Adds support for Discussions entity

1. Instructions for testing:

1. Closes Issues: [5906](https://devtopia.esri.com/dc/hub/issues/5906)

### DIFF
--- a/packages/common/src/core/HubItemEntity.ts
+++ b/packages/common/src/core/HubItemEntity.ts
@@ -161,6 +161,14 @@ export abstract class HubItemEntity<T extends IHubItemEntity>
     return this.entity.canDelete;
   }
 
+  /**
+   * The orgId of the Entity, if available
+   * @returns
+   */
+  get orgId() {
+    return this.entity.orgId;
+  }
+
   //#endregion IWithStoreBehavior
 
   //#region IWithSharingBehavior
--- a/packages/common/src/discussions/HubDiscussion.ts
+++ b/packages/common/src/discussions/HubDiscussion.ts
@@ -1,0 +1,145 @@
+import { DEFAULT_DISCUSSION } from "./defaults";
+import { IWithSharingBehavior, IWithStoreBehavior } from "../core/behaviors";
+import { IArcGISContext } from "../ArcGISContext";
+import { IHubDiscussion } from "../core/types";
+import { HubItemEntity } from "../core/HubItemEntity";
+import {
+  fetchDiscussion,
+  updateDiscussion,
+  createDiscussion,
+  deleteDiscussion,
+} from "./HubDiscussions";
+
+/**
+ * Hub Discussion Class
+ */
+
+export class HubDiscussion
+  extends HubItemEntity<IHubDiscussion>
+  implements IWithStoreBehavior<IHubDiscussion>, IWithSharingBehavior
+{
+  /**
+   * Create an instance from an IHubDiscussion object
+   * @param json JSON object to create a HubDiscussion from
+   * @param context ArcGIS context
+   * @returns a HubDiscussion
+   */
+  static fromJson(
+    json: Partial<IHubDiscussion>,
+    context: IArcGISContext
+  ): HubDiscussion {
+    // merge what we have with the default values
+    const pojo = this.applyDefaults(json, context);
+    return new HubDiscussion(pojo, context);
+  }
+
+  /**
+   * Create a new HubDiscussion, returning a HubDiscussion instance.
+   * Note: This does not persist the Discussion into the backing store
+   * @param partialDiscussion a partial IHubDiscussion
+   * @param context ArcGIS context
+   * @returns promise that resolves a HubDiscussion
+   */
+  static async create(
+    partialDiscussion: Partial<IHubDiscussion>,
+    context: IArcGISContext,
+    save: boolean = false
+  ): Promise<HubDiscussion> {
+    const pojo = this.applyDefaults(partialDiscussion, context);
+    // return an instance of HubDiscussion
+    const instance = HubDiscussion.fromJson(partialDiscussion, context);
+    if (save) {
+      await instance.save();
+    }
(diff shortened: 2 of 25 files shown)


packages/common/src/core/HubItemEntity.ts
```diff
@@ -161,6 +161,14 @@ export abstract class HubItemEntity<T extends IHubItemEntity>
     return this.entity.canDelete;
   }
 
+  /**
+   * The orgId of the Entity, if available
+   * @returns
+   */
+  get orgId() {
```
rweber-esri, line 168: We have a requirement that the `/discussions/:id` routes can only be accessed if the item is associated with the site. Adding `orgId` here so we can leverage the [item-in-site](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-opendata-components/addon/utils/item-in-site.js#L1) utility from opendata-ui. This begs the question if that should maybe be a custom permission that can be applied to entities, but felt that was outside the scope of this story.

packages/common/src/discussions/HubDiscussions.ts
```diff
@@ -0,0 +1,166 @@
...
+ * information that may be required
+ * @param item the discussion item
+ * @param auth request options
+ * @returns a promise that resolves a IHubDiscussion
+ */
+export async function convertItemToDiscussion(
+  item: IItem,
+  requestOptions: IRequestOptions
+): Promise<IHubDiscussion> {
+  const model = await fetchModelFromItem(item, requestOptions);
```
rweber-esri, line 162: [6052](https://devtopia.esri.com/dc/hub/issues/6052) explores whether we want to have the item configuration live in the item data or in the discussions database. May need to revisit this fetching and mapping logic if we need to map entity props from a discussions database record vs item data.

rweber-esri:
I pulled this and supporting changes in hub-components into opendata-ui via `yalc` and verified the new discussions workspace routes in opendata-ui function as expected.

juliannaeapicella (approved):
LGTM

packages/common/src/core/HubItemEntity.ts
juliannemarik: I agree this feels like something that should be handled by the permissions system, but Dave would know better, and I think this is a fine solution for now

packages/common/src/discussions/HubDiscussions.ts
```diff
@@ -0,0 +1,166 @@
+import { IUserRequestOptions } from "@esri/arcgis-rest-auth";
```
juliannemarik, line 1: Not sure if we've established this as a definitive file-structure pattern, but for all other entities, we've separated this logic into two separate `edit.ts` and `fetch.ts` modules. Not saying it's necessary to do here, but it might be nice for consistency between entities.

packages/common/src/discussions/HubDiscussion.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  async save(): Promise<void> {
+    if (this.isDestroyed) {
+      throw new Error("HubDiscussion is already destroyed.");
+    }
+
+    if (this.entity.id) {
+      // update it
+      this.entity = await updateDiscussion(
```
juliannemarik, line 114: Quoting Tom from [his own PR](https://github.com/Esri/hub.js/pull/993#discussion_r1134363869) here, and wondering if we should be lazy loading these "edit" functions (same with `createDiscussion` and `deleteDiscussion` below):

> We do this so that the edit code is lazily loaded. Most users will only be viewing content (or any other entity). They should not have to wait for edit code to be downloaded and parsed before being able to interact w/ the application.
> 
> Another way to handle this would be to require that the entity classes are only imported from other modules that are themselves lazily loaded (i.e. Stencil components), but that is not currently the case (we import the classes in Ember routes) and would be hard to enforce.
> 
> So for now at least, all entity classes should use dynamic import() for the edit code.

juliannemarik (approved):
left a couple small comments, but otherwise, I think this looks like a good first pass at setting up the discussion entity/manager class/permissions

packages/common/src/core/HubItemEntity.ts
rweber-esri: I overlooked that the `HubSite` entity has a [contains](https://github.com/Esri/hub.js/blob/f/5906-discussions-entity/packages/common/src/sites/HubSite.ts#L246) method, which checks to see if an entity is in the site's catalog, but that only satisfies part of my requirements, still need the other checks from [item-in-site](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-opendata-components/addon/utils/item-in-site.js#L1) utility.

packages/common/src/discussions/HubDiscussions.ts
rweber-esri: Thanks for pointing that out, was following the conventions of the Initiative Entity. Happy to split this into separate `edit` and `fetch` modules, as I found the naming of this file confusing compared to `HubDiscussion.ts`.

juliannemarik: oh, you're totally right...it looks like what I suggested is only the convention for projects. I agree that the naming is a bit less confusing though

dbouwman: I think Tom's intent with edit vs fetch was to have better treeshaking in the future...

packages/common/src/core/HubItemEntity.ts
dbouwman: @rweber-esri Moving forward, "associating" something with a site is that it's included in the site's catalog. That's why the `.contains` method was developed. Long-term the item-in-site function would have to be updated to handle the `IHubCatalog` structure, which... given how flexible that is, basically can't be resolved without using something like `.contains` anyhow...

PR #1012, opened 2023-03-31
refactor: move functions out of content index.ts files

dbouwman:
1. Description:

- move all functions out of common/src/content/index.ts into contentUtils.ts
- move all functions out of common/src/content/_internal/index.ts into internalContentUtils.ts
- move compose.test.ts under /content
- move content.test.ts under /content

This **should not** have any breaking changes in terms of what's exported, but given how the old way worked... it's hard to be 100% sure. The only changes to import paths were for the internal utils, which are now in `internalContentUtils.ts`, and I made a few other imports (in the /src tree) more specific. I also ensured that in the tests, all the public exports were imported from `../../src` and the internals were from deeper.

I'm not sure this will be _everything_ that's needed to fix the issue with `resolveMetrics` but I think this a step in the right direction.

1. Instructions for testing: - run tests


tomwayson (approved):
🎉

PR #1015, opened 2023-04-03
feat: implement metrics

dbouwman:
1. Description:

First version of metric resolution functions. Moved files to `/src/metrics` and suddenly the export/import issues all went away. Not asking questions because we need to move on.

Feature-wise, I'm sure we'll end up making changes to the interfaces as we move through formal implementation so the metrics types and functions should be treated as "beta" with breaking changes possible.

Look at the [project-metrics-e2e](https://github.com/Esri/hub.js/pull/1015/files#diff-26d57f2ae25b3d1cf9142348ecee95517cfec45a126afb515ebbc6a17ac92addR34) for examples of how to work with the resolutions. 

Generally, we will want to hand "dereferenced" metric definitions to components which then handle the actual resolution, so the most common usage will be the calls to `resolveMetric(metric, context)` from a component. If we use `instance.resolveMetrics()` and then pass the resolved values into chart cards etc, we will block the render cycle... and since this is all async stuff, that could be a loooong wait.

Still todo: 
- write a guide
- return disaggregated values (only an option for some types of metrics)

1. Instructions for testing:
- run tests

1. Closes Issues: #[6038](https://devtopia.esri.com/dc/hub/issues/6083)

1. [ ] N/A updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.


dbouwman:
After indepth review, the metrics system was substantially refactored and landed in https://github.com/Esri/hub.js/pull/1028

PR #1018, opened 2023-04-05
feat(hub-common): add optional categories prop to IHubItemEntity

vivzhang:
affects: @esri/hub-common

1. Description:

- Add an optional `categories` prop in `IHubItemEntity`, this is needed for the Categories field on Project and will be useful for future item related features.
- Add "Tags" and "Categories" to ProjectSchema and ProjectUiSchemaEdit so we can add those two comboxbox fields to Project

After testing them in hub-components:

image

image

1. Instructions for testing:

### DIFF
--- a/packages/common/src/projects/_internal/ProjectSchema.ts
+++ b/packages/common/src/projects/_internal/ProjectSchema.ts
@@ -30,6 +30,18 @@ export const ProjectSchema: IConfigurationSchema = {
       enum: Object.keys(PROJECT_STATUSES),
     },
     extent: ENTITY_EXTENT_SCHEMA,
+    tags: {
+      type: "array",
+      items: {
+        type: "string",
+      },
+    },
+    categories: {
+      type: "array",
+      items: {
+        type: "string",
+      },
+    },
     view: {
       type: "object",
       properties: {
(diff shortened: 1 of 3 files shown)


packages/common/src/projects/_internal/ProjectSchema.ts
```diff
@@ -51,6 +51,18 @@ export const ProjectSchema: IConfigurationSchema = {
         timeline: {
           type: "object",
         },
+        tags: {
```
juliannemarik, line 54: `tags` and `categories` live at the top-level of the entity, so we shouldn't nest them under `view` in the `schema`

packages/common/src/projects/_internal/ProjectUiSchemaEdit.ts
```diff
@@ -58,6 +58,22 @@ export const uiSchema: IUiSchema = {
             },
           },
         },
+        {
+          labelKey: "{{i18nScope}}.fields.combobox.label",
+          scope: "/properties/view/properties/combobox",
```
juliannemarik, line 63: per the comment above, this will become `/properties/tags` rather than being nested under `/properties/view`

packages/common/src/projects/_internal/ProjectUiSchemaEdit.ts
```diff
@@ -58,6 +58,22 @@ export const uiSchema: IUiSchema = {
...
+          type: "Control",
+          options: {
+            control: "hub-field-input-combobox",
```
juliannemarik, line 66: I think you'll want to pass in `allowCustomValues` and `selectionMode` as options here as well:

```ts
{
  control: "hub-field-input-combobox",
  allowCustomValues: true,
  selectionMode: multiple
}
```

same for categories below

juliannemarik (approved):
Looks good!

PR #1022, opened 2023-04-11
feat(hub-common): add default categories to project and handlers for clearing tags and categories

vivzhang:
affects: @esri/hub-common

1. Description:
- Add a default categories to project so we can handle adding/editing projects
- Handling saving when 'tags' and 'categories' are empty

1. Instructions for testing:


packages/common/src/projects/defaults.ts
```diff
@@ -31,6 +32,7 @@ export const DEFAULT_PROJECT_MODEL: IModel = {
     description: "",
     snippet: "",
     tags: [],
+    categories: [],
```
juliannemarik, line 35: I'm not sure adding these defaults is actually necessary. It looks like when an item is created, it already defaults categories to an empty array, so this might be redundant.

packages/common/src/models/index.ts
```diff
@@ -287,5 +287,8 @@ export async function fetchModelResources(
  * @return {*} boolean
  */
 function shouldClearEmptyFields(item: IItem) {
-  return ["description", "snippet"].some((field) => item[field] === "");
+  return ["description", "snippet", "tags", "categories"].some(
+    (field) =>
+      item[field] === "" || (item[field]?.length === 1 && item[field][0] === "")
```
juliannemarik, line 292: hmm, this seems really specific. Why are we getting an array back with an empty string? Is that a result of the combobox field implementation? If so, it seems to me this logic should be handled by the field e.g. we should emit an empty array if nothing has been selected

vivzhang: Good point, I refactored the logic in combobox to return an empty array

juliannemarik (approved):

Issue #1024, opened 2023-04-12
Update fast-xml-parser to remove xml2js vulnerability


MikeTschudi:
[xml2js is vulnerable to prototype pollution](https://github.com/advisories/GHSA-776f-qx25-q3cc)

It appears to only be used by fast-xml-parser, currently at 3.21.0 in @esri/hub-common. The latest version (4.2.0) does not use xml2js.

brollison:
@dbouwman can you create a `0_TECH` issue in the dc/hub repo in Devtopia so that we can monitor priority of resolution. Once you have an issue, please link cross-link the issues if possible.

brollison:
I've created a `0_TECH` issue in the dc/hub repo for refinement and prioritization of this vulnerability.
https://devtopia.esri.com/dc/hub/issues/10767

tomwayson:
we should have closed this issue when closing the PR w/ [this resolution](https://github.com/Esri/hub.js/pull/1025#issuecomment-1505908136)

tomwayson:
re-opening as per https://github.com/Esri/hub.js/pull/1025#issuecomment-2411634048

PR #1025, opened 2023-04-12
chore: update fast-xml-parser

MikeTschudi:
1. Description:
[xml2js is vulnerable to prototype pollution](https://github.com/advisories/GHSA-776f-qx25-q3cc)

1. Instructions for testing:

#1024

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,235 +1,236 @@
 {
-	"name": "@esri/hub.js",
-	"version": "9.10.0",
-	"description": "compact, modular JavaScript wrappers for ArcGIS Hub that run in Node.js and modern browsers.",
-	"private": true,
-	"workspaces": [
-		"packages/common",
-		"packages/discussions",
-		"packages/downloads",
-		"packages/events",
-		"packages/initiatives",
-		"packages/surveys",
-		"packages/teams",
-		"packages/sites",
-		"packages/surveys",
-		"packages/search",
-		"./demos/*"
-	],
-	"devDependencies": {
-		"@chiragrupani/karma-chromium-edge-launcher": "^2.0.0",
-		"@commitlint/cli": "^11.0.0",
-		"@commitlint/config-conventional": "^11.0.0",
-		"@commitlint/config-lerna-scopes": "^11.0.0",
-		"@commitlint/prompt": "^11.0.0",
+    "name": "@esri/hub.js",
+    "version": "9.10.0",
+    "description": "compact, modular JavaScript wrappers for ArcGIS Hub that run in Node.js and modern browsers.",
+    "private": true,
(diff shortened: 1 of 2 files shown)


tomwayson (changes requested):
@MikeTschudi 

Thanks for this. We'd need to verify that none of the [breaking changes introduced in 4.0 and it's betas](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/CHANGELOG.md) affect how this library is used in hub.js. Running unit tests is not sufficient, b/c we probably stub fast-xml-parser. We'd need to actually call `fetchContent()` w/ an item id that has metadata XML and ensure that it was correctly converted to JSON. I don't think we have an e2e test set up yet that does this, so we'd have to either add one, or at least run it from the command line and verify.

Also, we should remove the changes to the root package.json. If you do that, I can work on the e2e test.

MikeTschudi:
Removed root changes. This seems to produce a confusing dual state for fast-xml-parser:  part old version, part new version.

tomwayson:
Also, it looks like the unit tests did fail: 

image

It looks like part of the breaking change is that the parser is invoked with a new signature: 

https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/2.XMLparseOptions.md

So our code and tests would need to be updated accordingly from:

```ts
const { parse } = await import("fast-xml-parser");
....
return parse(metadataXml, opts);

```

To something like:

```ts
const { XMLParser } = await import("fast-xml-parser");
....
const parser = new XMLParser(opts);
return parser.parse(metadataXml);
```

And again, we'd need to verify that there aren't other breaking changes in the options and/or what it returns.

tomwayson:
After @MikeTschudi and I looked into this a bit further, it turns out that the library w/ the vulnerability, `xml2js` is not even installed. For whatever reason, `fast-xml-parser` just [uses "xml2js" as the name of executable](https://github.com/NaturalIntelligence/fast-xml-parser/blob/v3.20.3/package.json#L18-L20) in it's [package.bin](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin) so that it's CLI can be called at the command line.

Closing this for now.

olga-knyazeva:
@tomwayson @MikeTschudi thank you for looking into this. I just want to clarify one thing: our vulnerability scan complains about `fast-xml-parser` itself, not xml2js. And as far as I can see [here](https://github.com/advisories/GHSA-x3cc-x39p-42qx) the `fast-xml-parser` below 4.1.2 has a vulnerability and still needs an upgrade.  Am I missing something?

tomwayson:
thank you @olga-knyazeva 

I've re-opened the original ticket

PR #1026, opened 2023-04-13
feat(hub-discussions): remove settings from ChannelRelation

velveetachef:

BREAKING CHANGE:
settings is no longer an available ChannelRelation

1. Description: Remove `settings` from enum ChannelRelation. This service will be updated to return the channel settings fields by default, without having to specify the `relations` query parameter

1. Instructions for testing:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22644,10 +22644,9 @@
 		},
 		"node_modules/cz-lerna-changelog/node_modules/npm/node_modules/lodash._baseindexof": {
 			"version": "3.1.0",
-			"dev": true,
+			"extraneous": true,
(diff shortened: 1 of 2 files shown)


rgwozdz (commented):
Wondering if this is a breaking change.  If I understand correctly, `settings` used to only be returned when requested, but now they are returned by default.  That seems like some additional data in some cases, but nothing missing.

Is the break because the endpoint 400s when you ask for settings?

rgwozdz:
> Yes, after this change, including settings as a channelRelation will return 400 in the service

If the service starts returning a 400, that's a breaking change on the service layer.

PR #1028, opened 2023-04-17
feat: implement metrics

dbouwman:
1. Description:

Adds the metric definitions, preprocessing fns, resolution fns, and aggregation fns. Connects into entities via `.resolveMetric(metricId)`. Guide outlines how to use it all

1. Instructions for testing:
- run tests

1. Closes Issues: #6030

1. [x]  n/a updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.


packages/common/src/core/types/Metrics.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Defines the information required from a Metric
+ */
+export interface IMetric {
+  /**
+   * Identifier for the metric. This must be static over time or linkages
+   * between items will break. Can be the camelCase version of the name
+   * as long as it's a valid Javascript property name (can't start with numbers).
+   */
+  id: string;
```
jordantsanz, line 14: Jupe and I were thinking about this yesterday when you gave your proposal for moving this to a dynamic field for the value:

I understand the concern, but I had imagined the metric id would be the cardId when used in a site's layout -- so it very well could start with numbers in that case...there's no way to safeguard this by wrapping it with quotes when we create the property or refer to it, right?

If this is the camelCase version of the name, I guess this same safeguarding comes into play with naming on the project side of things too -- so that would just be a note for front-end implementation, that project metrics wouldn't be able to start with numbers.

docs/src/guides/advanced/metrics.md
```diff
@@ -0,0 +1,214 @@
+---
```
juliannemarik, line 1: Overall I think this doc is great - very informative and readable. One thing that's maybe missing (sorry if I just missed it myself) is an explanation of how we can reference other values in a metric definition's source using `$use`? Is that still relevant/worth calling out here?

juliannemarik (commented):
This all makes sense and looks great. One thing I feel is missing is where we're planning to store metric display information on the entity. But perhaps that isn't relevant to this particular PR anymore since you've separated the metric definition from the metric display information?

jordantsanz (commented):
Overall, this makes a lot of sense to me. I think my only questions are around the creation/usage of the metricId (especially within a sites context, rather than projects), the usage of variance/standard deviation as aggregations, and to confirm that display information is no longer used for an IMetric. 

Left comments throughout, but let me know if something doesn't make sense. 

Thanks for your work, Dave 🚀

docs/src/guides/advanced/metrics.md
dbouwman: @juliannemarik thats a whole can of worms that I'd like  you and Jordan to work out in code. 

`$use` is an option, but we need to decide when to do the resolution b/c the "path" for `$use` is dependent on the shape of the whole object when it's resolved. It's  also more complex to go from a single object, into two objects, one with a `$use` to the other. Mind-bendy... 

We could also just store the metric `id` and decide to always pass a hash of "data sources/metrics" into all cards along with their configurations. This resolves a number of other similar issues we'll need to address with the editors, such as the scenario where we want to add a Metric Card to a Site, pulling the metric value comes from an Initiative item.

packages/common/src/metrics/aggregateMetrics.ts
```diff
@@ -0,0 +1,52 @@
...
+      aggregate = values.reduce((acc, v) => {
+        if (acc[v]) {
+          acc[v] += 1;
+        } else {
+          acc[v] = 1;
+        }
+        return acc;
+      }, {});
+      break;
+  }
```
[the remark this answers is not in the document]
dbouwman: The problem here is that while a feature service can return those stats, I'm not sure how we merge a set of those values meaningfully. The function above runs these stats on an `IMetricFeature[]`, which is composed of values which may be static values or service query values. Only the latter has the option for standard deviation or variance. If we had three "standard deviation" values, how would we meaningfully merge those? Not that we can't but I'm unclear what that would even mean

packages/common/src/core/types/Metrics.ts
```diff
@@ -0,0 +1,171 @@
...
+  name?: string;
+  /**
+   * Description of the metric. Will be shown in UI when user is configuring
+   * the value or source of the value
+   */
+  description?: string;
+  /**
+   * Unit of the value. In the future this will be a list of well-known units
+   */
+  units?: string;
```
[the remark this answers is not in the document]
dbouwman: good to know - the think andrew had some idea that if we have a standard set of units, in the future we could use arcade to do conversions... but that seems like a really far future.

packages/common/src/core/types/Metrics.ts
dbouwman: We can also use `createId(..)`, which takes an alpha char to create a random id. It's just easier to debug if it's something readable vs `c1238`. Main thing is that it can't change over time. So if it's not set, create it from camelCased name, removing leading numbers. And never change it :)

packages/common/src/metrics/aggregateMetrics.ts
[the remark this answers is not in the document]
jordantsanz: Sure, that makes sense -- we're covered on the actual query for that value, but was curious if that was something we'd ever need to include for aggregations -- I'm not certain how we'd meaningfully merge it, either. A non-issue, then

juliannemarik (approved):

jordantsanz (approved):
Thanks for all this, Dave!

PR #1031, opened 2023-04-18
feat: deepSet and setProp support replace for objects

benstoltz:
1. Description: deepSet previously did a merge of old/new, rather than a replace. That's fine for non-object values, but cases where we might want to go from a fleshed out object to it's 'null' equivalent need a replace path.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/objects/deep-set.ts
+++ b/packages/common/src/objects/deep-set.ts
@@ -5,7 +5,7 @@ import { cloneObject } from "../util";
  * @param {Object} target Object we want to set the property on
  * @param {String} path Dotted path to the property we want to set
  * @param {Any} value Value we want to assign to the property
- * @param {Boolean} replace If true, replace the value at the path with the new value instead of set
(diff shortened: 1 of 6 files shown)


packages/common/src/objects/deep-set.ts
```diff
@@ -5,11 +5,13 @@ import { cloneObject } from "../util";
  * @param {Object} target Object we want to set the property on
  * @param {String} path Dotted path to the property we want to set
  * @param {Any} value Value we want to assign to the property
+ * @param {Boolean} replace If true, replace the value at the path with the new value instead of set
```
dbouwman, line 8: `instead of merging`?

benstoltz: UPdated here and in setProp.

juliannemarik (commented):
I would add tests to the `deep-set.test.ts` and `set-prop.test.t` functions directly. Otherwise, this looks good to me 👍

juliannemarik (approved):

dbouwman (approved):

PR #1034, opened 2023-04-19
fix(): serialize geometry parameter for searchPosts call

rweber-esri:
1. Description:
- `searchPosts` needs to serialize the `geometry` parameter before issuing the `GET /v1/posts` request since so it gets encoded correctly.

1. Instructions for testing:

### DIFF
--- a/packages/discussions/src/posts/posts.ts
+++ b/packages/discussions/src/posts/posts.ts
@@ -26,7 +26,6 @@ export function searchPosts(
 ): Promise<IPagedResponse<IPost>> {
   const url = `/posts`;
   options.httpMethod = "GET";
-  // let data: any = options.data;
   // need to serialize geometry since this is a GET request.
   // we should consider requiring this to be a base64 string
   // to safeguard against large geometries that will exceed
--- a/packages/discussions/src/posts/posts.ts
+++ b/packages/discussions/src/posts/posts.ts
@@ -26,17 +26,18 @@ export function searchPosts(
 ): Promise<IPagedResponse<IPost>> {
   const url = `/posts`;
   options.httpMethod = "GET";
-  let data: any = options.data;
+  // let data: any = options.data;
   // need to serialize geometry since this is a GET request.
   // we should consider requiring this to be a base64 string
   // to safeguard against large geometries that will exceed
   // URL character limits
-  if (data?.geometry) {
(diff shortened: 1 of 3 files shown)


packages/discussions/src/posts/posts.ts
```diff
@@ -26,7 +26,18 @@ export function searchPosts(
 ): Promise<IPagedResponse<IPost>> {
   const url = `/posts`;
   options.httpMethod = "GET";
-  return request(url, options);
+  let data: any = options.data;
+  // need to serialize geometry since this is a GET request.
+  // we should consider requiring this to be a base64 string
+  // to safeguard against large geometries that will exceed
+  // URL character limits
+  if (data?.geometry) {
```
velveetachef, line 34: Will need to do the same for `featureGeometry`

velveetachef (approved):
Looks good. Just double checking, this is not a breaking change? Was `geometry` passed in as a json string previously?

tannerjt (approved):

tannerjt:
@rweber-esri - I agree it would be good to look into base64 encoding the geometry object in the URL.

tannerjt:
> Looks good. Just double checking, this is not a breaking change? Was `geometry` passed in as a json string previously?

It was passed in previously as a string, but from outside of searchPosts.  Now it looks like we can correctly pass in a Geometry type and it will be stringified correctly from within searchPosts.

rweber-esri:
> Looks good. Just double checking, this is not a breaking change? Was `geometry` passed in as a json string previously?

TLDR; Hub.js wasn't working correctly and the client was hacking around it.

Before this PR, if any client _correctly_ passed `geometry` or `featureGeometry` params as type `Geometry` (expected by the interface) to `searchPosts` method, those parameter values would be incorrectly encoded on the URL.

E.g. 

* `encodeURIComponent({ hello: 'world' }) // => '%5Bobject%20Object%5D'`,  which does not pass server validation. 
* `encodeURIComponent('{"hello":"world"}') // => '%7B%22hello%22%3A%22world%22%7D'`, which is correctly encoded and passes server validation.

The client application was serializing `geometry` and casted the passed options as `ISearchPostsParams` before calling `searchPosts` which prevented typescript from throwing an error, but skated around this bug in Hub.js

PR #1036, opened 2023-04-20
feat(hub-common): add automatic api detection and wellKnownCollections to hubSearch

sonofflynn89:
Description:
Part of https://devtopia.esri.com/dc/hub/issues/5748

This PR consists of two parts:
- Users of `hubSearch` no longer have to explicitly pass an API object. The target API can derived from the requestOptions and a new `site` option. If a `site` is passed and the OGC API is available, it will be targeted. Otherwise, the portal API will be targeted
- Users of `hubSearch` can now append a `collection` to `Query`. This will automatically append the filters from a well known collection to the existing query filters. If targeting the Portal API, the filters will be populated using `getWellKnownCollection`. If targeting the OGC API, those filters are populated server-side as we target a corresponding collection endpoint.

### DIFF
--- a/packages/common/src/search/types/IHubCatalog.ts
+++ b/packages/common/src/search/types/IHubCatalog.ts
@@ -1,3 +1,5 @@
+import { WellKnownCollection } from "../wellKnownCatalog";
+
 export interface IHubCatalog {
   /**
    * Title for the Gallery
@@ -65,9 +67,13 @@ export interface IQuery {
    * ensure we query the correct API
    */
   targetEntity: EntityType;
-
-  wellKnownQueryId?: string;
-
+  /**
+   * An id for a well known collection that the query should use.
+   *
+   * Note: The well known collection's filters will be used _in addition_ to
+   * any existing filters within the IQuery.filters array
+   */
+  wellKnownCollectionId?: WellKnownCollection;
   /**
    * Filters that make up the query
    */
--- a/packages/common/src/search/types/IHubCatalog.ts
+++ b/packages/common/src/search/types/IHubCatalog.ts
@@ -65,6 +65,9 @@ export interface IQuery {
    * ensure we query the correct API
    */
   targetEntity: EntityType;
+
+  wellKnownQueryId?: string;
+
   /**
    * Filters that make up the query
    */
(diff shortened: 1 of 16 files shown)


packages/common/src/search/utils.ts
```diff
@@ -79,6 +80,59 @@ export function expandApi(api: NamedApis | IApiDefinition): IApiDefinition {
...
+ * Determines Which API should be hit for the given search parameters.
+ * Hierarchy:
+ * - Target options.api if available
+ * - Target the OGC API current parameters allow
+ * - Target the Portal API based off options.requestOptions.portal
+ * @param targetEntity target entity of the query
+ * @param options search options
+ * @returns an API Definition object describing what should be targeted
+ */
+export function getApi(
```
tomwayson, line 94: can we move this and `shouldUseOgcApi` to an internal file? We should only add stuff to this file if we intend for it to be used outside of this package. If needed we can move the expandApi logic to the internal file (i.e. leave `expandApi` here and likely deprecate it, and have it call a private/internal `_expandApi()`).

tomwayson (approved):
nice work. A couple of minor suggestions, but the main thing is keeping things internal if we can.

packages/common/src/search/utils.ts
sonofflynn89: Good catch! Let me see what I can do

packages/common/src/search/types/IHubCatalog.ts
```diff
@@ -65,6 +67,13 @@ export interface IQuery {
    * ensure we query the correct API
    */
   targetEntity: EntityType;
+  /**
+   * An id for a well known collection that the query should use.
+   *
+   * Note: The well known collection's filters will be used _in addition_ to
+   * any existing filters within the IQuery.filters array
+   */
+  wellKnownCollectionId?: WellKnownCollection;
```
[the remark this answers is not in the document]
sonofflynn89: Are you opposed to calling it `collectionId`? I do just get worried that non-typescript users (i.e. opendata-ui devs) might get confused and think `collection` is a complex object.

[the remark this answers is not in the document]
tomwayson: i'm going to channel my inner Turner and say just `collection`

packages/common/src/search/utils.ts
sonofflynn89: done

PR #1058, opened 2023-05-11
feat(hub-common): change hubSearchItems to use the umbrella OGC API with the ?target query param

sonofflynn89:
1. Description:

This PR introduces some critical changes that we need to leverage the OGC API in a local dev environment for Opendata UI. For example:
- `hubSearchItems` now only hits the OGC api for the _umbrella domain_ (e.g., `hubqa.arcgis.com`) and leverages the `?target` query param as needed.
- I added utils for detecting whether a string is a collection key or a legacy search category AND for converting legacy search categories to their new collection counterparts. This functionality is needed both an opendata-ui and on the OGC API since at times we are dealing with a mixture of both.

### DIFF
--- a/packages/common/test/search/_internal/hubSearchItems.test.ts
+++ b/packages/common/test/search/_internal/hubSearchItems.test.ts
@@ -7,7 +7,6 @@ import {
   IHubSearchResult,
   IPredicate,
   IQuery,
-  RemoteServerError,
 } from "../../../src";
 
 import { UserSession } from "@esri/arcgis-rest-auth";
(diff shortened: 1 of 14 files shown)


packages/common/test/search/_internal/hubSearchItems.test.ts
```diff
@@ -818,35 +818,76 @@ describe("hubSearchItems Module |", () => {
...
+      const queryParams = { type: "CSV" };
+
+      it("appends the ?target query param when the target site domain is different from the url passed in", async () => {
+        const fakeResponse = {
+          ok: true,
+          statusText: "200: Ok",
+          status: 200,
+          json: () => Promise.resolve(),
+        };
+        const _fetch: any = async (finalUrl: string) => {
```
sonofflynn89, line 835: Maybe I should use fetchmock for these tests instead...

tomwayson (approved):
fire away

PR #1073, opened 2023-06-01
feat: add slug uri conversion functions and use in getItemBySlug

dbouwman:
1. Description:

- add helper functions that convert to / from uri slugs (`myorg::the-slug`) to (`myorg|the-slug`) and back
- use functions in entity functions that convert items to search results
- use function in `getItemBySlug` (lowest level function for fetching) so the uri slugs can be passed waaaay down, ensuring all upstream fns can handle the uri slugs cleanly.

1. Closes Issues: #6932

### DIFF
--- a/packages/common/src/projects/fetch.ts
+++ b/packages/common/src/projects/fetch.ts
@@ -5,7 +5,8 @@ import { getFamily } from "../content/get-family";
 import { getHubRelativeUrl } from "../content/_internal/internalContentUtils";
 import { IHubProject } from "../core/types";
 import { PropertyMapper } from "../core/_internal/PropertyMapper";
-import { getItemBySlug, keywordSlugToUriSlug } from "../items/slugs";
+import { getItemBySlug } from "../items/slugs";
+
 import { fetchItemEnrichments } from "../items/_enrichments";
 import { fetchModelFromItem, fetchModelResources } from "../models";
 import { IHubSearchResult } from "../search";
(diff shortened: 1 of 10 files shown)


sonofflynn89 (approved):
LGTM!

packages/common/src/projects/fetch.ts
```diff
@@ -122,7 +122,7 @@ export async function enrichProjectSearchResult(
   result.links.self = getItemHomeUrl(result.id, requestOptions);
   let slugOrId = item.id;
   if (item.properties?.slug) {
-    slugOrId = item.properties.slug.replace("|", "::");
+    slugOrId = keywordSlugToUriSlug(item.properties.slug);
   }
```
tomwayson, line 126: I'd like to wrap this up into a fn like:

```ts
const getItemIdentifier (item: IItem): string => {
  const slug = item.properties?.slug
  return slug
    ? keywordSlugToUriSlug(slug)
    : item.id
}
```

I think we would use that here as well:

https://github.com/ArcGIS/opendata-ui/pull/11993/files#diff-34f8d7a2d5fcba858cc86ce2a4cd5a6eadb17c8ff2a29469c837ebc3d61dce62R28-R32

tomwayson (approved):
my one suggestion could come in a subsequent PR if needed

packages/common/src/projects/fetch.ts
dbouwman: Great idea - added. Also made the slug converters internal only.

sonofflynn89 (approved):

tomwayson (approved):
thanks for adding `getItemIdentifier()`

PR #1080, opened 2023-06-06
feat(hub-discussions): expanded enum

juliannaeapicella:

ISSUES CLOSED: 6882

1. Description: Expands entities to have isDiscussable capabilities.

### DIFF
--- a/packages/common/src/core/_internal/PropertyMapper.ts
+++ b/packages/common/src/core/_internal/PropertyMapper.ts
@@ -1,4 +1,5 @@
 import { CANNOT_DISCUSS } from "../../discussions";
+import { isDiscussable } from "../../discussions/utils";
 import { deepSet, getProp, setProp } from "../../objects";
 import { IModel } from "../../types";
 import { cloneObject } from "../../util";
@@ -90,17 +91,11 @@ export function mapObjectToModel<T>(
   mappings: IPropertyMap[]
 ): IModel {
   if (model.item?.typeKeywords) {
-    if (
-      object.hasOwnProperty("isDiscussable") &&
-      getProp(object, "isDiscussable") === true &&
-      model.item.typeKeywords.includes(CANNOT_DISCUSS)
-    ) {
-      model.item.typeKeywords = model.item.typeKeywords.filter(
-        (keyword) => keyword !== CANNOT_DISCUSS
-      );
-    } else {
-      model.item.typeKeywords.push(CANNOT_DISCUSS);
-    }
+    model.item.typeKeywords = getProp(object, "isDiscussable")
+      ? model.item.typeKeywords.filter(
+          (typeKeyword) => typeKeyword !== CANNOT_DISCUSS
+        )
+      : [...model.item.typeKeywords, CANNOT_DISCUSS];
   }
   return mapProps(object, "objectKey", model, "modelKey", mappings);
 }
@@ -118,11 +113,7 @@ export function mapModelToObject<T>(
   object: T,
   mappings: IPropertyMap[]
 ): T {
-  if (model.item.typeKeywords?.includes(CANNOT_DISCUSS)) {
-    deepSet(object, "isDiscussable", false, true);
-  } else {
-    deepSet(object, "isDiscussable", true, true);
-  }
+  setProp("isDiscussable", isDiscussable(model.item), object);
   return mapProps(model, "modelKey", object, "objectKey", mappings);
 }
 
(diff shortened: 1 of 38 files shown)


packages/common/src/core/_internal/PropertyMapper.ts
```diff
@@ -104,6 +118,11 @@ export function mapModelToObject<T>(
   object: T,
   mappings: IPropertyMap[]
 ): T {
+  if (model.item.typeKeywords?.includes(CANNOT_DISCUSS)) {
```
rweber-esri, line 121: Maybe we should move [isDiscussable](https://github.com/Esri/hub.js/blob/master/packages/discussions/src/utils/posts/index.ts#L61) utility into `hub-common` similar to how we moved the `CANNOT_DISCUSS` constant and use that here. We probably want to avoid duplicating that logic in a lot of places in case we pivot to a different strategy of determining if something is discussable or not.

```jsx
object.isDiscussable = isDiscussable(model.item);
```

packages/common/src/core/_internal/PropertyMapper.ts
```diff
@@ -88,6 +89,19 @@ export function mapObjectToModel<T>(
   model: IModel,
   mappings: IPropertyMap[]
 ): IModel {
+  if (model.item?.typeKeywords) {
```
rweber-esri, line 92: Do we need `hasOwnProperty` and `getProp` here? See my below comment about reusing the `isDiscussable` utility currently in `hub-discussions`. I'm curious if this could be simplified to just:

```jsx
if (model.item?.typeKeywords) {
  model.item.typeKeywords = object.isDiscussable
    ? model.item.typeKeywords.filter(typeKeyword => typeKeyword !== CANNOT_DISCUSS)
    : [...model.item.typeKeywords, CANNOT_DISCUSS];
}
```

packages/common/src/core/types/IHubItemEntity.ts
```diff
@@ -41,6 +41,11 @@ export interface IHubItemEntity extends IHubEntityBase, IWithPermissions {
    */
   extent?: number[][];
 
+  /**
+   * If the item has discussions enabled
+   */
+  isDiscussable?: boolean;
```
rweber-esri, line 47: The notion of something being discussable currently applies to both groups and content. Maybe this would be better defined in an `IWithDiscussions` interface in the `traits` directory and this interface extend `IWithDiscussions`. Make sure you import/export this new trait interface from the `index.ts` file in the `traits` directory.

packages/discussions/src/types.ts
```diff
@@ -98,6 +98,7 @@ export enum PostStatus {
 export enum DiscussionType {
   GROUP = "group",
   CONTENT = "content",
+  BOARD = "board",
```
rweber-esri, line 101: I'm curious if this is explicitly defined anywhere in the A.C.? I _believe_ this is desirable, so that we can separate discussion board posts from our existing content posts in discussions panel. Andrew's vision is that an idea board can pertain to other types of items (e.g. Hub Site, PDF, Event, etc), not just items with type=discussion. In the future we'll be tackling forums, so that would probably follow this same convention to isolate those posts from board & content.

packages/discussions/src/utils/constants.ts
```diff
@@ -1,3 +1 @@
-export const CANNOT_DISCUSS = "cannotDiscuss";
```
rweber-esri, line 1: This would cause a breaking change for anything that currently imports `CANNOT_DISCUSS` from `@esri/hub-discussions`, which we should try to avoid. I'd instead import `CANNOT_DISCUSS` from `@esri/hub-common` here and simply re-export, preserving existing imports of this constant from `hub-discussions`, so we can publish these changes as a minor vs major.

rweber-esri (commented):
Overall this is looking pretty good. I added a couple minor comments about moving/re-using the isDiscussable util, simplifying the logic a bit, pulling `isDiscussable` property out into a `trait`, and re-exporting `CANNOT_DISCUSS` in `hub-discussions` to prevent a breaking change.

packages/discussions/src/types.ts
juliannaeapicella: @rweber-esri From a functionality standpoint, this is mostly to make it easier to refactor [arcgis-hub-discussions-options](https://github.com/ArcGIS/opendata-ui/blob/3213f14f287e7b6cdf447f26e660bf8979bcfbc8/packages/hub-components/src/components/arcgis-hub-discussions/components/arcgis-hub-discussions-options/arcgis-hub-discussions-options.tsx#LL35C28-L35C28), which uses DiscussionType to adjust the strings. But I figured expanding this would also be a good idea for the reasons you mention.

rweber-esri (approved):
This looks good to me. One minor comment about tweaking the `updateIsDiscussable` tests to only test the logic within that method.

PR #1089, opened 2023-06-15
feat(hub-common): add isDiscussable mapping

juliannaeapicella:
affects: @esri/hub-common

1. Description: Allows isDiscussable property on entities to be updated.

1. Instructions for testing: Can be tested in the hub-components workspace harness. Note: there is a known bug in the radio buttons where they do not reflect the current value of isDiscussable, this is known and being fixed in a separate PR.

1. Closes Issues: 6883

### DIFF
--- a/packages/common/src/core/HubItemEntity.ts
+++ b/packages/common/src/core/HubItemEntity.ts
@@ -37,7 +37,7 @@ import { IWithThumbnailBehavior } from "./behaviors/IWithThumbnailBehavior";
 import { IHubItemEntity, SettableAccessLevel } from "./types";
 import { sharedWith } from "./_internal/sharedWith";
 import { IWithDiscussionsBehavior } from "./behaviors/IWithDiscussionsBehavior";
-import { CANNOT_DISCUSS } from "../discussions";
+import { setDiscussableKeyword } from "../discussions";
 
 const FEATURED_IMAGE_FILENAME = "featuredImage.png";
 
@@ -347,11 +347,10 @@ export abstract class HubItemEntity<T extends IHubItemEntity>
    * @param isDiscussable whether to enable or disable discussions
    */
   updateIsDiscussable(isDiscussable: boolean): void {
-    const typeKeywords = isDiscussable
-      ? this.entity.typeKeywords.filter(
-          (typeKeyword) => typeKeyword !== CANNOT_DISCUSS
-        )
-      : [...this.entity.typeKeywords, CANNOT_DISCUSS];
+    const typeKeywords = setDiscussableKeyword(
+      this.entity.typeKeywords,
+      isDiscussable
+    );
     this.update({ typeKeywords, isDiscussable } as Partial<T>);
   }
 }
--- a/packages/common/src/discussions/utils.ts
+++ b/packages/common/src/discussions/utils.ts
@@ -27,10 +27,11 @@ export function setDiscussableKeyword(
   typeKeywords: string[],
   discussable: boolean
 ): string[] {
-  const updatedTypeKeywords = discussable
-    ? typeKeywords.filter(
-        (typeKeyword: string) => typeKeyword !== CANNOT_DISCUSS
-      )
-    : [...typeKeywords, CANNOT_DISCUSS];
+  const updatedTypeKeywords = typeKeywords.filter(
+    (typeKeyword: string) => typeKeyword !== CANNOT_DISCUSS
+  );
+  if (!discussable) {
+    updatedTypeKeywords.push(CANNOT_DISCUSS);
+  }
   return updatedTypeKeywords;
 }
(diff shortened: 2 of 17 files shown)


packages/common/src/discussions/utils.ts
```diff
@@ -1,16 +1,36 @@
...
+ */
+export function setDiscussableKeyword(
+  typeKeywords: string[],
+  discussable: boolean
+): string[] {
+  const updatedTypeKeywords = discussable
+    ? typeKeywords.filter(
+        (typeKeyword: string) => typeKeyword !== CANNOT_DISCUSS
+      )
+    : [...typeKeywords, CANNOT_DISCUSS];
```
dbouwman, line 34: On the off chance that things get out of sync, the second part of this will always add `CANNOT_DISCUSS` even if it's already present. I'm not sure of the risk of this occurring, but it's a possible issue. In other places I think we always remove the keyword (if present), and only add it back in if it's needed. Your call if you want to add this guard

dbouwman (approved):
Looks good - only one suggestion :)

packages/common/src/discussions/utils.ts
rweber-esri: Agreed, would be nice to avoid duplicate entries added to the array. Additionally complicating things is that `typeKeywords` is _optional_ as defined on [IItem interface](https://github.com/Esri/arcgis-rest-js/blob/97322cabd6ad07bf0b5303b9719d2c085bde3c54/packages/arcgis-rest-types/src/item.ts#L19). Maybe this should account for that and always return an Array even if the given `typeKeywords` reference is `undefined`? This method's arg definition should probably reflect the optional nature of `typeKeywords`, given that, maybe it'd make sense to reverse the order of the args since we're dealing with positional arguments and required args can't come after optional args?

packages/common/src/discussions/utils.ts
```diff
@@ -1,16 +1,36 @@
...
  * Utility to determine if a given IGroup, IItem, IHubContent, or IHubItemEntity
  * is discussable.
- * @export
- * @param {IGroup|IItem|IHubContent|IHubItemEntity} The subject to evaluate
+ * @param {IGroup|IItem|IHubContent|IHubItemEntity} subject
  * @return {boolean}
  */
 export function isDiscussable(
-  subject: IGroup | IItem | IHubContent | IHubItemEntity
+  subject: Partial<IGroup | IItem | IHubContent | IHubItemEntity>
```
rweber-esri, line 15: I _think_ this is OK, i.e. I don't _believe_ this should require a breaking change.

rweber-esri (approved):
Looks good to me. I a few quick comments to address in addition to @dbouwman recommendations.

packages/common/src/discussions/utils.ts
dbouwman: agreed - type "widening" should not break anything, where as type "tightening" should be a breaking change as it could cause builds to fail.

packages/common/src/discussions/utils.ts
dbouwman: I just verified this, but `typeKeywords` is optional for the POST to create an item... but the API will inject type specific keywords, and thus always return the `typeKeywords` as an array. What's more - even if you update the item with an empty array, the API will re-add the type specific keywords... unexpected but interesting to know.

packages/common/src/core/HubItemEntity.ts
```diff
@@ -37,7 +37,7 @@ import { IWithThumbnailBehavior } from "./behaviors/IWithThumbnailBehavior";
 import { IHubItemEntity, SettableAccessLevel } from "./types";
 import { sharedWith } from "./_internal/sharedWith";
 import { IWithDiscussionsBehavior } from "./behaviors/IWithDiscussionsBehavior";
-import { CANNOT_DISCUSS } from "../discussions";
+import { setDiscussableKeyword } from "../discussions";
```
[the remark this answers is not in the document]
juliannaeapicella: @rweber-esri I agree it's unintuitive, but it is consistent with other typeKeyword functions like `setSlugKeyword` and `setStatusKeyword`.

packages/common/src/discussions/utils.ts
rweber-esri: Nice find @dbouwman. If that's the case, maybe we're OK not checking for truthiness of the given `typeKeywords` like I proposed above and disregard my request to reorder the args add the optional indicator to `typeKeywords`

packages/common/src/core/HubItemEntity.ts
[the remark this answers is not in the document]
rweber-esri: If that convention is already used by other typekeyword building utils, then I'm fine keeping the name as-is. Thanks for the additional context!

PR #1101, opened 2023-07-05
feat: add project card view model functionality 

juliannemarik:
[7161](https://devtopia.esri.com/dc/hub/issues/7161)

### Description
- add standalone functions to convert a project entity OR hub search result to an `IHubCardViewModel` to be consumed by the suite of hub gallery components
- add `IWithCardBehavior` interface to enforce the `convertToCardViewModel` on entity classes
- implement the `IWithCardBehavior` on the `IHubProject` class
- add `getRelativeWorkspaceUrl` and `isValidEntityType` helper utils
- add/update tests


packages/common/test/projects/edit.test.ts
```diff
@@ -0,0 +1,192 @@
+import {
```
juliannemarik, line 1: Note: these tests were just moved to their own file

packages/common/test/projects/fetch.test.ts
```diff
@@ -0,0 +1,183 @@
+import {
```
juliannemarik, line 1: Note: these tests were just moved to their own file

packages/common/src/core/types/IHubCardViewModel.ts
```diff
@@ -0,0 +1,43 @@
+import { AccessLevel } from "./types";
+
+// structure the arcgis-hub-card expects to receive
+export interface IHubCardViewModel {
```
juliannemarik, line 4: Note: these interfaces were copied from `hub-components`

packages/common/src/core/behaviors/IWithCardBehavior.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface IWithCardBehavior {
+  /**
+   * Convert an entity into a card view model that can
+   * be consumed by the suite of hub gallery components
+   *
+   * @param target card link contextual target
+   * @param locale internationalization locale
+   */
+  convertToCardViewModel(
+    target: "ago" | "view" | "workspace",
```
sonofflynn89, line 12: Will the Entity class have access to the AGO home url? Or will we need to pass this in? Also remember that the `Site` entities `view` target will need to visit the site rather than a content view.

packages/common/src/projects/HubProject.ts
```diff
@@ -133,6 +137,24 @@ export class HubProject
     return getEntityEditorSchemas(i18nScope, type, options);
   }
 
+  /**
+   * Convert the project entity into a card view model that can
+   * be consumed by the suite of hug gallery components
```
sonofflynn89, line 142: I don't know, the gallery doesn't seem like the hugging type 😂

packages/common/src/projects/HubProject.ts
```diff
@@ -133,6 +137,24 @@ export class HubProject
...
+   *
+   * @param project project entity
+   * @param context auth & portal information
+   * @param target card link contextual target
+   * @param locale internationalization locale
+   */
+  static convertToCardViewModel(
```
sonofflynn89, line 149: Do we really need to have the static method? I'm wondering if we should just use the instance method and require users to leverage `Project.fromJson()` if they have an `IHubProject` instead.

packages/common/src/projects/view.ts
```diff
@@ -0,0 +1,121 @@
...
+    additionalInfo: [
+      {
+        i18nKey: "type",
+        value: entityOrSearchResult.type,
+      },
+      {
+        i18nKey: "dateUpdated",
+        value: entityOrSearchResult.updatedDate.toLocaleDateString(locale),
+      },
+      ...(entityOrSearchResult.tags?.length
```
sonofflynn89, line 97: In one way it is nice to inline all this, but it can make debugging difficult. I'd almost prefer that you append the `additionalInfo` field on later.

packages/common/src/content/_internal/internalContentUtils.ts
```diff
@@ -624,3 +624,16 @@ export const getPublisherInfo = (
...
+/**
+ * returns the last section of each category path.
+ * Example: "/categories/parent/child" > "child"
+ *
+ * @param categories an item's categories
+ */
+export const getShortenedCategories = (categories: string[]) => {
+  return categories.map((category) => {
+    const segments = category.split("/");
+    return segments[segments.length - 1];
```
sonofflynn89, line 637: Make sure you account for the empty string. I've seen customer use `''` for tags and categories before and it has tanked the gallery component.

sonofflynn89 (commented):
Thank you so much for taking this on @juliannemarik! I just have a couple questions but I think you are really close 🙏

packages/common/src/core/behaviors/IWithCardBehavior.ts
juliannemarik: This class method ultimately delegates out to the `getCardViewModelFromProjectEntity` function which constructs the card model's `titleUrl` based on the `target` supplied. To do this, we're using the same functions used to "enrich" a project's search result with links...so for `target` = "ago", we use the `getItemHomeUrl` which constructs the AGO home url, so I don't think we need to pass that in. Let me know if that makes sense though.  

r.e. `Site` entities, yes, that's a valid point, but I don't necessarily think that's a concern here since the `Site` entity will similarly define its own `getCardViewModelFromSiteEntity` function which can handle constructing a unique `view` url.

packages/common/src/projects/HubProject.ts
juliannemarik: This is a good question, and one I'd defer to @dbouwman.

packages/common/src/projects/view.ts
```diff
@@ -0,0 +1,123 @@
...
+ * @param target card link contextual target
+ * @param locale internationalization locale
+ */
+export const getCardViewModelFromProjectEntity = (
+  project: IHubProject,
+  context: IArcGISContext,
+  target: "ago" | "view" | "workspace" = "ago",
+  locale: string = "en-US"
+): IHubCardViewModel => {
+  const titleUrl = {
```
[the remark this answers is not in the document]
dbouwman: Ah - I see we don't have a fn that goes from `IHubProject` -> `IHubSearchResult`... we just have one that goes from `IItem`. This is good. Leave as-is

packages/common/src/projects/HubProject.ts
```diff
@@ -133,6 +137,24 @@ export class HubProject
...
+    project: IHubProject,
+    context: IArcGISContext,
+    target: "ago" | "view" | "workspace" = "ago",
+    locale: string = "en-US"
```
dbouwman, line 153: seeing locale drilled deep into this code makes me 🤨. Not sure how we avoid this, but it just kinda smells like we're entangling some UI layer concerns deep in here. Like maybe the `additionalInfo` interface has a `transform` prop, that tells the UI layer to convert the value to LocaleTime vs LocaleDate vs RelativeDate etc. 
Not something to change now, but something we should consider

dbouwman (approved):

packages/common/src/projects/HubProject.ts
dbouwman: I think we should see where else we need this sort of "transform" logic - it smells kinda like some stuff we need for metrics / stats card as well, so maybe we can hoist out a pattern, and solve it consistently across the product. i'll create a quick issue

PR #1111, opened 2023-07-12
feat: teach hubSearch channels

tannerjt:
1. Description:

This PR extends `hubSearch` to search for `channels` from the `discussions API`.  As part of this PR, some code has moved into `hub-common` from `hub-discussions`, but should still be a non-breaking change since anything that was moved was re-exported.  

This work is done in preparation for channels being made available in the hub gallery picker.

1. Instructions for testing:

1. Closes Issues: [7231](https://devtopia.esri.com/dc/hub/issues/7231)


packages/common/src/discussions/api/channels/channels.ts
```diff
@@ -0,0 +1,16 @@
+import { request } from "../request";
+import { IChannel, IPagedResponse, ISearchChannelsParams } from "../types";
+
+/**
+ * search channels
+ *
```
dbouwman, line 6: More comments would be great here :)

packages/common/src/discussions/api/utils/request.ts
```diff
@@ -0,0 +1,101 @@
...
+      authentication,
+      authentication.portal
+    );
+  }
+
+  return tokenPromise();
+}
+
+/**
+ * parses IHubRequestOptions and makes request against Discussions API
```
dbouwman, line 41: `IHubRequestOptions` seems wrong - I think it's `IDiscussionsRequestOptions`?

packages/common/src/search/_internal/discussionsSearchChannelsHelpers/index.ts
```diff
@@ -0,0 +1,86 @@
+import { IHubSearchOptions, IHubSearchResponse, IQuery } from "../../types";
+import HubError from "../../../HubError";
+import {
+  IChannel,
+  IPagedResponse,
+  ISearchChannels,
+  ISearchChannelsParams,
+} from "../../../discussions";
+import { discussionsSearchChannels } from "..";
+
```
dbouwman, line 10: I think you could / should put all these in the `discussionsSearchChannels.ts` file - does not seem like a benefit to add a folder, with just this file in it.

packages/common/src/search/types/IHubSearchOptions.ts
```diff
@@ -73,7 +73,7 @@ export interface IHubSearchOptions {
   /**
    * Sort direction
    */
-  sortOrder?: "desc" | "asc";
+  sortOrder?: "desc" | "asc" | "DESC" | "ASC";
```
dbouwman, line 76: If the  discussion API needs it in caps, transform it internally vs widening the type... b/c that may introduce issues elsewhere.

packages/common/src/search/types/types.ts
```diff
@@ -82,7 +82,7 @@ export interface IApiDefinition {
   // - for "arcgis-hub", the /v3/search will be added
   url: string;
   // We can add types as we add support for more
-  type: "arcgis" | "arcgis-hub";
+  type: "arcgis" | "arcgis-hub" | "discussions";
```
dbouwman, line 85: Check with @sonofflynn89 re: how we are going to pass the api type into the gallery - it's changed since the early implementation and I'm not sure how it's done.

dbouwman (approved):
Looks good - I think most of these functions could use better comments, esp descriptions of "what" the function does, assuming the reader has no idea what a discussion channel is, and then internal to the fns, add comments about "why" something is being done. Also a note to check with Caleb re: how the apiType will get set in the gallery and passed into hubSearch. I'd recommend yalc-ing this into hub-components and getting a Channel Gallery harness up and running before merging this... save some time / releases :)

packages/common/src/search/types/types.ts
sonofflynn89: @dbouwman I already went over the PR with Josh in a call and gave him some action items. There are some serious considerations we'll need to make before this stuff is gallery ready.

This API type is exactly the type of thing I was hoping you could give Josh more guidance on. Since the discussions API is technically a Hub API, I was initially wary about adding this additional type. However I am ok leaving this as is IF we think that we will ever need to get discussion channels via the OGC api instead of accessing the discussions API directly. What do you think?

sonofflynn89 (commented):
Josh and I talked at length about my concerns and he is on his way to fixing them. Just a couple more comments.

sonofflynn89 (commented):
Looking good! Just a couple remaining things

packages/common/src/search/_internal/hubSearchChannels.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @param {IPagedResponse{IChannel}} channelsResponse
+ * @param {IQuery} query
+ * @param {IHubSearchOptions} options
+ * @returns IHubSearchResponse<IChannel>
+ */
+export const toHubSearchResult = (
+  channelsResponse: IPagedResponse<IChannel>,
+  query: IQuery,
+  options: IHubSearchOptions
+): IHubSearchResponse<IChannel> => {
```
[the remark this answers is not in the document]
tannerjt: Thanks for pointing this out, working on it now.

sonofflynn89 (approved):
Looking good to me Josh! Thanks for taking all this on!

PR #1120, opened 2023-07-14
feat(hub-common): add content capability and define default collections to entity catalogs

sonofflynn89:
1. Description:

Part of [7182](https://devtopia.esri.com/dc/hub/issues/7182) and lays groundwork for [7181](https://devtopia.esri.com/dc/hub/issues/7181).

The purpose of this PR is to set the pattern for how entities with catalogs should store their default collections within the catalog object. This approach simplifies display logic and provides a consistent way for entities to eventually edit and customize the default collections.

The `entity.catalog` field will be a fully-fledged `IHubCatalog` object with the following 5 default collections (in order):
- `all`
- `dataset`
- `site`
- `document`
- `appAndMap`

Each default collection is a lightweight object that leverages the `IQuery.collection` shortcut, saving space and allowing `hubSearch` to calculate collection filters on-the-fly. For example:

```
{
  key: 'all',
  targetEntity: 'item'
  scope: {
    collection: 'all',
    targetEntity: 'item',
    filters: []
  }
}
```
Users can change the display names of a default collection by editing the `collection.label` label field, or change the order in which the collections appear by rearranging their order in the `catalog.collections` array. We've also added a wrapper `IHubCollectionPersistance` interface that allows us to store other configuration information right on the collection object before it goes into storage. For example, users can indicate that a collection should be hidden from the public by setting `collection.hidden = true`.

This PR implements our new pattern on the `Site` entity. This is the most difficult entity to deal with in this scenario since we have to juggle the new catalog structure we are implementing with old structure that sites currently rely on. We also have to convert the previous collection configurations stored in `data.values.searchCategories` into our new pattern ~~(and vice-versa for backwards compatibility).~~

**NOTE:** This PR transforms the old structure into the new structure, but DOES NOT persist the new structure in AGO. That will be done at a later date. As such, I removed the modification of `item.properties.schemaVersion` from inside the catalog migrations (the change in schema version wasn't being persisted anyways).

We need to simultaneously have both the old and new formats co-exist for a bit, ~~but due to technical differences it is very difficult to transform our new format perfectly back into the old format~~. While we believe our current transformation is "good-enough" for now, we want to call out the shortcomings we know about:

- **IHubSite.catalog will have no notion of an `event` collection**: Our new search page doesn't currently support searching for events, so we purposefully ignore the event collection in `data.values.searchCategories` when migrating the `searchCategories` configuration over to `IHubSite.catalog`.
~~- **Saving a site via the Workspaces UI will remove the `event` collection from the old search page**: 
When a site is being saved via the workspace UI, we reflect any ordering / label changes in the catalog collections back to `data.values.searchCategories` so the old search page gets them. However, because `IHubSite.catalog` doesn't have any information about the `event` search category, this presents a problem. The event collection will not appear on the workspaces UI, so it cannot be explicitly re-ordered. Rather than trying to figure out where the `event` search category should appear in `searchCategories`, we've opted to remove it entirely.~~
~~- **Saving a site via the Workspaces UI will rename the `initiative` collection to `site` on the old search page**:
On the old search page, the `site` collection is labeled "Sites" for hub basic, but is labeled "Initiatives" for hub premium. There is no change in functionality, just labeling. Unfortunately when we save a site via the workspace UI, we can't reliably detect which of the two labels was originally being uses. Since the new search view always shows "Sites" regardless of premium status, we've opted to always reflect the "Sites" label back to `searchCategories`.~~

NOTE: Testing coverage on my new functions is not complete, but once we're locked into our final approach I'll make sure to get it


packages/common/src/sites/_internal/applyDefaultCollectionMigration.ts
```diff
@@ -0,0 +1,105 @@
...
+        // name has been explicitly set or if it is relying on the default translation.
+        // As such, we only set the label on the collection object IFF an explicit
+        // override text has been configured.
+        label: null,
+        key,
+        include: [],
+        targetEntity: "item",
+        scope: {
+          targetEntity: "item",
+          collection: key,
```
sonofflynn89, line 52: This is a weird thing that happened over the evolution of the `hubSearch` subsystem. `IHubCollection` has an `IQuery` scope which is passed to `hubSearch`. In an ideal world, `hubSearch` shouldn't even know that `IHubCollection` exists. However, by using this `collection: WellKnownCollection` field in `IQuery`, we're creating a circular "dependency" of sorts where `IHubCollection` knows about `IQuery`, and `IQuery` and `hubSearch` know about `IHubCollection`.

Before we merge this in, we'll want to evaluate whether `IQuery.collection` is the best semantic name for the field and make any needed changes.

packages/common/src/sites/_internal/applyDefaultCollectionMigration.ts
```diff
@@ -0,0 +1,105 @@
...
+      .map((searchCategory) => {
+        const collectionKey = searchCategoryToCollection[searchCategory.key];
+        const collection = baseCollectionMap[collectionKey];
+        collection.label = searchCategory.overrideText || null;
+        collection.hidden = searchCategory.hidden;
+        return collection;
+      });
+
+    // the "all" collection always goes first
+    configuredCollections.unshift(baseCollectionMap.all);
```
sonofflynn89, line 95: Question: should the `all` collection actually be apart of `catalog.collections`? Or should we have it be a construct introduced by the new `arcgis-hub-catalog` component? I think there are arguments for and against persisting this (and even more philosophical arguments about whether it should exist), but I want us to really consider the ramifications before we persist the all collection in site.data for all eternity.

packages/common/src/sites/HubSites.ts
```diff
@@ -418,14 +421,19 @@ export function convertModelToSite(
  */
 export function convertSiteToModel(
   site: IHubSite,
-  requestOptions: IRequestOptions
+  _requestOptions: IRequestOptions
```
tomwayson, line 424: maybe we should deprecate this if it's not used

packages/common/src/sites/_internal/applyCatalogStructureMigration.ts
```diff
@@ -46,6 +46,8 @@ export function catalogMigration(model: IModel): IModel {
       model.data.catalogv2 = siteCatalog;
     }
     // set the schema version
+    // TODO: Should we really be changing the schema version of a site
```
tomwayson, line 49: cc @dbouwman

packages/common/src/sites/_internal/applyDefaultCollectionMigration.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @returns a migrated model with default `IHubCollectionPersistance` objects added
+ * to the catalog
+ */
+export function applyDefaultCollectionMigration(model: IModel): IModel {
+  if (getProp(model, "item.properties.schemaVersion") >= 1.8) {
+    return model;
+  } else {
+    const baseCollectionKeys: WellKnownCollection[] = [
+      // TODO: add 'all' as a wellknown collection and figure out the
+      // ramifications of doing so across the app
```
tomwayson, line 28: or have another type like: `type WellKnownCollectionOrAll = WellKnownCollection | 'all'`

tomwayson (commented):
This LGTM. I can't say I have my head completely around the ramifications of your open questions. I wonder if those can be deferred into the next story?

re: omitting events, that's fine as far as search goes, but I wonder if there will ever be a need for other entities to have events in their catalog? Or will they always be handled outside of catalogs? Same question goes for discussions. If not, does anything in this PR preclude us from making those enhancements?

tomwayson (approved):
great work @sonofflynn89!

PR #1127, opened 2023-07-19
feat: EntityEditor class 

dbouwman:
1. Description:

1. Instructions for testing:
- run tests
- pull opendata-ui branch ADD BRANCH
- build and copy output into 
  - ~/projects/opendata-ui/packages/hub-components/node_modules/@esri/hub-common
  - ~/projects/opendata-ui/node_modules/@esri/hub-common
- start harness app `yarn hc start`
- use project workspace harness to verify editing

1. Closes Issues:  7183


packages/common/src/core/schemas/internal/getCategoryItems.ts
```diff
@@ -0,0 +1,43 @@
...
+import { HubEntity } from "../../types/HubEntity";
+import { IUiSchemaComboboxItem } from "../types";
+
+/**
+ * Fetch the categorySchema for all the categories, including
+ * the ones with 0 count, then recursively collet all the child
+ * categories on the deepest level. Parse into a format that
+ * can be consumed by the combobox field
+ *
+ * TODO: the request portion of this util should probably
```
juliannemarik, line 12: this TODO can be removed

packages/common/src/core/schemas/internal/getFeaturedImageUrl.ts
```diff
@@ -0,0 +1,14 @@
...
+import { HubEntity } from "../../types/HubEntity";
+
+export function getFeaturedImageUrl(
+  entity: HubEntity,
+  context: IArcGISContext
+) {
+  const queryParams = context.isAuthenticated
+    ? `?token=${context.session.token}`
+    : "";
+  // TODO: Decide if the url should be passed in or plucked out of this deep path here
```
juliannemarik, line 12: I think this is a good question, and generally speaking, I feel like it would be nice to make this util more generic/less featured-image-specific. I just ran into needing similar functionality with Aaron who's trying to hook up editing an item's thumbnail. The differences there are: 
1. the url lives at `entity.thumbnailUrl`
2. `entity.thumbnailUrl` already has the token appended to it

So not entirely sure how to make this generic to work in both cases, but you could do something like: 

```ts
getImageUrl (
  url: string,
  context?: IArcGISContext // make this optional if you want to append the token
) {
  const queryParams = context?.isAuthenticated
      ? `?token=${context.session.token}`
      : "";

  return cacheBustUrl(`${url}${queryParams}`);
}

File: packages/common/src/core/schemas/internal/getLocationOptions.ts [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+        entityType: getTypeFromEntity(entity),
+        location: {
+          type: "custom",
+          spatialReference: defaultExtent.spatialReference,
+        },
+      },
+    ] as IHubLocationOption[]
+  ).map((option) => {
+    // TODO: Verify logic with Ben as code ^^ fails if entity is null
+    // if (!entity && option.location.type === "custom") {
```
juliannemarik, line 55: This is needed for the scenario 1 in the description - the creation workflow (where entity is `null`). If you make line 26 have optional chaining (`entity?.location`), I believe this should all still work?

packages/common/src/core/schemas/internal/getTagItems.ts
```diff
@@ -0,0 +1,48 @@
...
+import { IHubRequestOptions } from "../../../types";
+import { HubEntity } from "../../types/HubEntity";
+import { IUiSchemaComboboxItem } from "../types";
+
+/**
+ * Fetch the entity's org tags (limited to the top 200),
+ * merge with any configured on the entity itself, and convert
+ * into a format that can be consumed by the combobox field
+ *
+ * TODO: the search portion of this util should probably
```
juliannemarik, line 13: This TODO can be removed

packages/common/src/core/schemas/internal/groupsToComboBoxItems.ts
```diff
@@ -0,0 +1,32 @@
+import { IGroup } from "@esri/arcgis-rest-portal";
+import { IUiSchemaComboboxItem } from "../types";
+
+/**
+ * Convert an array of groups into a UiSchemaComboboxItem array
+ */
+export function groupsToComboBoxItems(
```
juliannemarik, line 7: minor, but I feel like this function name is a little misleading since we're not _just_ converting groups to combobox items, we're also performing a "filter" of sorts to only include groups a user can share to

packages/common/src/core/schemas/types.ts
```diff
@@ -72,3 +72,10 @@ export type UiSchemaElementOptions = Pick<
   IUiSchemaElement,
   "scope" | "options"
 >;
+
+export interface IUiSchemaComboboxItem {
```
juliannemarik, line 76: Not part of this work, but we really need a chore to remove the duplicated configuration editor types/interfaces in this file from `opendata-ui` and to update all of the imports to use the ones here.

packages/common/src/core/types/IHubLocation.ts
```diff
@@ -26,3 +27,20 @@ export interface IHubLocation {
   graphic?: any;
   geoJson?: any;
 }
+
+/**
+ * Location options for the location picker
+ */
+export interface IHubLocationOption {
```
juliannemarik, line 34: same here - if we're hoisting this interface here, can we remove it from `opendata-ui` so we don't end up with two out-of-sync instances?

packages/common/src/core/types/IHubProject.ts
```diff
@@ -26,3 +27,16 @@ export enum PROJECT_STATUSES {
...
   complete = "complete",
 }
+
+/**
+ * This type redefines the IHubProject interface in such a way
+ * that it can be consumed by the entity editor.
+ */
+export type IHubProjectEditor = Omit<IHubProject, "extent"> & {
+  extent: IExtent | number[][];
+  // TODO: if this is an ephemeral property, we should prefix with _
```
juliannemarik, line 37: I agree. We would just need to change the reference in the `uiSchema` to include an underscore as well.

packages/common/src/core/types/IHubProject.ts
```diff
@@ -26,3 +27,16 @@ export enum PROJECT_STATUSES {
   onHold = "onHold",
   complete = "complete",
 }
+
+/**
+ * This type redefines the IHubProject interface in such a way
+ * that it can be consumed by the entity editor.
+ */
+export type IHubProjectEditor = Omit<IHubProject, "extent"> & {
+  extent: IExtent | number[][];
```
juliannemarik, line 36: I don't _think_ extent needs to be here anymore?

packages/common/src/projects/HubProject.ts
```diff
@@ -161,6 +161,113 @@ export class HubProject
...
+   * @param i18nScope translation scope to be interpolated into the uiSchema
+   * @param type editor type - corresonds to the returned uiSchema
+   * @param options optional hash of dynamic uiSchema element options
+   */
+  async getEditorConfig(
+    i18nScope: string,
+    type: EditorType
+  ): Promise<IEditorConfig> {
+    // get the options first...
+    // TODO: Move to util fn
```
juliannemarik, line 175: can this TODO be removed? looks like you moved it to a util

packages/common/src/projects/_internal/getProjectEditorConfigOptions.ts
```diff
@@ -0,0 +1,82 @@
...
+import { IHubProject } from "../../core/types/IHubProject";
+import { checkPermission } from "../../permissions/checkPermission";
+
+/**
+ * Construct the Project Editor configuration options
+ * @param entity
+ * @param context
+ * @returns
+ */
+export async function getProjectEditorConfigOptions(
```
juliannemarik, line 20: Similar to what we just did with creating a base `IHubItemEntitySchema` that all entities can spread, I think it might be nice to create a `baseItemEntityConfigOptions` array that all item entities can spread because I don't think any of this is necessarily project-specific.

juliannemarik (commented):
This is all looking really good to me! Left some minor comments. I think getting this logic hoisted over will be extremely helpful as teams start scaffolding their settings panes to include entity-specific editing experiences as well

packages/common/src/core/schemas/internal/getFeaturedImageUrl.ts
dbouwman: I agree - I'll make an issue to try to resolve this - I don't want to add additional concerns / refactors to this PR

packages/common/src/core/schemas/internal/getLocationOptions.ts
dbouwman: Ah - I've eliminated the code path where the entity is `null`. 

When we create a new entity, in order to create the `EntiyEditor` instance, I'm using this..
```
const entity = this.entity || this.getDefaultEntity(this.entityType);
```
Which calls this to construct the minimum viable entity...
```
  private getDefaultEntity(entityType: string): HubEntity {
    let entity;
    switch (entityType) {
      case 'project':
        entity = { type: 'Hub Project' } as HubEntity;
        break;
    }
    return entity;

```
So there is always an entity object - even if it's filled with defaults.

Now that i understand the intent, I'll check for `.id` and set to default if that's not present.

packages/common/src/core/schemas/internal/groupsToComboBoxItems.ts
dbouwman: 100% agree. Will rename to `getSharableGroupsComboBoxItems`

packages/common/src/core/schemas/types.ts
dbouwman: Yes. My next PR to -ui will swap to importing them from hub.js

packages/common/src/core/types/IHubProject.ts
dbouwman: I remove it, and verified in the harness...

packages/common/src/core/types/IHubProject.ts
dbouwman: Done and verified in harness.

packages/common/src/projects/_internal/getProjectEditorConfigOptions.ts
dbouwman: Let's revisit this on a subsequent PR / issue

dbouwman:
Follow on issues; 
- https://devtopia.esri.com/dc/hub/issues/7362
- https://devtopia.esri.com/dc/hub/issues/7363

juliannemarik (approved):
awesome! Looking forward to using this

PR #1128, opened 2023-07-20
feat(hub-common): workspace edit content details

abp6318:
This PR and branch are new and created after the old branch had some funny yalc + build issues.

opendata-ui PR: [#12282](https://github.com/ArcGIS/opendata-ui/pull/12282)

1. Description:
Creating a new view in workspaces for viewing content item's details and editing them.

1. Instructions for testing:
- visit /workspace/content/:id/details
- make edits to properties available
- save edits
- repeat with different documentations and properties

1. Closes Issues: #7207


packages/common/src/core/schemas/getEntityEditorSchemas.ts
```diff
@@ -24,6 +24,10 @@ import {
   PageEditorType,
   PageEditorTypes,
 } from "../../pages/_internal/PageSchema";
+import {
+  ContentEditorType,
```
juliannemarik, line 28: remove duplicated import

packages/common/src/core/schemas/getEntityEditorSchemas.ts
```diff
@@ -32,10 +36,12 @@ import {
 export type EditorType = (typeof validEditorTypes)[number];
 export const validEditorTypes = [
   ...ProjectEditorTypes,
+  ...ContentEditorTypes,
   ...InitiativeEditorTypes,
   ...SiteEditorTypes,
   ...DiscussionEditorTypes,
   ...PageEditorTypes,
+  ...ContentEditorTypes,
```
juliannemarik, line 44: remove duplicated addition of content editor types

packages/common/src/content/_internal/ContentUiSchemaEdit.ts
```diff
@@ -0,0 +1,119 @@
...
+          labelKey: "{{i18nScope}}.fields.title.label",
+          scope: "/properties/name",
+          type: "Control",
+          options: {
+            helperText: {
+              labelKey: "{{i18nScope}}.fields.title.hint",
+            },
+            messages: [
+              {
+                type: "ERROR",
```
sonofflynn89, line 21: Is `ERROR` supposed to be in all caps?

packages/common/src/items/setItemThumbnail.ts
```diff
@@ -13,12 +13,14 @@ export async function setItemThumbnail(
   id: string,
   file: any,
   filename: string,
-  requestOptions: IUserRequestOptions
+  requestOptions: IUserRequestOptions,
+  owner?: string
```
juliannemarik, line 17: Should `owner` really be optional? As it's implemented now, since we don't provide `owner `as a standalone `opt` or on the `item`, the `updateItem` call falls back to using the `authentication.username` which (as we saw) causes false failures.

I'd argue we should make this a required arg; however, if we do that, I understand this is technically a breaking change...although I don't believe it will have far-reaching consequences.

packages/common/test/content/edit.test.ts
```diff
@@ -39,12 +39,19 @@ describe("content editing:", () => {
   });
   describe("update content:", () => {
     it("converts to a model and updates the item", async () => {
-      const getModelSpy = spyOn(modelUtils, "getModel").and.returnValue(
+      // const getModelSpy = spyOn(modelUtils, "getModel").and.returnValue(
```
juliannemarik, line 42: can this be removed?

sonofflynn89 (commented):
Looking good! I have a couple outstanding questions but you are really close!

packages/common/src/content/_internal/ContentUiSchemaEdit.ts
juliannemarik: Yes, that is what the configuration editor expects - it's one of the [`UiSchemaMessageTypes`](https://github.com/ArcGIS/opendata-ui/blob/poc/settings-pane-entity-editor/packages/hub-components/src/components/arcgis-configuration-editor/resources.ts#L18-L22) enums. I realize it's confusing though because there's a lot of inconsistency with the casing of our enums in the config editor 🙈

packages/common/src/models/index.ts
```diff
@@ -159,7 +159,9 @@ export function updateModel(
   return updateItem(opts as IUpdateItemOptions).then(() => {
     // To ensure we have the exact modified timestamp, we need to
     // get the item again
-    return getModel(item.id, requestOptions);
+    return item.data
+      ? getModel(item.id, requestOptions)
+      : getItem(item.id, requestOptions).then((i) => ({ item: i }));
```
sonofflynn89, line 164: Ok, my  comment here wasn't saved, but I am worried about the fact that we are introducing this change. `updateModel`  is used in a variety of places so we need to have a pretty solid reason to change its behavior.

tomwayson: I understand your concern, but I suggested this change. Up until now every item type we've thrown at `updateModel` has not only had `data`, but also known that the data was JSON. With this PR, we are now handling item types like PDF, that have blob data and if you call `getModel` it _always_ tries to fetch the item data and parse it as JSON.

To me the only risk is using the existence if `item.data` as a signal that we should should fetch the data. My assumption is that it will _always_ be truthy for things that have JSON data b/c they will have passed it in even if they did not make any changes. If that assumption is wrong, then we may need to think of another way to tell this fn when to use `getModel`, like adding a `fetchModel = true` to the options that are passed in. That would be safer, but more cumbersome, and I'm not sure it's needed.

tomwayson: Another way to do deal w/ this is to have `getModel()` support types that don't have JSON data. I don't want to get into the business of basing that on `item.type`, but we might be able to do something like having it catch JSON parse errors, but then we'd be suppressing actual error's in the item's data.

@abp6318 - either way, at the very least we need to add a comment here saying that we cannot just call `getModel` as it stands today, b/c the content entity needs to support types that may not have JSON data.

packages/common/src/content/edit.ts
```diff
@@ -86,7 +91,8 @@ export async function updateContent(
   // let resources;
 
   // get the backing item & data
-  const model = await getModel(content.id, requestOptions);
+  const item = await getItem(content.id, requestOptions);
```
tomwayson, line 94: @abp6318 we probably want a similar comment here about not calling `getModel()` b/c some content types have non-JSON data

tomwayson (changes requested):
Great work @aaronpi 

Other than fixing the failing tests, please just add the comments I suggest.

tomwayson (approved):
Well @abp6318, once merges are unblocked, go ahead and merge.

PR #1130, opened 2023-07-24
fix(hub-common): saving sites guarantees a unique slug and prevents field loss in site data

sonofflynn89:
affects: @esri/hub-common

1. Description:
Issue - https://devtopia.esri.com/dc/hub/issues/7337
Fixes a couple saving issues with the `updateSite` logic

### DIFF
--- a/packages/common/src/sites/HubSites.ts
+++ b/packages/common/src/sites/HubSites.ts
@@ -12,7 +12,7 @@ import { applyPermissionMigration } from "./_internal/applyPermissionMigration";
 import { computeProps } from "./_internal/computeProps";
 import { getPropertyMap } from "./_internal/getPropertyMap";
 import { IHubSite } from "../core/types/IHubSite";
-import { constructSlug, setSlugKeyword } from "../items/slugs";
+import { constructSlug, getUniqueSlug, setSlugKeyword } from "../items/slugs";
 import { slugify } from "../utils/slugify";
 import { ensureUniqueDomainName } from "./domains/ensure-unique-domain-name";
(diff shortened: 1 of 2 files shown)


packages/common/src/sites/HubSites.ts
```diff
@@ -309,35 +309,38 @@ export async function updateSite(
   site: IHubSite,
   requestOptions: IHubRequestOptions
 ): Promise<IHubSite> {
+  // verify that the slug is unique, excluding the current site
+  site.slug = await getUniqueSlug(
```
sonofflynn89, line 313: Dave told me that we should use the same slug verification logic as projects, so I introduced it here

mjuniper (approved):

mjuniper (approved):

packages/common/src/sites/HubSites.ts
sonofflynn89: @dbouwman, @mjuniper wasn't aware that sites had / needed slugs. Just to confirm before we merge, can we expect _every_ site to have a slug in `item.properties.slug`, or just enterprise sites?

dbouwman: All "Entity" based items should create a slug, and store it in `item.properties.slug` and a corresponding typekeyword - all done using the same set of helper functions. 

Whether or not anything uses that is a separate question that comes down to how we want urls to work. 

And while it's true that Site's on AGO are unlikely to ever use this property, I'm more inclined to have consistent code for Enterprise and AGO than to have yet-more permutations. :)

dbouwman (approved):
Thanks @sonofflynn89!

PR #1151, opened 2023-08-07
feat: workspace gating

mjuniper:
1. Description: Proposed interim approach for gating workspaces.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/permissions/HubPermissionPolicies.ts
+++ b/packages/common/src/permissions/HubPermissionPolicies.ts
@@ -64,7 +64,7 @@ import { PlatformPermissionPolicies } from "./PlatformPermissionPolicies";
 
 const TempPermissionPolicies: IPermissionPolicy[] = [
   {
-    permission: "hub:workspace:released",
+    permission: "temp:workspace:released",
     subsystems: [],
     assertions: [
       {
--- a/packages/common/src/permissions/HubPermissionPolicies.ts
+++ b/packages/common/src/permissions/HubPermissionPolicies.ts
@@ -62,6 +62,21 @@ import { PlatformPermissionPolicies } from "./PlatformPermissionPolicies";
 //   },
 // ];
 
+const TempPermissionPolicies: IPermissionPolicy[] = [
+  {
+    permission: "hub:workspace:released",
+    subsystems: [],
+    assertions: [
+      {
+        // not prod
+        property: "context:hubUrl",
+        type: "not-ends-with",
+        value: "hub.arcgis.com",
+      },
+    ],
+  },
+];
+
 /**
  * All the permission policies for the Hub
  */
@@ -74,6 +89,7 @@ export const HubPermissionsPolicies: IPermissionPolicy[] = [
   ...GroupPermissionPolicies,
   ...PagePermissionPolicies,
   ...PlatformPermissionPolicies,
+  ...TempPermissionPolicies,
 ];
 
 /**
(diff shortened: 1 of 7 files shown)


packages/common/src/permissions/HubPermissionPolicies.ts
```diff
@@ -62,6 +62,21 @@ import { PlatformPermissionPolicies } from "./PlatformPermissionPolicies";
...
+    assertions: [
+      {
+        // not prod
+        property: "context:hubUrl",
+        type: "not-ends-with",
+        value: "hub.arcgis.com",
+      },
+    ],
+  },
+];
```
mjuniper, line 78: A permission policy that will only pass if context.hubUrl does not end with hub.arcgis.com. Ie it will fail in prod.

packages/common/src/permissions/_internal/checkAssertion.ts
```diff
@@ -48,6 +48,12 @@ export function checkAssertion(
       case "neq":
         response = equalityAssertions(assertion, propValue, val);
         break;
+      case "starts-with":
+      case "ends-with":
+      case "not-starts-with":
+      case "not-ends-with":
+        response = stringAssertions(assertion, propValue, val);
+        break;
```
mjuniper, line 56: I believe there was no other way to do this. Since the value of hubUrl will contain a `:`, the `neq` assertion type understands it to not be a literal value but an instruction to lookup another value on the entity/context hash.

packages/common/src/permissions/_internal/checkAssertion.ts
```diff
@@ -169,6 +175,34 @@ function equalityAssertions(
...
+  ) {
+    response = "assertion-failed";
+  } else if (assertion.type === "ends-with" && !propValue.endsWith(val)) {
+    response = "assertion-failed";
+  } else if (assertion.type === "not-ends-with" && propValue.endsWith(val)) {
+    response = "assertion-failed";
+  }
+  return response;
+}
+
```
mjuniper, line 205: String starts with, ends with, not starts with, & not ends with assertions. See above for why this was needed.

packages/common/src/projects/_internal/ProjectBusinessRules.ts
```diff
@@ -36,7 +36,7 @@ export const ProjectCapabilityPermissions: ICapabilityPermission[] = [
   {
     entity: "project",
     capability: "settings",
-    permissions: ["hub:project:owner"],
+    permissions: ["hub:project:owner", "hub:workspace:released"],
```
mjuniper, line 39: Wrapping it up in a project capability.

So, we can use `checkCapability('settings', context, projectEntity)`.

Or, in the context of hub components WorkspaceLinkDefinitions, we would do something like:
```js
{
    entity: 'project',
    pane: 'settings',
    component: 'arcgis-hub-entity-settings',
    i18nLabel: 'panes.settings',
    icon: PANE_ICONS.settings,
    capabilities: ['settings'],
  }
```

Finally, we can do `checkPermission('hub:workspace:released', context)`.

packages/common/src/permissions/HubPermissionPolicies.ts
dbouwman: Could we call this `flag:workspace:released` and `FlagPermissionPolicies`? or leave it as `TempPermissionPolicies` but change the string to `temp:workspace:released` ?

++ on making use of the existing assertion infrastructure :)

packages/common/src/projects/_internal/ProjectBusinessRules.ts
dbouwman: using `temp:` of `flag:` would make this easier to reason about what will be removed down the line...

dbouwman (approved):
I think this is a good solution that works for now, and should be workable moving forward as well. Thanks!

PR #1155, opened 2023-08-09
fix: user selects none; location disappears correctly

abp6318:
### Important
Make sure you have your opendata-ui set to https://github.com/ArcGIS/opendata-ui/pull/12378 until it merges, I added a new translation this morning.

1. Description: 
- If a user selects "none" in the location picker, it no longer appears in the /about view
- Location picker options in workspace content now properly shows two options
  - This will be further built off of in future issues and PRs that allow for a bounding box option as discussed with Tom, Dave, and Ben on Aug 10, 2023

1. Instructions for testing:
- Clone this PR and PR with updated translations: https://github.com/ArcGIS/opendata-ui/pull/12378 
- Copy the hubjs dist folder over
- `yarn hc start` -- switch back and forth between "none" and "Item's Current Extent" in content's location picker
- Double check that I haven't changed other workspace entities like projects

1. Closes Issues: #7208

### DIFF
--- a/packages/common/src/content/edit.ts
+++ b/packages/common/src/content/edit.ts
@@ -104,6 +104,14 @@ export async function updateContent(
   // we are not attempting to handle "concurrent edit" conflict resolution
   // but this is where we would apply that sort of logic
   const modelToUpdate = mapper.entityToStore(content, model);
+
+  // prevent map from displaying when boundary is 'none'
+  const location = modelToUpdate.item.properties?.location;
+  if (location) {
+    modelToUpdate.item.properties.boundary =
+      location?.type === "none" ? "none" : "item";
+  }
+
   // TODO: if we have resources disconnect them from the model for now.
   // if (modelToUpdate.resources) {
   //   resources = configureBaseResources(
--- a/packages/common/src/content/edit.ts
+++ b/packages/common/src/content/edit.ts
@@ -20,6 +20,7 @@ import { getPropertyMap } from "./_internal/getPropertyMap";
 import { cloneObject } from "../util";
 import { IModel } from "../types";
 import { computeProps } from "./_internal/computeProps";
+import { getProp } from "..";
 
 // TODO: move this to defaults?
 const DEFAULT_CONTENT_MODEL: IModel = {
@@ -106,11 +107,9 @@ export async function updateContent(
   const modelToUpdate = mapper.entityToStore(content, model);
 
   // prevent map from displaying when boundary is 'none'
-  const location = modelToUpdate.item.properties?.location;
-  if (location) {
-    modelToUpdate.item.properties.boundary =
-      location?.type === "none" ? "none" : "item";
-  }
+  const itemType = getProp(modelToUpdate, "item.properties.location.type");
+  modelToUpdate.item.properties.boundary =
+    itemType === "none" ? "none" : "item";
 
   // TODO: if we have resources disconnect them from the model for now.
   // if (modelToUpdate.resources) {
--- a/packages/common/src/content/edit.ts
+++ b/packages/common/src/content/edit.ts
@@ -5,7 +5,7 @@ import {
   getItem,
   removeItem,
 } from "@esri/arcgis-rest-portal";
-import { IHubContent, IHubContentEditor, IHubEditableContent } from "../core";
+import { IHubContentEditor, IHubEditableContent } from "../core";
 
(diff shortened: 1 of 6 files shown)


packages/common/src/content/edit.ts
```diff
@@ -104,6 +105,12 @@ export async function updateContent(
   // we are not attempting to handle "concurrent edit" conflict resolution
   // but this is where we would apply that sort of logic
   const modelToUpdate = mapper.entityToStore(content, model);
+
+  // prevent map from displaying when boundary is 'none'
+  const itemType = getProp(modelToUpdate, "item.properties.location.type");
+  modelToUpdate.item.properties.boundary =
+    itemType === "none" ? "none" : "item";
```
tomwayson, line 112: `locationType` is more accurate

```suggestion
  const locationType = getProp(modelToUpdate, "item.properties.location.type");
  modelToUpdate.item.properties.boundary =
    locationType === "none" ? "none" : "item";
```

packages/common/test/content/edit.test.ts
```diff
@@ -70,6 +70,52 @@ describe("content editing:", () => {
...
+        createdDate: new Date(1595878748000),
+        createdDateSource: "item.created",
+        updatedDate: new Date(1595878750000),
+        updatedDateSource: "item.modified",
+        thumbnailUrl: "",
+        permissions: [],
+        schemaVersion: 1,
+        canEdit: false,
+        canDelete: false,
+        location: { type: "item" },
```
tomwayson, line 118: we prob want to add expectations on `modelToUpdate.item.properties.boundary` for this test to be meaningful

tomwayson (commented):
As we discussed, please add a commit where get a distinct location editor config for content.

packages/common/src/core/schemas/internal/getLocationOptions.ts
```diff
@@ -22,9 +23,69 @@ export async function getLocationOptions(
...
+  ).map((option) => {
+    // If this is a new entity, select the custom option by default
+    if (!entity.id && option.location.type === "custom") {
+      option.selected = true;
+    } else if (entity.id && !location && option.location.type === "none") {
+      option.selected = true;
+    } else if (location?.type === option.location.type) {
+      option.location = location;
+      option.selected = true;
+    }
```
sonofflynn89, line 74: It took me a while to figure out why you had these if statements. I think my biggest hangup is the dual purpose of checking for the option's type AND seeing if a fallback is necessary. The re-assigning of `option.location` also is oddly placed, since the `custom` option is the only one with a location to begin with.

Don't take too long, but take a crack at simplifying this. Try to break up these 3 steps:
- Determining which location type is selected, including fallbacks
- Setting the `custom` option's `location` field
- Actually marking an option as selected

sonofflynn89 (commented):
Great work! Just a couple comments

sonofflynn89 (approved):
Great work with all this Aaron! Thank you so much for taking this on 🙏

packages/common/src/core/schemas/internal/getLocationOptions.ts
tomwayson: I agree that coupling those operations do make this a little hard to follow, but there is a bigger issue that needs to be resolved first, which is that we need to base the options and selected state on the item's extent and `properties.boundary` and **not**  the org's extent and `properties.location`.

tomwayson (changes requested):
All in all this is good, we just need to update the logic to work with existing content items. Let's pair on this.

packages/common/src/core/schemas/internal/getLocationOptions.ts
```diff
@@ -37,50 +48,33 @@ export async function getLocationOptions(
...
-  );
-  const location: IHubLocation = entity.location;
-  return (
-    [
-      {
-        label: "{{shared.fields.location.none:translate}}",
-        location: { type: "none" },
+  const defaultExtent: IExtent = getItemExtent(entity.extent);
+  const defaultExtentIsBBox = isBBox(entity.extent);
+  const boundary = entity.boundary;
```
tomwayson, line 55: is `entity.boundary` a thing? I was expecting that you'd have to reach into `item.properties` to get `boundary`?

tomwayson (approved):
Good work. As long as content entities have a `.boundary` the logic in this is sound.

PR #1166, opened 2023-08-17
feat(hub-common): add user membership enrichment and have updateSite persist catalog changes

sonofflynn89:
1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/7177

### DIFF
--- a/packages/common/src/sites/_internal/convertCatalogToLegacyFormat.ts
+++ b/packages/common/src/sites/_internal/convertCatalogToLegacyFormat.ts
@@ -1,0 +1,36 @@
+import { getProp } from "../../objects";
+import { IPredicate, IQuery } from "../../search/types/IHubCatalog";
+import { IModel } from "../../types";
+import { cloneObject } from "../../util";
+
+// TODO: consider exposing this so it can be used on the front end
+function getGroupPredicate(scope: IQuery): IPredicate {
+  const isGroupPredicate = (predicate: IPredicate) => !!predicate.group;
+  const groupFilter = scope.filters.find((f) =>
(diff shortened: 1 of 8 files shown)


packages/common/src/sites/_internal/convertCatalogToLegacyFormat.ts
```diff
@@ -0,0 +1,36 @@
...
+export function convertCatalogToLegacyFormat(
+  modelToUpdate: IModel,
+  currentModel: IModel
+): IModel {
+  const updatedModel = cloneObject(modelToUpdate);
+  const updatedItemScope = getProp(modelToUpdate, "data.catalog.scopes.item");
+  const updatedGroupPredicate = getScopeGroupPredicate(updatedItemScope);
+  if (updatedGroupPredicate) {
+    const updatedLegacyCatalog = cloneObject(currentModel.data.catalog);
+    // TODO: Do we need to worry about whether predicate.group is an array or not? This applies here and in opendata-ui
```
dbouwman, line 24: fwiw - It could also be a `MatchOptions` ({any: [], all: [], not: []}) - but that's highly unlikely, and I'm not sure how we'd express that in the legacy format (other than `any` which is how it's treated at the api level)

sonofflynn89: @dbouwman that's a good point! Unfortunately, a lot of the search branches developed by others (Viv, Josh, etc.) also don't take into account the `MatchOptions` interface as it isn't always feasibly translated. Once things slow down we should definitely take a high level overview of what we want to do long term with our search interfaces. I'll create a tech issue to review this.

PR #1192, opened 2023-08-28
feat: adding license picker to content workspace

abp6318:
1. Description: Add license picker to content workspaces

**OD-UI PR:** https://github.com/ArcGIS/opendata-ui/pull/12448

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/7209


packages/common/src/content/_internal/computeProps.ts
```diff
@@ -41,6 +41,7 @@ export function computeProps(
   }
   // thumbnail url
   content.thumbnailUrl = getItemThumbnailUrl(model.item, requestOptions, token);
+  content.licenseInfo = model.item.licenseInfo || "";
```
sonofflynn89, line 44: Is there a reason you are defaulting to `""`?

sonofflynn89 (commented):
Looking good on this hub.js end! Just one question

packages/common/src/content/_internal/computeProps.ts
abp6318: `licenseInfo` can sometimes be null (I'm pretty sure)

sonofflynn89: For sure, but since `null` and `""` are both falsy it shouldn't matter much. Unless we're getting errors because something is checking string length, we should be able to leave things as-is and not provide the fallback.

sonofflynn89:
@tomwayson I notice that @abp6318 hasn't been using `npm run c` in his commits and instead is manually doing commits like `fix: change behavior...` (omitting the scope). It looks like everything still works with Semantic, but are there any significant drawbacks to doing this? I know Josh has done this a few times as well.

abp6318:
> @tomwayson I notice that @abp6318 hasn't been using `npm run c` in his commits and instead is manually doing commits like `fix: change behavior...` (omitting the scope). It looks like everything still works with Semantic, but are there any significant drawbacks to doing this? I know Josh has done this a few times as well.

@sonofflynn89 oops, I didn't know that existed. Will use that from now on to prevent any possible issues.

sonofflynn89 (approved):
Looks good to me! Please don't merge though until we get confirmation from Tom on the commit syntax.

sonofflynn89:
> oops, I didn't know that existed. Will use that from now on to prevent any possible issues.

@abp6318 it's all good! I'm sorry we didn't tell you about it sooner 😅

PR #1196, opened 2023-08-29
fix: account for index values that are 0 in card view models

juliannemarik:
### Description
Currently if a search result has an index of 0, it doesn't get added to the card view model because 0 is a `falsey` value. Instead of checking the truthiness of `searchResult.index`, we need to check whether `searchResult.index` is a number. If it is, we add it to the view model. 


benstoltz (approved):
My only recommendation is to alter tests to see if you can test the permutations for this case. (0, invalid, and non zero valid)

PR #1197, opened 2023-08-30
feat: add invite users to hub group

benstoltz:
1. Description: This copies over the whole add/invite/email logic flow from teams into hub-common groups. The only net new function is `addOrInviteUsersToHubGroup` which is in `HubGroups.ts`. Everything else has been in use for two years now.

1. Instructions for testing:


packages/common/src/groups/index.ts
```diff
@@ -3,3 +3,8 @@ export * from "./add-users-workflow";
 export * from "./types";
 export * from "./HubGroups";
 export * from "./HubGroup";
+// TODO: The below are being used in hub-teams. When we deprecate that package we can move
+// The below into _internal and remove the exports from here.
+export * from "./autoAddUsers";
```
juliannemarik, line 8: two questions here: 
1. I think I'm confused why these need to be exported and not `_internal` from the start here. If you just copied over all of this functionality from `hub-teams`, isn't `hub-teams` still just using its own internal versions of these?
2. I do think we'll need to export `addOrInviteUsersToGroup` and `addOrInviteUserToGroup` so we can use those from `opendata-ui`

juliannemarik (approved):
Assuming all of this functionality is simply being copied from `hub-teams`, this looks good to me

PR #1199, opened 2023-09-01
feat(hub-common): add unified "links" hash as an optional property for all entities

sonofflynn89:
1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/7521

Moves the "links" property into IHubEntityBase and adds "layoutRelative" as an option.

Before this PR,  "links" was part of IHubSearchResult. Since IHubSearchResult extends IHubEntityBase, no
change is needed on the search side. However, by moving "links" to IHubEntityBase, we are laying the
foundation for IHubSearchResults and IHubEntitys to have the same link structure.

I started by implementing basic links for IHubPage, IHubSite and IHubEditableContent.

Longer term, we'll want to figure out some basic questions like:
- should thumbnail have the token appended?
- should link generation be abstracted into a function or be created ad-hoc in each entity's `computeProps` function?
- should IHubSearchResults always populate the same links as fetchHubEntity? How does ^^^ relate to that goal?
- What's our migration strategy for removing `thumbnailUrl` and moving to `links.thumbnail`?
- Should `IHubItemEntity.url` actually be the `item.url`? How do we migrate usage to `links.self` instead?

### DIFF
--- a/packages/common/src/content/_internal/computeProps.ts
+++ b/packages/common/src/content/_internal/computeProps.ts
@@ -1,11 +1,15 @@
 import { IRequestOptions } from "@esri/arcgis-rest-request";
 import { UserSession } from "@esri/arcgis-rest-auth";
 import { getItemThumbnailUrl } from "../../resources";
-import { IHubEditableContent, IHubLocation } from "../../core";
 import { IModel } from "../../types";
 import { bBoxToExtent, extentToPolygon, isBBox } from "../../extent";
 import { IExtent } from "@esri/arcgis-rest-types";
 import Geometry = __esri.Geometry;
+import { getItemHomeUrl } from "../../urls/get-item-home-url";
+import { getHubRelativeUrl } from "./internalContentUtils";
+import { IHubLocation } from "../../core/types/IHubLocation";
+import { IHubEditableContent } from "../../core/types/IHubEditableContent";
+import { getRelativeWorkspaceUrl } from "../../core/getRelativeWorkspaceUrl";
 
 // if called and valid, set 3 things -- else just return type custom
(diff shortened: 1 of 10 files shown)


packages/common/src/core/types/IHubEntityBase.ts
```diff
@@ -48,4 +64,37 @@ export interface IHubEntityBase {
...
+     * Url to Thumbnail. May not include a token
+     * TODO: Why should we not include the token?
+     */
+    thumbnail?: string;
+    /**
+     * Url to the entities canonical "self"
+     * For Items/Groups/Users, this will be the Home App url
+     * For other entities, it will be the canonical url
+     */
+    self: string;
```
sonofflynn89, line 82: Should we leave this as required?

packages/common/src/content/_internal/computeProps.ts
```diff
@@ -40,7 +44,19 @@ export function computeProps(
     token = session.token;
   }
   // thumbnail url
-  content.thumbnailUrl = getItemThumbnailUrl(model.item, requestOptions, token);
+  const thumbnailUrl = getItemThumbnailUrl(model.item, requestOptions, token);
+  // TODO: Remove this once opendata-ui starts using `links.thumbnail` instead
+  content.thumbnailUrl = thumbnailUrl;
```
tomwayson, line 49: can we make the top level `thumnailUrl` a computed prop?

packages/common/src/core/types/IHubEntityBase.ts
tomwayson: I think so

tomwayson (approved):
get'r dun

PR #1206, opened 2023-09-11
feat: add workspaceRelative to enrichGroupSearchResult

benstoltz:
1. Description: We weren't determining the workspaceRelative link for groups. This implements that.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/groups/HubGroups.ts
+++ b/packages/common/src/groups/HubGroups.ts
@@ -85,10 +85,7 @@ export async function enrichGroupSearchResult(
   result.links.thumbnail = getGroupThumbnailUrl(requestOptions.portal, group);
   result.links.self = getGroupHomeUrl(result.id, requestOptions);
   result.links.siteRelative = `/teams/${result.id}`;
-  result.links.workspaceRelative = getRelativeWorkspaceUrl(
-    result.type,
-    result.id
-  );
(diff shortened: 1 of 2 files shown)


packages/common/src/groups/HubGroups.ts
```diff
@@ -84,6 +85,10 @@ export async function enrichGroupSearchResult(
   result.links.thumbnail = getGroupThumbnailUrl(requestOptions.portal, group);
   result.links.self = getGroupHomeUrl(result.id, requestOptions);
   result.links.siteRelative = `/teams/${result.id}`;
+  result.links.workspaceRelative = getRelativeWorkspaceUrl(
+    result.type,
```
sonofflynn89, line 89: In this case you should just hardcode `"group"` since this function is only used for group search results. I'm not sure what `group.type` will give you, but I suspect it won't work. The reason typescript doesn't give you an error is that the first argument is labeled as a `string` instead of an `HubEntityType` (something another PR should definitely consider changingg

benstoltz: Updated it to `"Group"` as it would expect.

sonofflynn89: Good catch!

sonofflynn89 (approved):
Thanks for this!

PR #1209, opened 2023-09-12
feat(hub-common): add content settings schemas and derive props from server definitions

sonofflynn89:
1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/7210
Adds the content settings schema. This included adding the ability to fetch / update the underlying service definition for items that represent hosted feature services. All work was done within the `fetchHubEntity` and `updateHubEntity` flows

### DIFF
--- a/packages/common/src/content/_internal/hostedServiceUtils.ts
+++ b/packages/common/src/content/_internal/hostedServiceUtils.ts
@@ -1,0 +1,75 @@
+import { IFeatureServiceDefinition } from "@esri/arcgis-rest-feature-layer";
+import { IItem } from "@esri/arcgis-rest-portal";
+import { IHubEditableContent } from "../../core/types/IHubEditableContent";
+
+export function isHostedFeatureService(
+  content: IHubEditableContent | IItem
+): boolean {
+  return (
+    content.type === "Feature Service" &&
+    content.typeKeywords.includes("Hosted Service")
+  );
+}
+
+export enum ServiceCapabilities {
+  EXTRACT = "Extract",
+}
+
+/**
+ * Returns a whether a service has a capability
+ * @param capability
+ * @param serviceDefinition
+ *
+ * @returns {boolean}
+ */
+export function hasCapability(
+  capability: ServiceCapabilities,
+  serviceDefinition: Partial<IFeatureServiceDefinition>
+) {
+  const capabilities = (serviceDefinition.capabilities || "").split(",");
+  return capabilities.includes(capability);
+}
+
+/**
+ * Toggles a single capability on a given feature service.
+ * Returns a service definition object with updated capabilities
+ * @param capability capability to toggle
+ * @param serviceUrl url of service to modify
+ * @param serviceDefinition current definition of the service
+ *
+ * @returns updated definition
+ */
+export function toggleCapability(
+  capability: ServiceCapabilities,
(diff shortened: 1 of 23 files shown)


packages/common/src/content/_internal/hostedServiceUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Returns a whether a service has a capability
+ * @param capability
+ * @param serviceDefinition
+ *
+ * @returns {boolean}
+ */
+export function hasCapability(
```
tomwayson, line 33: I wonder if this should be `hasServiceCapability()` in order to avoid confusion w/ other capabilities.

tomwayson: never mind. I think this is fine given the other fns.

packages/common/src/content/_internal/hostedServiceUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+): boolean {
+  return isHostedFeatureService(content.type, content.typeKeywords);
+}
+
+function isHostedFeatureService(
+  type: string,
+  typeKeywords: string[] = []
+): boolean {
+  return type === "Feature Service" && typeKeywords.includes("Hosted Service");
+}
```
tomwayson, line 20: Did you copy this from somwhere? Seems like we use different logic elsewhere in the app:

https://github.com/ArcGIS/opendata-ui/blob/05196f9f840d820015543ab3d76800e7f97685ef/packages/opendata-ui/app/utils/hosted-downloads.js#L120-L121

Which relies on this function that I copied from the JSAPI:

https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/esri/layers.js/#L187

Either way, we should try and consolidate the logic into a hub.js fn that is expored and re-used throughout.

tomwayson (approved):
LGTM - one question about hosted FS logic.

packages/common/src/content/_internal/hostedServiceUtils.ts
sonofflynn89: @tomwayson this was the logic that Paul Pederson gave us in [this comment thread]([this comment thread](https://teams.microsoft.com/l/message/19:4e25c74df327460489119a58a2c86105@thread.skype/1691441197342?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=7711b3bb-549e-42dd-af7b-3688b395b0c2&parentMessageId=1691441197342&teamName=ArcGIS%20Online%20Development&channelName=General&createdTime=1691441197342&allowXTenantAccess=false)). Apparently this check is what they use on the home app when deciding whether or not to show the `export data` button on the item settings page.

> Either way, we should try and consolidate the logic into a hub.js fn that is expored and re-used throughout.

I 100% agree. The `isHostedFeatureServiceItem()` helper should be able to accomplish what the old logic in opendata-ui is currently doing.

packages/common/src/content/_internal/hostedServiceUtils.ts
sonofflynn89: @tomwayson I agree with your original sentiments. I'll be a small change.

PR #1220, opened 2023-09-18
feat: add option to enrich an entity 

juliannemarik:
[7621](https://devtopia.esri.com/dc/hub/issues/7621)

### Description
- make `toEditor` in `IWithEditorBehavior` and all associated class methods async
- add new arg, `include`, to the `toEditor` class method
- add `enrichEntity` util that delegates to helper functions to add async enrichments to the entity. For now, only a followers group enrichment is supported. `enrichEntity` gets called from all `toEditor` class methods with the array of `include` strings provided

### DIFF
--- a/packages/common/src/content/HubContent.ts
+++ b/packages/common/src/content/HubContent.ts
@@ -15,7 +15,7 @@ import { IEntityEditorContext } from "../core/types/HubEntityEditor";
 import { cloneObject } from "../util";
 import { editorToContent } from "./edit";
 import { ContentEditorType } from "./_internal/ContentSchema";
-import { enrichEntity } from "../core/schemas/internal/enrichEntity";
+import { enrichEntity } from "../core/enrichEntity";
 
 export class HubContent
   extends HubItemEntity<IHubEditableContent>
--- a/packages/common/src/content/HubContent.ts
+++ b/packages/common/src/content/HubContent.ts
@@ -15,6 +15,7 @@ import { IEntityEditorContext } from "../core/types/HubEntityEditor";
 import { cloneObject } from "../util";
 import { editorToContent } from "./edit";
 import { ContentEditorType } from "./_internal/ContentSchema";
+import { enrichEntity } from "../core/schemas/internal/enrichEntity";
 
 export class HubContent
   extends HubItemEntity<IHubEditableContent>
@@ -110,13 +111,20 @@ export class HubContent
    * @returns
    */
   async toEditor(
-    editorContext: IEntityEditorContext = {}
+    editorContext: IEntityEditorContext = {},
+    include: string[] = []
   ): Promise<IHubContentEditor> {
-    // Cast the entity to it's editor
-    const editor = cloneObject(this.entity) as IHubContentEditor;
+    // 1. optionally enrich entity and cast to editor
+    const editor = include.length
+      ? ((await enrichEntity(
+          cloneObject(this.entity),
+          include,
+          this.context.hubRequestOptions
+        )) as IHubContentEditor)
+      : (cloneObject(this.entity) as IHubContentEditor);
 
-    // Add other transforms here...
-    // NOTE: Be sure to add tests for any transforms in the test suite
+    // 2. Apply transforms to relevant entity values so they
+    // can be consumed by the editor
     return editor;
   }
 
--- a/packages/common/src/content/HubContent.ts
+++ b/packages/common/src/content/HubContent.ts
@@ -109,7 +109,9 @@ export class HubContent
    * @param editorContext
    * @returns
    */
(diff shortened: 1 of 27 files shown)


packages/common/src/core/schemas/internal/enrichEntity.ts
```diff
@@ -0,0 +1,87 @@
+import { IGroup } from "@esri/arcgis-rest-types";
```
juliannemarik, line 1: wondering where this should live - should it live here within the `schemas` directory and only be leveraged in the context of `IWithEditorBehavior`, or should it live up a level and be a util that can be called from anywhere to enrich an entity with various information?

packages/common/src/core/schemas/internal/enrichEntity.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * This function delegates out to other helper functions
+ * to fetch the entity enrichments.
+ */
+const fetchEnrichments = async (
+  entity: HubEntity,
+  enrichments: EntityEditorEnrichment[],
+  requestOptions?: IHubRequestOptions
+): Promise<IEntityEditorEnrichments> => {
+  const operations = enrichments.reduce((ops, enrichment) => {
```
dbouwman, line 59: Maybe add a comment in here re: why we chose not to do the whole OperationPipeline thing and/or note that it as intentional

packages/common/src/core/schemas/internal/enrichEntity.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+/**
+ * fetch the entity's followers group
+ */
+const fetchFollowersGroupEnrichment = async (
+  entity: HubEntity,
+  requestOptions: IHubRequestOptions
+): Promise<IGroup> => {
+  const followersGroupId = getProp(entity, "followersGroupId");
```
dbouwman, line 85: what should this do it `followersGroupId` is null? letting getGroup throw in the middle of this will be messy

dbouwman (approved):
This all looks good to me - added a few comments: re adding more comments in the code, but otherwise 👍🏻. I think Tom should weigh in re: alignment / connection to the content enrichments - we want this to be conceptually the same, and even better if it could interoperate.

packages/common/src/content/HubContent.ts
```diff
@@ -109,12 +110,21 @@ export class HubContent
...
-    const editor = cloneObject(this.entity) as IHubContentEditor;
-
-    // Add other transforms here...
-    // NOTE: Be sure to add tests for any transforms in the test suite
+  async toEditor(
+    editorContext: IEntityEditorContext = {},
+    include: string[] = []
+  ): Promise<IHubContentEditor> {
+    // 1. optionally enrich entity and cast to editor
+    const editor = include.length
```
tomwayson, line 118: I'd like for @sonofflynn89 to take a look at this (after you merge) when he gets back and consider whether or not it would make sense to use this as the point to fetch the server enrichment for feature service content.

packages/common/src/initiatives/HubInitiative.ts
```diff
@@ -253,11 +254,22 @@ export class HubInitiative
...
    */
-  toEditor(editorContext: IEntityEditorContext = {}): IHubInitiativeEditor {
-    // Cast the entity to it's editor
-    const editor = cloneObject(this.entity) as IHubInitiativeEditor;
+  async toEditor(
+    editorContext: IEntityEditorContext = {},
+    include: string[] = []
+  ): Promise<IHubInitiativeEditor> {
+    // 1. optionally enrich entity and cast to editor
+    const editor = include.length
```
tomwayson, line 262: I was sort of expecting to see each entity type provide it's own list of default `include`s. So the default for initiatives (or sites) would be `['followersGroup']`. That way the default behavior of `initiativeEntity.toEditor()` would be to include the followers group information. You could override that default behavior by passing `[]` as the `include` argument.

tomwayson (approved):
This looks good.

It will be good to have this second point in the workflow to fetch enrichments. One thing to watch out for is if devs are confused as to whether it should be done in the initial fetch, or here.

packages/common/src/initiatives/HubInitiative.ts
juliannemarik: Leaving this here for posterity even though we discussed in component standup: rather than having the entities themselves define defaults, defaults are being defined/passed in at the ui level (e.g. from the `arcgis-hub-entity-editor`). This is because the defaults differ depending on the ui-context we're in (not necessarily the entity) - on the followers pane (`editorType = "hub:<entity>:followers"`), we'll always want the `followersGroup` enrichment, but we don't need this on the details pane (`editorType = "hub:<entity>:edit"`)

PR #1253, opened 2023-10-02
feat: scaffold solution template entity

juliannemarik:
### Description:
Scaffolds the Hub Solution Template entity

note: we made the decision to simply call this entity a "Hub Template"/`type = "template"` rather than "Hub Solution Template"/`type = "solutionTemplate"`


packages/common/src/templates/HubTemplate.ts
```diff
@@ -0,0 +1,225 @@
...
+  deleteTemplate,
+  editorToTemplate,
+  updateTemplate,
+} from "./edit";
+import { fetchTemplate } from "./fetch";
+import { IEntityEditorContext } from "../core/types";
+import { enrichEntity } from "../core/enrichEntity";
+import { cloneObject } from "../util";
+
+/** Hub Template Class */
```
dbouwman, line 19: Maybe add more explanation here that this is used for all Solution items despite the Hub team not "owning" that item type. Main goal is to facilitate editing the meta info (the "item"), managing sharing etc

packages/common/src/templates/HubTemplate.ts
```diff
@@ -0,0 +1,225 @@
...
+   * @param context
+   * @param save
+   */
+  static async create(
+    partialTemplate: Partial<IHubTemplate>,
+    context: IArcGISContext,
+    save: boolean = false
+  ): Promise<HubTemplate> {
+    const pojo = this.applyDefaults(partialTemplate, context);
+    const instance = HubTemplate.fromJson(pojo, context);
```
dbouwman, line 64: Should this throw if we don't want to allow creating them from Hub?

packages/common/src/templates/HubTemplate.ts
```diff
@@ -0,0 +1,225 @@
...
+  /** Save the HubTemplate to the backing store */
+  async save(): Promise<void> {
+    if (this.isDestroyed) {
+      throw new Error("HubTemplate is already destroyed.");
+    }
+
+    // 1. create or update
+    this.entity = this.entity.id
+      ? await updateTemplate(this.entity, this.context.userRequestOptions)
+      : await createTemplate(this.entity, this.context.userRequestOptions);
```
dbouwman, line 207: Similar to above - should we create new Solution items? If we allow it, someone will do it :)

packages/common/src/templates/_internal/TemplateSchema.ts
```diff
@@ -0,0 +1,18 @@
+import { IConfigurationSchema } from "../../core";
+import { HubItemEntitySchema } from "../../core/schemas/shared/HubItemEntitySchema";
+
+export type TemplateEditorType = (typeof TemplateEditorTypes)[number];
+
+export const TemplateEditorTypes = ["hub:template:edit"] as const;
+
+export const TemplateSchema: IConfigurationSchema = {
+  ...HubItemEntitySchema,
```
dbouwman, line 9: Really like how these are getting quite concise :)

packages/common/src/templates/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * is supposed to become when activated. This util
+ * extracts the expected item type from the
+ * hubSolutionType typeKeyword on the solution item
+ *
+ * We use this in various UIs to show the expected
+ * type rather than "Solution"
+ *
+ * @param template
+ */
+export const getActivatedTemplateType = (template: IItem): string => {
```
dbouwman, line 16: Wonder if we should call this `getDeployedTemplateType` ... or something using the term "deployed" vs "Activated", as it's more aligned with Solutions terminology (I believe... at least they add a Deployed keyword...)

packages/common/src/templates/HubTemplate.ts
juliannemarik: I was considering this as well. I think that's a good call. I've updated the underlying `createTemplate` function to throw an error and added comments to the class methods that call that function letting users know we don't currently support creation

jordantsanz (approved):
Looks good, nice work! 🚀

PR #1277, opened 2023-10-16
feat(): add structure for getting card editor schemas

jordantsanz:
1. Description:
Adds getCardEditorSchemas pattern, starting with stat card.

1. Instructions for testing:

1. Closes Issues: [#6613](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/6613)


packages/common/src/core/schemas/internal/ConfigurableEntity.ts
```diff
@@ -2,3 +2,8 @@
  * Generic interface for a Hub Entity
  */
 export type ConfigurableEntity = Record<string, any>;
```
juliannemarik, line 4: I don't think this `type` accurately represents what we'd expect. I think we should make it more explicit: 

```ts
export type ConfigurableEntity = HubEntity & {};
```

And if there are additional properties we need to pass in with the entity, we can add them to the intersection here. 

This change would require small tweaks in the `getFeaturedImageUrl`, `getLocationOptions`, and `getLocationExtent` internal utils, but I think that's okay. I think you could do one of two things: 

```ts
// getFeaturedImageUrl
export function getFeaturedImageUrl(
  entity: Exclude<ConfigurableEntity, IHubGroup>, // fixes the type error
  ...
) {...}
```
OR

```ts
// getFeaturedImageUrl
export function getFeaturedImageUrl(
  entity: ConfigurableEntity,
  ...
) {
  const featuredImageUrl = getProp(entity, "view.featuredImageUrl") // getProp avoids the type error
  ...
}
```

packages/common/src/core/schemas/internal/ConfigurableEntity.ts
```diff
@@ -2,3 +2,8 @@
...
+
+/**
+ * Generic interface for a layout card
+ */
+export type ConfigurableCard = Record<string, any>;
```
juliannemarik, line 9: Also thinking we should make this type more explicit. For now it would just be: 

```ts
export type ConfigurableCard = {
  themeColors: string[]
}
```

And we can add additional card config properties as we need.

juliannemarik: Also wondering if this should be in its own `ConfigurableCard.ts` file rather than bundled into this file?

packages/common/src/core/schemas/internal/ConfigurableEntity.ts
```diff
@@ -2,3 +2,8 @@
  * Generic interface for a Hub Entity
  */
 export type ConfigurableEntity = Record<string, any>;
+
+/**
+ * Generic interface for a layout card
```
juliannemarik, line 7: I think we could make these descriptions a bit more clear. It's honestly a bit difficult for me to even describe what they are, but I don't think this is necessarily accurate. Basically this is an interface defining what dynamic config options we can pass into the `getEditorConfig` function to help us build the `schema`/`uiSchema`.

juliannemarik (commented):
Overall this is looking great. Left some comments and questions for consideration. I still have some uncertainties myself, so happy to discuss if that would be helpful.

packages/common/src/core/schemas/internal/metrics/interfaces.ts
```diff
@@ -0,0 +1,62 @@
+export enum ALIGNMENTS {
+  start = "start",
+  center = "center",
+  end = "end",
+}
+export enum CORNERS {
```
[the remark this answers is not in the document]
jordantsanz: Makes sense...I think we can move them up into a generic types file.

I feel like `/schemas` might be a good place for them...I would see us using them in the `schema` if anything, listing properties as enums -- we'll also probably use them in the actual ui (like in the `arcgis-stat-card` component), but who knows what other components in the ui will end up using them...so having them in a central place feels necessary. 

Since we need them to be centralized, and we know we'll share them in the schema before using them in each component as well, I feel like there's a somewhat solid case for them to live in `schemas/types.ts`. Thoughts?

packages/common/src/core/schemas/internal/getEditorSchemas.ts
```diff
@@ -68,7 +74,7 @@ export async function getEntityEditorSchemas(
       }[type as SiteEditorType]();
       uiSchema = await siteModule.buildUiSchema(
         i18nScope,
-        entity as IHubSite,
+        options as IHubSite,
```
juliannemarik, line 77: I'm wondering if this should now be `options as IEntityEditorOptions` and all of the corresponding `buildUiSchema` functions should also update their functional interface to expect an `IEntityEditorOptions`...?

packages/common/src/core/schemas/internal/metrics/interfaces.ts
```diff
@@ -0,0 +1,45 @@
...
+  always = "always",
+  hover = "hover",
+}
+
+export enum SOURCE {
+  dynamic = "dynamic",
+  static = "static",
+}
+
+export interface IShareableCard {
```
juliannemarik, line 33: still wondering if this and `IMAGE_TYPES` should be exported from `hub.js` at all. I don't think they're related to metrics?

packages/common/src/initiative-templates/HubInitiativeTemplate.ts
```diff
@@ -156,7 +153,7 @@ export class HubInitiativeTemplate
    */
   async getEditorConfig(
     i18nScope: string,
-    type: EditorType
+    type: EntityEditorType
```
juliannemarik, line 156: I think you could make this more explicit with `InitiativeTemplateEditorType` since this is a method on the `HubInitiativeTemplate` class

packages/common/test/projects/HubProject.test.ts
```diff
@@ -118,7 +118,7 @@ describe("HubProject Class:", () => {
     // });
 
     // it("returns editorConfig integrating options", async () => {
-    //   const spy = spyOn(schemasModule, "getEntityEditorSchemas").and.callFake(
+    //   const spy = spyOn(schemasModule, "getEditorSchemas").and.callFake(
```
juliannemarik, line 121: I'd honestly just remove these commented out tests. This one at least doesn't seem relevant anymore

packages/common/src/core/getTypeFromEntity.ts
```diff
@@ -8,7 +8,7 @@ import { HubEntity, IHubItemEntity, HubEntityType } from "./types";
  * @returns
  */
 export function getTypeFromEntity(
-  entity: IHubItemEntity | HubEntity | ConfigurableEntity
+  entity: IHubItemEntity | HubEntity | IEntityEditorOptions
```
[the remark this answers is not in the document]
jordantsanz: Could we instead keep IEntityEditorOptions and add `{type: string, [key: string]: string;}`? IEntityEditorOptions doesn't necessarily have a type on it, so there's some sort of casting weirdness that has to happen here if not, I think...

juliannemarik (approved):
Really great work!

packages/common/src/content/HubContent.ts
```diff
@@ -1,10 +1,7 @@
 import { IArcGISContext } from "../ArcGISContext";
 import { HubItemEntity } from "../core/HubItemEntity";
-
-import {
-  IEditorConfig,
-  IWithEditorBehavior,
-} from "../core/behaviors/IWithEditorBehavior";
+import { IEditorConfig } from "../core";
```
dbouwman, line 3: import from the file :)

packages/common/src/core/schemas/getEditorConfig.ts
```diff
@@ -1,22 +1,23 @@
-import { IEditorConfig } from "../behaviors/IWithEditorBehavior";
+import { IEditorConfig } from ".";
```
dbouwman, line 1: import from file

packages/common/src/core/schemas/internal/getCardEditorSchemas.ts
```diff
@@ -0,0 +1,49 @@
+import { IEditorConfig } from "..";
+import { CardEditorType } from "../types";
+import { getCardType } from "./getCardType";
+import { filterSchemaToUiSchema } from "./filterSchemaToUiSchema";
+import { ICardEditorOptions } from "./EditorOptions";
+import { cloneObject } from "../../../util";
+import { IArcGISContext } from "../../..";
+
+export async function getCardEditorSchemas(
```
dbouwman, line 9: Doc comment please :)

packages/common/src/core/schemas/internal/getEditorSchemas.ts
```diff
@@ -1,6 +1,11 @@
 import { cloneObject } from "../../../util";
-import { IEditorConfig } from "../../behaviors/IWithEditorBehavior";
-import { EditorType, IConfigurationSchema, IUiSchema } from "../types";
+import { IEditorConfig } from "..";
```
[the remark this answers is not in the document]
jordantsanz: Will check them -- didn't realize that it caused issues. Will note for future 👍

packages/common/src/core/schemas/internal/EditorOptions.ts
```diff
@@ -0,0 +1,19 @@
+import { HubEntity } from "../../types";
+
+/** Intersection type of all EditorOptions */
+export type EditorOptions = IEntityEditorOptions | ICardEditorOptions;
```
dbouwman, line 4: I wonder about using a union type here. As you saw, in `getEditorConfig(...)` fn, you had to use `as ...` or typescript would complain. And really, what options structure is valid, depends on the `type` that's passed into `getEditorConfig(...)`... as in it's invalid to pass `type: "card"` and `options: IEntityEditorOptions` but this form of typing does not actually prevent that. 
One way to get around this is to use a [Discriminated Unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) which would combine the type string + the options props, and then you can be sure you're not getting mixed values that result in invalid conditions

PR #1283, opened 2023-10-18
refactor: rows with new calcite text area for details summary

abp6318:
1. Description: OD-UI PR: https://github.com/ArcGIS/opendata-ui/pull/12664
- calcite input --> calcite text area
- rows can now be used for textarea types
- vertical + none now used for textarea types

1. Instructions for testing: Confirm each workspace details pane has a summary textarea input with 4 rows

1. Closes Issues: [#8125](https://devtopia.esri.com/dc/hub/issues/8125)

### DIFF
--- a/packages/common/src/content/_internal/ContentUiSchemaEdit.ts
+++ b/packages/common/src/content/_internal/ContentUiSchemaEdit.ts
@@ -52,6 +52,7 @@ export const buildUiSchema = async (
             options: {
               control: "hub-field-input-input",
               type: "textarea",
+              rows: 4,
               helperText: {
                 labelKey: `${i18nScope}.fields.summary.hint`,
               },
--- a/packages/common/src/discussions/_internal/DiscussionUiSchemaEdit.ts
+++ b/packages/common/src/discussions/_internal/DiscussionUiSchemaEdit.ts
@@ -66,6 +66,7 @@ export const buildUiSchema = async (
             options: {
               control: "hub-field-input-input",
               type: "textarea",
+              rows: 4,
               helperText: {
                 labelKey: `${i18nScope}.fields.summary.helperText`,
               },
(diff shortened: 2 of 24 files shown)


packages/common/src/content/_internal/ContentUiSchemaEdit.ts
```diff
@@ -52,6 +52,7 @@ export const buildUiSchema = async (
             options: {
               control: "hub-field-input-input",
               type: "textarea",
+              rows: 4,
```
sonofflynn89, line 55: Should we change this to be the default in hub-components but still respect an override? Might be good so we don't have to always duplicate this option.

abp6318: Maybe... the only reason I could see that going wrong is if we're using the config editor with textareas elsewhere and it's an unwanted change, but I don't know if we are.

sonofflynn89: @juliannemarik any thoughts? If you agree with Aaron then we can totally leave it as it is

juliannemarik: Yes, what Aaron said would be my primary concern. @abp6318, I'm also wondering if you were given any instruction to make the same "row" adjustment for the "description" fields? If not, and we want the "description" field to remain as is, we'd essentially need this same PR to override the default of 4.

In any case, I feel like this might be a bit more of a product question as to what they want the default height to be set at for textarea fields.

sonofflynn89: @juliannemarik and @abp6318 good points, let's get this merged then and we can open up a ticket to get more product guidance for the future.

PR #1288, opened 2023-10-19
feat(hub-discussions): channel creator not considered an owner (v2)

brittneyjb:
1. Description: Modified `canModifyChannel` to remove the channel creator check as we are moving to rely only on ACL

1. Related Issue: [#6124] (https://devtopia.esri.com/dc/hub/issues/6124)


ethanspiro (approved):
👍

velveetachef (changes requested):
Nice work. Will also need to adjust the access check related to channelAcl (ChannelPermission class block 20-23, and the underlying method logic https://github.com/Esri/hub.js/blob/102fceeb8128fda14f48a9e801c41c6c3e9a874b/packages/discussions/src/utils/channel-permission.ts#L78). Double check, but I believe that is the only method using the `creator`, param, so that can be removed from the constructor as well.

Also this PR will be on ice until the coordinated v2 release with the front end, since this is shared logic with the front end)

PR #1296, opened 2023-10-23
feat(): add metrics schema and uiSchema to projects

jordantsanz:
1. Description:
Allows projects to use the metrics schema to soon have metrics in the projects workspace.
Adds editorContext as an optional param for fromEditor and toEditor on projects. 
Hoists the transformValuesToMetricAndCardConfig function into hub.js and renames it editorToMetric, adds metricToEditor and other metric utils. 

1. Instructions for testing:

1. Closes Issues: [#7629](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/7629)


packages/common/test/core/EntityEditor.test.ts
```diff
@@ -95,7 +95,7 @@ describe("EntityEditor:", () => {
       expect(toEditorSpy).toHaveBeenCalled();
       expect(chk.id).toBe("00c");
       await editor.save(chk);
-      expect(fromEditorSpy).toHaveBeenCalledWith(chk);
+      expect(fromEditorSpy).toHaveBeenCalledWith(chk, undefined);
     });
   });
   describe("supports pages:", () => {
```
jordantsanz, line 101: editorContext is optional -- and we don't use it anywhere else yet -- so it's undefined right now

packages/common/src/projects/HubProject.ts
```diff
@@ -288,6 +302,28 @@ export class HubProject
       }
     }
 
+    // handle metrics
```
juliannemarik, line 305: not necessarily something to consider as part of this work, but something that's been on my mind is that these `toEditor`/`fromEditor` methods are becoming a bit verbose, and technically to stand this up for initiatives, we'll need to copy this same code into the initiative class...maybe that's not a bad thing, and it allows flexibility for when logic differs between entities, but it would be nice if this entire block was abstracted into a reusable/modular util that's like `handleFromEditorMetrics` so this entire block could be replaced by a single line. Same with the `toEditor` logic above. But I think that's probably unnecessary/over-optimization for now.

packages/common/src/projects/_internal/ProjectSchema.ts
```diff
@@ -24,5 +26,9 @@ export const ProjectSchema: IConfigurationSchema = {
       default: PROJECT_STATUSES.notStarted,
       enum: Object.keys(PROJECT_STATUSES),
     },
+    _metrics: {
```
juliannemarik, line 29: could this just be `_metrics: MetricSchema`?

juliannemarik (commented):
Left some minor comments to consider, but otherwise I think this is looking great

juliannemarik (approved):
Looks good to me. Really great work!

PR #1298, opened 2023-10-24
fix(): remove scope from uiSchema for stat card

jordantsanz:
1. Description:
Removes the i18nscope from the stat card uiSchema to allow for no scope to be passed in.
1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/internal/metrics/StatCardUiSchema.ts
+++ b/packages/common/src/core/schemas/internal/metrics/StatCardUiSchema.ts
@@ -8,6 +8,7 @@ import { UiSchemaRuleEffects, IUiSchema } from "../../types";
  * @returns
  */
 export const buildUiSchema = (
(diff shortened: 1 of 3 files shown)


packages/common/src/core/schemas/internal/metrics/StatCardUiSchema.ts
```diff
@@ -8,7 +8,6 @@ import { UiSchemaRuleEffects, IUiSchema } from "../../types";
  * @returns
  */
 export const buildUiSchema = (
```
juliannemarik, line 10: I'm just wondering if we should still leave `i18nScope` as an arg but just not use it anywhere in the `uiSchema` for now 🤔 I'm fine either way though. I can't really think of a use-case where these strings would change based on the context in which this editor is used.

juliannemarik (approved):

PR #1315, opened 2023-11-03
feat(hub-common): add default predicate for portal search

sansth1010:
1. Description: This PR adds a function `getItemQueryWithDefaultPredicates` that adds default item predicate such as ignoring 'Code Attachment' item types to `IQuery` which can be used before calling portal search or hub search.

Example of `getItemQueryWithDefaultPredicates()` usage in hub-search in this [PR](https://github.com/ArcGIS/hub-search/pull/32/files#diff-5dd293b7273f0b2983d5ebde4c99856cc706e30b119f5f51c892bb87e0238874)  

1. Instructions for testing:

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/8144


packages/common/src/search/utils.ts
```diff
@@ -349,3 +349,25 @@ export function getResultSiteRelativeLink(
...
   return siteRelativeLink;
 }
+
+/**
+ * Adds default predicates for item target entity
+ *
+ * @param query IQuery to search items
+ * @returns an IQuery with default predicates
+ */
+export function getItemQueryWithDefaultPredicates(query: IQuery): IQuery {
```
sonofflynn89, line 359: Looking great! Since you are modifying an existing query (albeit via cloning) as part of this function, I would rename the function to `addDefaultItemSearchPredicates` and specify in the TS DOC comment that a clone is returned.

packages/common/src/search/utils.ts
```diff
@@ -349,3 +349,25 @@ export function getResultSiteRelativeLink(
...
+  if (query.targetEntity === "item") {
```
sonofflynn89, line 360: Normally it's good to do a check like this, but because the name of the function specifies that it works on item queries, you can remove this. Within hub.js directly, this function will never get called with a `query.targetEntity` other than item.

packages/common/test/search/_internal/portalSearchItems.test.ts
```diff
@@ -73,6 +73,44 @@ describe("portalSearchItems Module:", () => {
...
+      expect(expectedParams.q).toEqual('(water) AND (-type:"Code Attachment")');
+      expect(expectedParams.countFields).not.toBeDefined();
+    });
+    it("search query does not contain default item predicate for non-item entities", async () => {
+      const searchItemsSpy = spyOn(Portal, "searchItems").and.callFake(() => {
+        return Promise.resolve(cloneObject(AllTypesResponse));
+      });
+
+      const qry: IQuery = {
+        targetEntity: "group",
```
sonofflynn89, line 89: See my comment above. I know you have to do this to get 100% code coverage, but by the time we are within `portalSearchItems`, we are guaranteed to have `targetEntity: "item"`. Go ahead and remove this test.

sonofflynn89 (commented):
Looking good San! Just a few minor tweaks and then you should be good to go!

packages/common/src/search/utils.ts
sansth1010: I agree! changed the name to `addDefaultItemSearchPredicates` and added the comment.

packages/common/src/search/utils.ts
sansth1010: I agree and that is a great point! I have removed the check.

sonofflynn89 (approved):
Great work San! Thank you so much for doing this 🙏

PR #1316, opened 2023-11-03
feat(hub-common): persist location onto discussion allowedLocations settings

tannerjt:
1. Description:

Persists `IHubItemEntity` `location` geometries onto `allowedLocations` discussion settings, converting to GeoJSON in Discussions DB.  For this PR, I've opted to port over only what we need from `terraformer` instead of adding an additional dependency.  This saves us ~30kb and avoids an additional external dependency to fetch.

1. Instructions for testing:


packages/common/src/discussions/edit.ts
```diff
@@ -126,13 +127,34 @@ export async function updateDiscussion(
...
+  // persist location geometries to discussion settings as Geometry[]
+  let allowedLocations;
+  try {
+    allowedLocations =
+      updatedDiscussion.location?.geometries?.map((geometry) =>
+        arcgisToGeoJSON(geometry)
+      ) || null;
+  } catch (e) {
+    /* istanbul ignore else */
+    if (e instanceof Error) {
```
rweber-esri, line 139: I'm curious why this is needed? Can `arcgisToGeoJson` throw a non-error value (e.g. a `string`, suspect related to [this line](https://github.com/Esri/hub.js/pull/1316/files#diff-e52ff993e5b69ce6496000695934c6bbafe6dd867d76e42052486a335836188fR176))? If so, is the full array of geometries successfully converted to geojson when that occurs? `allowedLocations` will be `undefined`, curious if it should be `null`?

Additionally, the way this logic is written, `allowedLocations` could be one of `undefined`, `null` or `Geojson[]`, is that intentional? I would think it would be either `null` or `Geojson[]`?

packages/common/src/discussions/edit.ts
```diff
@@ -126,13 +127,34 @@ export async function updateDiscussion(
...
+  }
+
   // create or update entity settings
   const defaultSettings = getDefaultEntitySettings("discussion");
   const settings = {
     ...defaultSettings.settings,
     discussions: {
       ...defaultSettings.settings.discussions,
-      ...updatedDiscussion.discussionSettings,
+      ...{
```
rweber-esri, line 154: Do we need to wrap this in an object? Curious if this could be simplified to just:

```tsx
  const settings = {
    ...defaultSettings.settings,
    discussions: {
      ...defaultSettings.settings.discussions,
      ...updatedDiscussion.discussionSettings,
      allowedLocations,
    },
  };
```

packages/common/src/utils/internal/arcgisToGeoJson.ts
```diff
@@ -0,0 +1,249 @@
+// @terraformer/arcgis
+// https://github.com/terraformer-js/terraformer/blob/main/packages/arcgis/README.md
```
rweber-esri, line 2: I'm curious why we need to copy this into the project? I suspect either we can't import it for some reason, or it imports way more than we need?

tomwayson: agree, according to bundlephobia `@terraformer/arcgis` is tree-shakable, so we _should_ only end up w/ what we need: https://bundlephobia.com/package/@terraformer/arcgis@2.1.2

rweber-esri (approved):
LGTM, thanks for making those updates!

PR #1319, opened 2023-11-06
feat: add status to hub initiative

vivzhang:
affects: @esri/hub-common

1. Description:

- Add support for hub initiative status
- Replace certain supporting code with generic ones to support both projects and initiatives

1. Instructions for testing:

1. Closes Issues: [8322](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/8322)


packages/common/src/core/types/IHubInitiative.ts
```diff
@@ -11,6 +12,8 @@ export interface IHubInitiative
     IWithSlug,
     IWithCatalog,
     IWithMetrics,
-    IWithPermissions {}
+    IWithPermissions {
+  status: HubEntityStatus;
+}
```
benstoltz, line 17: I think we might want to have an `IWithStatus` which exposes `{ status: HubEntityStatus; }`. Thoughts @dbouwman ?

packages/common/src/core/types/IHubProject.ts
```diff
@@ -17,14 +18,7 @@ export interface IHubProject
     IWithLayout,
     IWithMetrics,
     IWithPermissions {
-  status: PROJECT_STATUSES;
-}
-
-export enum PROJECT_STATUSES {
```
juliannemarik, line 23: Technically removing this is a breaking change. I'm not sure how strict we are about breaking changes, since it's likely this is only being used internally and we can replace our own instances in `opendata-ui`; however, an alternative to releasing a breaking change is to mark this as deprecated and direct people to start using the new `HubEntityStatus`, and then we can remove this during our next planned breaking change

vivzhang: That's a good point. It's likely safe but keeping it and marking it deprecated is a good idea.

packages/common/src/core/types/IHubInitiative.ts
dbouwman: Agreed - good catch @benstoltz :)

packages/common/src/core/types/IHubProject.ts
dbouwman: Sorry if I was not clear in our chat about this @vivzhang - we should keep the PROJECT_STATUSES enum but mark as deprecated.

package.json
```diff
@@ -58,7 +58,7 @@
 		"jasmine": "^3.6.4",
 		"jasmine-core": "3.4.0",
 		"jasmine-ts": "^0.4.0",
-		"karma": "^6.1.0",
+		"karma": "^6.4.2",
```
dbouwman, line 61: Why is this changing? I'm always leery when we bump tooling versions in a PR that's not about upgrading tooling :)

packages/common/src/core/types/IHubProject.ts
```diff
@@ -35,3 +29,13 @@ export type IHubProjectEditor = IHubItemEntityEditor<IHubProject> & {
   // Groups is an ephemeral property, so we prefix with _
   _groups?: string[];
 };
+
+// *** THIS IS DEPRECATED ***
+// This is deprecated and will be removed during our next planned
+// breaking change. Please use HubEntityStatus instead
+export enum PROJECT_STATUSES {
```
dbouwman, line 36: Change this of a jsdoc style comment and add `@internal` to the start so it's kept out of the docs

package.json
vivzhang: Good catch! I upgraded it locally when I was debugging, accidentally got it in here.

dbouwman: If it works, maybe open a PR that just bumps that... we should try to keep our deps updated, but we should do it intentionally so it's obvious if it ends up causing a failure somewhere down the line :)

vivzhang: Pretty sure it's unrelated, I think we are good.

juliannemarik (approved):

benstoltz (approved):

PR #1321, opened 2023-11-07
fix: switch null to empty string for membershipAccess

benstoltz:
1. Description: request, from arcgis-rest-request, does not allow one to use null/undefined's as empty vals. An empty string needs to be sent. ( https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-request/src/utils/process-params.ts#L72-L75 ).

In this particular case membershipAccess was being stripped from the update call altogether when set to 'anyone'

1. Instructions for testing:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64981,7 +64981,7 @@
 		},
 		"packages/common": {
 			"name": "@esri/hub-common",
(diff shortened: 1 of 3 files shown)


juliannemarik (approved):
👍

PR #1322, opened 2023-11-07
feat(): allow metrics visibility

jordantsanz:
1. Description:
Adds the optional `visibility` prop to IMetricDisplayConfig.
1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/types/Metrics.ts
+++ b/packages/common/src/core/types/Metrics.ts
@@ -216,13 +216,13 @@ export enum ExpressionRelationships {
 }
 
 /**
- * Configuration for how to show a metric in the ui
+ * Configuration for how to show a metric in the ui. These are just values that are important to how to
+ * display the metric; values that are configured in the editor should be added to IMetricEditorValues instead.
  */
 export interface IMetricDisplayConfig extends IMetricEditorValues {
   metricId: string;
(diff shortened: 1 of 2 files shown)


packages/common/src/core/types/Metrics.ts
```diff
@@ -221,9 +221,20 @@ export enum ExpressionRelationships {
 export interface IMetricDisplayConfig {
   metricId: string;
   displayType: string;
+  visibility?: MetricVisibility;
```
juliannemarik, line 224: Maybe you already have plans to do this, but I'm wondering if we want to expose other explicit (optional) display config properties on this interface.
Actually, I'm taking a closer look at this file and wondering if `IMetricDisplayConfig` should `extend` `IMetricEditorValues` rather than the other way around (which is what is currently happening)? `IMetricEditorValues` should be properties that come directly out of the editor, whereas `IMetricDisplayConfig` could include other additional properties. Am I confused?

jordantsanz: Yeah, I think you've got it right -- I'll refactor it to the other way around.

I think we could refactor it to have the optional display config properties -- I'll start looking into that in the create/edit PRs

juliannemarik (approved):

PR #1340, opened 2023-11-21
fix(hub-common): for Channel group enrichment, account for null groups

tannerjt:
affects: @esri/hub-common

1. Description:

Handles Error encountered when searching for group enriched channels and `groups: null`.

Screenshot 2023-11-21 at 9 13 02 AM

1. Instructions for testing:

1. Relating to issue 8595

### DIFF
--- a/packages/common/src/search/_internal/hubSearchChannels.ts
+++ b/packages/common/src/search/_internal/hubSearchChannels.ts
@@ -142,24 +142,25 @@ export const toHubSearchResult = async (
       },
       includes: {
         // when 'include' requests 'groups' enrichment, get group details
-        groups: options.include?.includes("groups")
-          ? await Promise.all(
-              channel.groups.map(async (groupId) => {
-                let group;
-                try {
-                  group = await getGroup(groupId, options.requestOptions);
-                } catch (e) {
-                  group = null;
-                  /* tslint:disable-next-line: no-console */
-                  console.warn(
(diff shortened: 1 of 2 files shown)


packages/common/src/search/_internal/hubSearchChannels.ts
```diff
@@ -142,24 +142,25 @@ export const toHubSearchResult = async (
...
-                    `Cannot fetch group enhancement for id = ${groupId}`,
-                    e
-                  );
-                }
-                return group;
-              })
-            )
-          : [],
+        groups:
+          options.include?.includes("groups") && channel.groups
```
rweber-esri, line 146: Is this a valid use case? I see some channel records on QA that have `groups: null` that appear to have been created by an `E2E` user. I'm curious if those records were created intentionally or by mistake or if they somehow got into a weird state.. I've not run into this before and suspect this would cause a lot of other issues elsewhere. Those same records also have `access: null`, which I find equally alarming and could be problematic too.

tannerjt: I noticed these were created via e2e's.  We should check with backend.  Here's some examples:

Screenshot 2023-11-20 at 3 15 36 PM

tannerjt:
Closing this story since these were e2e anomalies reflecting v2 ACL.

PR #1345, opened 2023-12-01
feat: add enrichement to refetch item with additional props

dbouwman:
1. Description:

Add support of `include: ["item.itemControl"]` which requires re-fetching of the item to get this additional property that's not present on an item search result.

Also fixes a typo which was preventing includes from working.

1. Instructions for testing:

1. Closes Issues: n/a - side effect of 8462 (gallery callback)

### DIFF
--- a/packages/common/src/search/_internal/portalSearchItems.ts
+++ b/packages/common/src/search/_internal/portalSearchItems.ts
@@ -173,7 +173,7 @@ async function searchPortalAsHubSearchResult(
   const fn = (item: IItem) => {
     return itemToSearchResult(
       item,
-      searchOptions.includes,
+      searchOptions.include,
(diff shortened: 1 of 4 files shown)


packages/common/src/search/_internal/portalSearchItems.ts
```diff
@@ -173,7 +173,7 @@ async function searchPortalAsHubSearchResult(
   const fn = (item: IItem) => {
     return itemToSearchResult(
       item,
-      searchOptions.includes,
```
dbouwman, line 176: This typo was preventing the `include` list from working... downside of allowing `searchOptions` to have `[key: string]: any` as an outlet is that we also lost checks

PR #1347, opened 2023-12-04
fix(hub-common): use only capabilities key to update extract capability service definition

sansth1010:
…ty service definition

1. Description: This PR fixes `updateServiceDefinition` payload when saving extract capability option to include only `capabilities` key as required instead of passing the whole service definition. This will prevent errors like [#7010](https://devtopia.esri.com/dc/hub/issues/7010)

1. Instructions for testing:

1. Closes Issues: [#7010](https://devtopia.esri.com/dc/hub/issues/7010)

### DIFF
--- a/packages/common/test/content/edit.test.ts
+++ b/packages/common/test/content/edit.test.ts
@@ -189,7 +189,7 @@ describe("content editing:", () => {
     });
     it("updates the hosted service if configurations have changed", async () => {
       const currentDefinition: Partial<featureLayerModule.IFeatureServiceDefinition> =
-        { capabilities: "Query" };
+        { currentVersion: 11.2, capabilities: "Query" };
       getServiceSpy.and.returnValue(Promise.resolve(currentDefinition));
 
       const content: IHubEditableContent = {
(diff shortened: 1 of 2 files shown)


packages/common/test/content/edit.test.ts
```diff
@@ -227,6 +227,13 @@ describe("content editing:", () => {
...
       expect(getServiceSpy).toHaveBeenCalledTimes(1);
       expect(updateServiceSpy).toHaveBeenCalledTimes(1);
+      expect(updateServiceSpy).toHaveBeenCalledWith(
+        "https://services.arcgis.com/:orgId/arcgis/rest/services/:serviceName/FeatureServer",
+        {
+          authentication: MOCK_AUTH,
+          updateDefinition: { capabilities: "Query,Extract" },
+        }
+      );
       const [url, { updateDefinition }] = updateServiceSpy.calls.argsFor(0);
```
sonofflynn89, line 237: Heads up that this check is now redundant with the one you just introduced. You may also want to change this test case so the `getService` spy returns an additional field that `updateService` won't see.

sonofflynn89 (commented):
Looking great! Just please update the tests accordingly

packages/common/test/content/edit.test.ts
sansth1010: Good call and suggestion! I have made the change you suggested.

PR #1350, opened 2023-12-04
fix: enable featured image alt text on hub initiative

vivzhang:
affects: @esri/hub-common

1. Description:
Enable featured image alt text on hub initiative

1. Instructions for testing:

[8182](https://zentopia.esri.com/workspaces/collaboration-sprint-board-614f9bfc0946c40011ef574e/issues/dc/hub/8182)

### DIFF
--- a/packages/common/src/initiatives/_internal/computeProps.ts
+++ b/packages/common/src/initiatives/_internal/computeProps.ts
@@ -43,6 +43,7 @@ export function computeProps(
       model.data.view.featuredImageUrl,
       requestOptions
     ),
+    featuredImageAltText: model.data.view.featuredImageAltText,
     hero: model.data.view.hero,
   };
 


packages/common/src/initiatives/_internal/computeProps.ts
```diff
@@ -43,6 +43,7 @@ export function computeProps(
       model.data.view.featuredImageUrl,
       requestOptions
     ),
+    featuredImageAltText: model.data.view.featuredImageAltText,
```
juliannemarik, line 46: hmm, it seems to me that this should already be mapped over by the property mapper...and same with the hero property actually. I think the real issue here might be that we're overwriting `initiative.view` rather than updating the properties that actually need "computing". Does it work if you just have the following instead: 

`initiative.view = {
  ...model.data.view,
  featuredImageUrl: getAuthedImageUrl(
    model.data.view.featuredImageUrl,
    requestOptions
  )
}

vivzhang: You're right, we can get it from the model data

juliannemarik (approved):

PR #1352, opened 2023-12-06
[WIP] fix: context.refreshUser updated cached object

dbouwman:
1. Description:

`context.refreshUser()` will also update the cached context manage in localStorage

Also adds `STORAGE_KEYS` hash of localStorage keys.

There will be a follow-up PR to -ui to start using the `STORAGE_KEYS` in components and utils

1. Instructions for testing:

1. Closes Issues: [8694](https://devtopia.esri.com/dc/hub/issues/8694)

1. [ x PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

### DIFF
--- a/packages/common/src/ArcGISContext.ts
+++ b/packages/common/src/ArcGISContext.ts
@@ -226,7 +226,7 @@ export interface IArcGISContext {
   /**
    * Refresh the current user, including their groups
    */
-  refreshUser(win?: any): Promise<void>;
+  refreshUser(): Promise<void>;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


tomwayson (commented):
I guess this is fine, but I had been thinking that all the local storage stuff belongs at the hub-components level. We know that we're always working in a browser there, and that is where we have `storeContextManager()` and `retrieveContextManager()` 

Whatever is calling this could call:

```ts
await contextManager.refreshUser()
storeContextManager(contextManager)
```

and that could be wrapped into a util at that level like `refreshUser(contextManager)` that does both of the above.

dbouwman:
@tomwayson generally I agree with the sentiment of keeping localStorage stuff in the "ui layer", however, in this case, the value of encapsulating the fact that we have a serialized context in localStorage is more valuable. By checking `typeof(window)` we can have the function branch depending if it's in node vs browser, and be functionally the same for the consumer.

dbouwman:
Putting this on hold - it's not high priority at this point, but due to issues spying on localStorage in FireFox, along with refactors suggested by @tomwayson, this needs more thought before moving ahead.

PR #1349, opened 2023-12-04
chore: update github token for release

haoliangyu:
1. Description:

Update the GitHub token used in release to be a personal access token from Hub's GitHub service account. This will enable the implementation of the branch protection but allow the service account to bypass protection rules during the release (direct push to the master branch).

The secret has been created.

1. Instructions for testing:

1. Closes Issues: #<number> (if appropriate)

1. [ ] Updated meaningful TSDoc to methods including Parameters and Returns, see [Documentation Guide](https://esri.github.io/hub-components/storybook/?path=/story/guides-documentation--page)

1. [x] used semantic commit messages
  
1. [x] PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

1. [ ] updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.

PR #1365, opened 2023-12-16
chore: re-enable branch protection

haoliangyu:
Update the GitHub token used in release to be a personal access token from Hub's GitHub service account. This will enable the implementation of the branch protection but allow the service account to bypass protection rules during the release (direct push to the master branch).

Compared to the original PR https://github.com/Esri/hub.js/pull/1349, this new PR adds a necessary parameter for the `checkout` action. It makes sure the `somatic-release` would use the provided token (not the one in the environment) for release. See discussion at https://github.com/semantic-release/semantic-release/issues/2636.

The secret has been created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Release
         env:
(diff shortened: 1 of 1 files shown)


tomwayson (approved):
thanks @haoliangyu

PR #1371, opened 2023-12-27
feat(): add type for static metric value

jordantsanz:
1. Description:

1. Instructions for testing:


packages/common/src/core/schemas/internal/metrics/MetricSchema.ts
```diff
@@ -20,7 +20,12 @@ export const MetricSchema: IConfigurationSchema = {
       type: "string",
     },
     value: {
+      type: ["number", "string"],
```
juliannemarik, line 23: I think we could consider using conditional schemas (e.g. if/else/then) based on the value of `valueType` to make this `type` more explicit: 

```ts
{
  type: "object",
  properties: {
    ...,
    value: {},
    valueType: {
      type: "string",
      default: "string",
      enum: ["string", "number", "date"],
    },
    ...
  },
  if: {
    properties: { valueType: { enum: ["string", "date"] } },
  },
  then: {
    properties: { value: { type: "string" } },
  },
  else: {
    properties: { value: { type: "number" } },
  }
}
```

packages/common/src/core/schemas/internal/metrics/ProjectUiSchemaMetrics.ts
```diff
@@ -60,12 +63,65 @@ export const buildUiSchema = (
...
+                i18nScope: `${i18nScope}.fields.metrics.valueType.enum`,
+              },
+            },
+          },
           {
             labelKey: `${i18nScope}.fields.metrics.value.label`,
             scope: "/properties/_metric/properties/value",
             type: "Control",
-            rule: SHOW_FOR_STATIC_RULE_ENTITY,
+            rule: SHOW_FOR_STATIC_AND_STRING_RULE_ENTITY,
```
juliannemarik, line 87: yeah, I don't have any problem/see any other way around overloading this field with 3 different `uiSchema` definitions based on the value of `valueType`. I think this is fine.

packages/common/src/core/schemas/internal/metrics/ProjectUiSchemaMetrics.ts
```diff
@@ -60,12 +63,65 @@ export const buildUiSchema = (
...
+              ],
+            },
+          },
+          {
+            labelKey: `${i18nScope}.fields.metrics.value.label`,
+            scope: "/properties/_metric/properties/value",
+            type: "Control",
+            rule: SHOW_FOR_STATIC_AND_NUMBER_RULE_ENTITY,
+            options: {
+              type: "number",
```
juliannemarik, line 106: I know this is already how it's implemented in the `input` field, but I don't love that we need to specify the `type` in the `uiSchema.options`. It feels like this should come directly from the `schema.type` which already specifies that this is a `number`. Originally I was thinking we could change the [`type` getter](https://github.com/ArcGIS/opendata-ui/blob/master/packages/hub-components/src/components/arcgis-configuration-editor/components/arcgis-configuration-editor-field/fields/input/input.tsx/#L35-L45) in the input field to something like: 

```tsx
get type(): any {
  const shemaTypeToType = {
    integer: 'number',
    number: 'number',
    string: 'text'
  };
  const schemaFormatToType = {
    ['date-time']: 'datetime-local',
    ['idn-email']: 'email'
  }

  return Array.isArray(this.params.schema.type)
    ? "text"
    : schemaFormatToType[this.params.schema.format]
      || this.params.schema.format
      || this.params.uiSchema?.options?.type
      || shemaTypeToType[this.params.schema.type]
}
```

But I think we'd actually run into issues when using conditional subschemas like what I've proposed above because the `type` isn't actually stored in the property's schema 🤔

So anyways, this is just a long way of saying stick with this for now 😅

juliannemarik (commented):
left one comment for consideration, but otherwise, I think this is a solid proposal! I'm good with proceeding in this direction

juliannemarik (approved):
Left two minor comments, otherwise this looks good to me

PR #1376, opened 2024-01-04
feat(hub-common): adds IHubLocation to IHubSearchResult

tannerjt:
affects: @esri/hub-common

1. Description: Adds new `location: IHubLocation` property to `IHubSearchResult` interface, reflecting `item.properties.location`.  These locations are typically defined in the details pane of an entity workspace, and allow for multiple locations to be associated with a single item (example: multiple project locations).  This new `location` property is added in addition to the existing `geometry` property, the reason being that both are necessary for preserving spatial information about the item.

1. Instructions for testing:

1. Closes Issues: [#8910](https://devtopia.esri.com/dc/hub/issues/8910)


abp6318 (approved):

packages/common/src/search/_internal/hubSearchItemsHelpers/ogcItemToSearchResult.ts
```diff
@@ -19,13 +19,13 @@ export async function ogcItemToSearchResult(
   // as `license` and `source` if the OgcItem came from the index.
   const pseudoItem = ogcItem.properties as IItem;
   const result = await itemToSearchResult(pseudoItem, includes, requestOptions);
-  // Expose extraneous members like `license`, `source`, and `geometry`
+  // Expose extraneous members like `license`, `source`, `properties.location` and `geometry`
   result.source = ogcItem.properties.source;
   result.license = ogcItem.properties.license;
+  result.location = ogcItem.properties.properties?.location;
```
ajturner, line 25: do we really store location in `item.properties.properties.location`? Is this expected or was it an implementation bug?

ajturner: Additionally - @dbouwman and I had discussed that if a geometry was too large, it could be stored in an Item resource and the `item.properties.location` would have a JSON-ref to the resource file. Was that implemented? If so, this line would need to optionally fetch that resource into the result attribute.

tannerjt: This is a result of the [IOgcItem interface](https://esri.github.io/hub.js/api/common/IOgcItem/), which puts the entire item on the `properties` property.  Location is stored on `item.properties.location`, but on an `IOgcItem`, it's stored on `item.properties.properties.location`.

dbouwman: @ajturner @tannerjt at this point we have not implemented a consistent "json-ref" approach, but it's something that would be great to add to the hub.js core. What's interesting is that while `$ref` is part of the OpenAPI spec, it's not a spec itself - I'll follow up in a Teams chat b/c I think we should move forward on this for a number of things

sonofflynn89: > This is a result of the [IOgcItem interface](https://esri.github.io/hub.js/api/common/IOgcItem/), which puts the entire item on the `properties` property. Location is stored on `item.properties.location`, but on an `IOgcItem`, it's stored on `item.properties.properties.location`.

Yep, I can corroborate that this is correct. `IOgcItem` !== `IItem`. You can think of `IOgcItem` like this:
```
IOgcItem {
  ...
  properties: IItem & {
    source: string,
    license: string,
    ...
  }
  ...
}
```

sonofflynn89 (approved):
Looking good to me!

packages/common/src/search/_internal/hubSearchItemsHelpers/ogcItemToSearchResult.ts
tannerjt: I took a look at the `$ref` docs and agree this could add value, however I assume we would need to tie this to a broader implementation.  Moving forward with this current approach while keeping in mind this is a good candidate for future `$ref`.

PR #1382, opened 2024-01-11
feat(hub-common): consolidates entity location construction and depre…

tannerjt:
…cates IHubSearchResult geometry

affects: @esri/hub-common

1. Description: Deprecates `IHubSearchResult` `geometry` property, now preferring all location be consolidated to the `location: IHubLocation` property.  To consolidate logic and provide consistent `location` property across workflows, the creation of location from item extent has been moved to `internalContentUtils` and used in both `computeProps` and `search`.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/content/_internal/internalContentUtils.ts
+++ b/packages/common/src/content/_internal/internalContentUtils.ts
@@ -101,7 +101,7 @@ export const getContentBoundary = (item: IItem): IHubGeography => {
  * @param extent Raw item extent array
  * @returns IExtent
  */
-const getExtentObject = (extent: number[][]): IExtent => {
+export const getExtentObject = (extent: number[][]): IExtent => {
   return isBBox(extent)
     ? ({ ...bBoxToExtent(extent), type: "extent" } as unknown as IExtent)
     : undefined;
@@ -130,17 +130,23 @@ export const deriveLocationFromItem = (item: IItem): IHubLocation => {
     return { type: "none" };
   }
 
+  /* istanbul ignore else */
   if (!location) {
     // IHubLocation does not exist on item properties, so construct it
     // from item extent
     const geometry: any = getExtentObject(extent);
-    return {
-      type: "custom",
-      extent,
-      geometries: [geometry],
-      spatialReference: geometry.spatialReference,
-    };
+    if (geometry) {
+      return {
+        type: "custom",
+        extent,
+        geometries: [geometry],
+        spatialReference: geometry.spatialReference,
+      };
+    }
   }
+
+  // If we get here, we have no location and no extent, so return none
+  return { type: "none" };
 };
 
 /**
--- a/packages/common/src/content/_internal/internalContentUtils.ts
+++ b/packages/common/src/content/_internal/internalContentUtils.ts
@@ -111,30 +111,36 @@ const getExtentObject = (extent: number[][]): IExtent => {
  * Derives proper IHubLocation given an ArcGIS Item.  If no
  * location (item.properties.location) is present, one will be
  * constructed from the item's extent.
- * If item.properties.boundary === 'none', this function will strip
- * location and return { type: 'none' } per this document:
- * https://confluencewikidev.esri.com/display/Hub/Hub+Location+Management
  * @param item ArcGIS Item
  * @returns IHubLocation
  */
 export const deriveLocationFromItem = (item: IItem): IHubLocation => {
   const { properties, extent } = item;
+  const location: IHubLocation = properties?.location;
+
+  if (location) {
+    // IHubLocation already exists, so return it
+    return location;
+  }
+
   if (properties?.boundary === "none") {
+    // Per https://confluencewikidev.esri.com/display/Hub/Hub+Location+Management
+    // bounds = 'none' -> specifies not to show on map.  If this is true and
+    // no location is already present, opt to not generate location from item extent
     return { type: "none" };
   }
-  let location: IHubLocation = properties?.location;
+
   if (!location) {
     // IHubLocation does not exist on item properties, so construct it
     // from item extent
     const geometry: any = getExtentObject(item.extent);
-    location = {
+    return {
       type: "custom",
       extent,
       geometries: [geometry],
       spatialReference: geometry.spatialReference,
     };
   }
-  return location;
 };
 
 /**
--- a/packages/common/src/content/_internal/internalContentUtils.ts
+++ b/packages/common/src/content/_internal/internalContentUtils.ts
@@ -11,11 +11,12 @@
 import { parseServiceUrl } from "@esri/arcgis-rest-feature-layer";
 import { IItem, IPortal } from "@esri/arcgis-rest-portal";
 import {
+  IExtent,
   ILayerDefinition,
   ISpatialReference,
   IUser,
 } from "@esri/arcgis-rest-types";
-import { IHubContent, PublisherSource } from "../../core";
+import { IHubContent, IHubLocation, PublisherSource } from "../../core";
 import {
   IHubGeography,
   GeographyProvenance,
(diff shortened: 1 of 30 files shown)


packages/common/src/content/search.ts
```diff
@@ -45,22 +47,10 @@ export async function enrichContentSearchResult(
       siteRelative: "not-implemented",
       thumbnail: "not-implemented",
     },
-    location: item.properties?.location,
+    location: deriveLocationFromItem(item),
```
sonofflynn89, line 50: The only problem with this is that Projects and Sites do not go through this function... We may want to consider abstracting this higher up or duplicating this line across `enrichSiteSearchResult` and `enrichProjectSearchResult`.

packages/common/src/search/_internal/hubSearchItemsHelpers/ogcItemToSearchResult.ts
```diff
@@ -22,18 +22,5 @@ export async function ogcItemToSearchResult(
   // Expose extraneous members like `license`, `source`, `properties.location` and `geometry`
```
sonofflynn89, line 22: Just a reminder to update this comment before you merge

packages/common/src/content/_internal/internalContentUtils.ts
```diff
@@ -95,6 +96,47 @@ export const getContentBoundary = (item: IItem): IHubGeography => {
...
+ * location and return { type: 'none' } per this document:
+ * https://confluencewikidev.esri.com/display/Hub/Hub+Location+Management
+ * @param item ArcGIS Item
+ * @returns IHubLocation
+ */
+export const deriveLocationFromItem = (item: IItem): IHubLocation => {
+  const { properties, extent } = item;
+  if (properties?.boundary === "none") {
+    return { type: "none" };
+  }
```
tomwayson, line 124: I know you just copied @abp6318's code here, but it seems odd that we would favor the deprecated  `boundary` property. I assume that the code that sets `location` deletes `boundary`, which is fine and good if that's the case, but I can't think of a reason that we'd open ourselves up to the possibility of having `boundary` override `location` and I dont' see anything in https://confluencewikidev.esri.com/display/Hub/Hub+Location+Management that explains why we would do this.

tomwayson (commented):
This LGTM. Thanks for doing this @tannerjt!

I think we'll ultimately want to revise the logic of `deriveLocationFromItem()` to favor `item.properties.location` over `item.properties.broundary`, unless @abp6318 can remember why we did it that way.

packages/common/src/content/_internal/internalContentUtils.ts
tannerjt: Would you suggest removing this logic all-together?  I'm in favor of dropping it.

tomwayson: I see what happened, this fn wasn't just copied over verbatim. We need to update the logic here to early return location if it's present _first_. then early return if boundary is none, then finally derive from extent.

tomwayson (changes requested):

tomwayson (approved):
LGTM, thanks @tannerjt!

packages/common/src/core/_internal/computeBaseProps.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Base property mapping for item backed entity types
+ * @param item IItem
+ * @param entity IHubItemEntity
+ * @returns
+ */
+export function computeBaseProps<EntityType extends Partial<IHubItemEntity>>(
+  item: IItem,
+  entity: EntityType
```
sonofflynn89, line 13: This should probably be `IHubItemEntity` instead

tannerjt: If we use `IHubItemEntity`, we will need to cast as specific types `as IHubProject`, etc.  The above approach uses generics, so the return type will always match up.

sonofflynn89: Ahhhhhh ok, I was confused because we have an `EntityType` type already in the project (`IHubCatalog.ts`)

tannerjt:
@sonofflynn89 - Moved out of draft, flagged for re-review.

packages/common/src/discussions/_internal/computeProps.ts
```diff
@@ -25,6 +26,8 @@ export function computeProps(
     const session: UserSession = requestOptions.authentication as UserSession;
     token = session.token;
   }
+  // compute base properties on discussion
+  discussion = computeBaseProps(model.item, discussion);
```
sonofflynn89, line 30: This is great! The one thing I've noticed is that discussions don't have an `enrichDiscussionSearchResult` function, meaning they are relying on the base `enrichContentSearchResult` logic. We should probably let engagement know and verify that this was intentional and not accidental.

packages/common/src/initiative-templates/fetch.ts
```diff
@@ -82,6 +85,7 @@ export async function enrichInitiativeTemplateSearchResult(
       thumbnail: "not-implemented",
       workspaceRelative: "not-implemented",
     },
+    location: deriveLocationFromItem(item),
```
sonofflynn89, line 88: Thank you for adding this, heads up that `enrichInitiativeTemplateSearchResult` is _never called_. We should let collaboration know so they can add a switch statement to `itemToSearchResult()`

packages/common/src/initiatives/HubInitiatives.ts
```diff
@@ -269,6 +272,7 @@ export async function enrichInitiativeSearchResult(
       thumbnail: "not-implemented",
       workspaceRelative: "not-implemented",
     },
+    location: deriveLocationFromItem(item),
```
sonofflynn89, line 275: Ditto here, collaboration needs to add this function in the `itemToSearchResult` switch statement.

packages/common/test/content/_internal/internalContentUtils.test.ts
```diff
@@ -289,3 +295,95 @@ describe("getContentEditUrl", () => {
...
+          },
+          type: "extent",
+        } as any,
+      ],
+      spatialReference: {
+        wkid: 4326,
+      },
+    });
+  });
+  it("should finally resolve to none", () => {
```
sonofflynn89, line 384: Looks like this test is redundant with Line 345

sonofflynn89 (commented):
Looking great Josh! Thanks for doing all this. Just a few comments.

packages/common/src/discussions/edit.ts
```diff
@@ -131,6 +131,7 @@ export async function updateDiscussion(
   // persist location geometries to discussion settings as Polygon[]
   let allowedLocations: Polygon[];
   try {
+    /* istanbul ignore next: optional chaining appears not picked up in coverage */
     allowedLocations =
       updatedDiscussion.location?.geometries?.map(
```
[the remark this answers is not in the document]
tannerjt: It guarantees that `discussion.location` is present, but `geometries` may not be present if location = `{ type: 'none' }`

packages/common/src/content/_internal/internalContentUtils.ts
```diff
@@ -95,6 +96,59 @@ export const getContentBoundary = (item: IItem): IHubGeography => {
...
+
+  if (properties?.boundary === "none") {
+    // Per https://confluencewikidev.esri.com/display/Hub/Hub+Location+Management
+    // bounds = 'none' -> specifies not to show on map.  If this is true and
+    // no location is already present, opt to not generate location from item extent
+    return { type: "none" };
+  }
+
+  /* istanbul ignore else */
+  if (!location) {
```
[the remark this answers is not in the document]
tannerjt: Yes, you're right.  Removed from 'if' and was able to remove istanbul exception comment.  Nice catch!

packages/common/src/core/_internal/computeBaseProps.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @returns
+ */
+export function computeBaseProps<T extends Partial<IHubItemEntity>>(
+  item: IItem,
+  entity: T
+): T {
+  // TODO: Currently only location is determined for base
+  // properties, but all properties that are commonly shared
+  // across items should be moved here.
+  entity.location = entity.location || deriveLocationFromItem(item);
```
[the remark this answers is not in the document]
sonofflynn89: My bad, I once again mixed up `entity` and `item` 🤦‍♂️

packages/common/src/discussions/edit.ts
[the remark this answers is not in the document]
tannerjt: Yes, you are correct.  Removed optional chaining since `location` always present.

sonofflynn89 (approved):
Great work man! Thank you for this 🙏

PR #1386, opened 2024-01-16
fix(): update subsetSchema

jordantsanz:
1. Description:

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/internal/subsetSchema.ts
+++ b/packages/common/src/core/schemas/internal/subsetSchema.ts
@@ -7,10 +7,6 @@ import { IConfigurationSchema } from "../types";
  * @param props
  * @returns
  *
- * TODO: this function needs to be updated to include the
- * anyOf/allOf/oneOf/then/else properties. Right now, there is a temporary fix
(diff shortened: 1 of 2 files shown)


packages/common/src/core/schemas/internal/subsetSchema.ts
```diff
@@ -6,6 +6,11 @@ import { IConfigurationSchema } from "../types";
  * @param schema
  * @param props
  * @returns
+ *
+ * TODO: this function needs to be updated to include the
```
juliannemarik, line 10: minor, but I'd suggest moving this TODO down to the 3rd step specifically and making it a bit more generic - we don't need to call out that this is a temporary fix...we are still adding a valid enhancement to this function. I'd write _something_ like for the 3rd step:

```ts
// 3. filter schema compositions (anyOf, allOf, oneOf) and
// conditionals (if/then/else) down to the specified properties
// 
// TODO: enhance subset functionality to filter more complex
// compositional schemas (anyOf, oneOf) and non-if/then/else conditional structures
```

juliannemarik (approved):

PR #1394, opened 2024-01-24
feat: follow card permission and editor support

vivzhang:
1. Description:

- create a new `hub:card:follow` permission that's gated to qaext, alpha and hub premium
- create basic follow card schema, uiSchema and related editor support for the new stencil card editor

1. Instructions for testing:

1. Closes Issues: [8025](https://zentopia.esri.com/workspaces/sprint-board-60340b9721bc3a4ab6367db5/issues/gh/dc/hub/8025)


packages/common/src/core/schemas/internal/follow/FollowCardSchema.ts
```diff
@@ -0,0 +1,44 @@
+import { IConfigurationSchema } from "../../types";
```
vivzhang, line 1: This is a very basic schema so we could test the new stencil follow card editor, the props are not accurate and will be modified in [8722](https://zentopia.esri.com/workspaces/sprint-board-60340b9721bc3a4ab6367db5/issues/gh/dc/hub/8722)

packages/common/src/core/schemas/internal/follow/FollowCardUiSchema.ts
```diff
@@ -0,0 +1,230 @@
+import { IArcGISContext } from "../../../../ArcGISContext";
```
vivzhang, line 1: Same here

packages/common/src/core/schemas/internal/follow/FollowCardUiSchema.ts
```diff
@@ -0,0 +1,230 @@
...
+              schemaVersion: 1,
+              title: "Esri",
+              scopes: {
+                item: {
+                  targetEntity: "item",
+                  filters: [
+                    {
+                      predicates: [
+                        {
+                          type: "Hub Project",
```
dbouwman, line 36: Do we want this to be `Hub Site Application`?

packages/common/src/core/schemas/internal/follow/FollowCardUiSchema.ts
```diff
@@ -0,0 +1,230 @@
...
+                  predicates: [
+                    {
+                      owner: "paige_pa",
+                    },
+                  ],
+                },
+              ],
+            },
+            {
+              label: "Types",
```
dbouwman, line 95: Since this is limited to Sites, and in the future we'll add Projects & Initiatives... maybe remove the Types facet for now?

packages/common/test/core/schemas/internal/follow/FollowCardUiSchema.test.ts
```diff
@@ -0,0 +1,225 @@
+import { CardEditorOptions } from "../../../../../src/core/schemas/internal/EditorOptions";
+import { buildUiSchema } from "../../../../../src/core/schemas/internal/follow/FollowCardUiSchema";
+import { MOCK_CONTEXT } from "../../../../mocks/mock-auth";
+
+describe("buildUiSchema: follow", () => {
+  it("returns the full follow card uiSchema", async () => {
+    const uiSchema = buildUiSchema("", {} as CardEditorOptions, MOCK_CONTEXT);
+
+    expect(uiSchema).toEqual({
```
dbouwman, line 9: generally speaking, tests that use `.toEqual(...)` with large objects are difficult to manage over time. If one small thing in the structure changes (e.g. the `hub-input-field-gallery-picker` facets), the test will fail with a complex yet vague error message. To fix this failing test, most other devs will simply replace the large object with whatever object is now being passed in _without actually understanding why the test failed_.  

So while it's very efficient to use tests like this, it's not optimal. Instead of doing a full object comparison, consider checking that `elements` exist for a list of scope values, and maybe that those scopes have the expected type of control?

packages/common/src/core/schemas/internal/follow/FollowCardSchema.ts
juliannemarik, line 1: I might suggest naming the following schema something more generic, like `FollowSchema` or `FollowActionSchema` - as we discussed earlier this week, there may be other contexts (that aren't a card) in which we allow you to configure a follow action, and we should reuse this schema. The `uiSchemas` are where we can get more context-specific

packages/common/src/core/schemas/internal/follow/FollowCardSchema.ts
```diff
@@ -0,0 +1,44 @@
...
+      type: "array",
+      items: {
+        type: "string",
+      },
+      maxItems: 10,
+      default: [],
+    },
+    callToAction: {
+      type: "string",
+      default:
```
juliannemarik, line 17: hmm, I don't actually think we have a pattern in place for translating defaults in a schema - this may just need to be passed in from the frontend 🤔 I'll give it some thought

juliannemarik (approved):
Left a minor suggestion, and as you call out, I think the schema/uiSchema can use some polish in a follow-up, but otherwise this is looking good to me!

dbouwman (approved):
Give some thought to refactoring the schema test so it's not just `bigObject === otherBigObject`, and I think it would be a good idea to take a few minutes to update the gallery-picker definition in the ui schema so it's a reasonable mvp (e.g. just show sites)

packages/common/test/core/schemas/internal/follow/FollowCardUiSchema.test.ts
vivzhang: Good point! I will update the tests

packages/common/src/core/schemas/internal/follow/FollowCardSchema.ts
vivzhang: Good point, I'll change this to a more generic name

PR #1396, opened 2024-01-30
fix(): alpha-gate overview and remove extra util

jordantsanz:
1. Description:

Alpha gates the overview pane in the following workspaces:
- [x] Site
- [x] Group
- [x] Page
- [x] Content
- [x] Template
- [x] Initiative Template

 Removes a util for finding the default entity workspace as the logic exists in hub-components already.

1. Instructions for testing:

1. Closes Issues: [#8852](https://zentopia.esri.com/workspaces/sprint-board-60340b9721bc3a4ab6367db5/issues/gh/dc/hub/8852)


MarvinPerry (approved):

packages/common/src/core/getEntityDefaultWorkspacePane.ts
```diff
@@ -1,6 +0,0 @@
-export const getEntityDefaultWorkspacePane = (entityType: string): string => {
```
juliannemarik, line 1: Technically removing this util is a breaking change - I think what we should do for now is just add a deprecation notice letting users know something like:
1. This util is a duplicate and concern of the UI, and was moved as such
2. We will be removing it in the next breaking version

juliannemarik (changes requested):

packages/common/src/core/getEntityDefaultWorkspacePane.ts
jordantsanz: Good call -- am I able to stop exporting it from `core/index.ts`, or would that cause a breaking change as well?

juliannemarik: yeah, I think that would still be considered a breaking change since there _could_ be some other team importing this util - obviously that is not likely/almost certainly not happening...but I do think we still have to consider it a breaking change?

juliannemarik (approved):

PR #1401, opened 2024-02-03
feat: add deepDeletePropByValue and deepEqual util functions

dbouwman:
1. Description:

Add `deepDeletePropByValue(obj, value):obj` and `deepEqual(a,b):boolean` util functions

Hub commonly applies migrations to entities on load. During those migrations, often we want to delete properties to clean things up. However, during the save process, we typically fetch the entity from it's backing store and spread the migrated entity over the top of the fetched entity. This results in the deleted props being re-added.

To avoid this, instead of deleting the props in the migration, we can set them to a specific value (e.g. `remove-this-prop`) and then use `deepDeletePropByValue` function to remove them, after the merge.

`deepEquals` was a useful side-effect of building this function.

This is a precursor to changes in some migrations.

1. Closes Issues: n/a - just did this on a Saturday w/ a lousy weather 🌧️

### DIFF
--- a/packages/common/src/objects/deepDeletePropByValue.ts
+++ b/packages/common/src/objects/deepDeletePropByValue.ts
@@ -1,0 +1,66 @@
+import { deepEqual } from "./deepEqual";
+import { _isObject } from "./_deep-map-values";
+import { isFindable } from "./internal/isFindable";
+
+/**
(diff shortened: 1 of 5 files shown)


tomwayson (approved):
good idea.

I'm surprised we didn't already have deepEqual

PR #1411, opened 2024-02-14
feat: context has user history

dbouwman:
1. Description:

Add methods to context to add and removed entries from the auth'd users history, as well as option to clear the entire history. Also adds `updateUserHubSettings(...)` method as it was needed for the other methods, and streamlines some other use-cases

1. Closes Issues: #9265

### DIFF
--- a/packages/common/src/ArcGISContext.ts
+++ b/packages/common/src/ArcGISContext.ts
@@ -5,13 +5,30 @@ import {
 } from "@esri/arcgis-rest-auth";
 import { IGetUserOptions, IPortal, getUser } from "@esri/arcgis-rest-portal";
 import { IRequestOptions } from "@esri/arcgis-rest-request";
-import { HubServiceStatus } from "./core";
+import {
+  IHubHistory,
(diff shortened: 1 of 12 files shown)


packages/common/src/ArcGISContext.ts
```diff
@@ -225,6 +248,40 @@ export interface IArcGISContext {
...
+
+  /**
+   * Check specific permission for the current user, and optionally an entity
+   * @param permission
+   * @param entity
+   */
+  checkPermission(
+    permission: Permission,
+    entity?: HubEntity
+  ): IPermissionAccessResponse;
```
mjuniper, line 278: This will be convenient. I guess we should generally just call this method instead of importing checkPermission, calling it, and passing it context?

mjuniper (approved):

packages/common/src/ArcGISContext.ts
dbouwman: sure - I'd not really considered that, but it's same-same under the hood :)

PR #1413, opened 2024-02-14
feat(): association group creation form and permission

jordantsanz:
1. Description:

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/internal/getEditorSchemas.ts
+++ b/packages/common/src/core/schemas/internal/getEditorSchemas.ts
@@ -203,6 +203,8 @@ export async function getEditorSchemas(
       const groupModule = await {
         "hub:group:create:followers": () =>
           import("../../../groups/_internal/GroupUiSchemaCreateFollowers"),
+        "hub:group:create:association": () =>
+          import("../../../groups/_internal/GroupUiSchemaCreateAssociation"),
         "hub:group:edit": () =>
           import("../../../groups/_internal/GroupUiSchemaEdit"),
         "hub:group:settings": () =>
--- a/packages/common/src/core/schemas/internal/getEditorSchemas.ts
+++ b/packages/common/src/core/schemas/internal/getEditorSchemas.ts
@@ -5,6 +5,7 @@ import {
   IConfigurationSchema,
   IUiSchema,
   IEditorConfig,
+  IConfigurationValues,
 } from "../types";
 import { filterSchemaToUiSchema } from "./filterSchemaToUiSchema";
 import { SiteEditorType } from "../../../sites/_internal/SiteSchema";
@@ -46,6 +47,7 @@ export async function getEditorSchemas(
   // the entity type and the provided editor type
   let schema: IConfigurationSchema;
   let uiSchema: IUiSchema;
+  let defaults: IConfigurationValues;
   switch (editorType) {
     case "site":
       const { SiteSchema } = await import(
@@ -218,6 +220,12 @@ export async function getEditorSchemas(
         context
       );
 
+      defaults = await groupModule.buildDefaults(
+        i18nScope,
+        options as EntityEditorOptions,
+        context
+      );
+
       break;
 
     case "initiativeTemplate":
@@ -254,5 +262,5 @@ export async function getEditorSchemas(
   // filter out properties not used in the UI schema
   schema = filterSchemaToUiSchema(schema, uiSchema);
 
-  return Promise.resolve({ schema, uiSchema });
+  return Promise.resolve({ schema, uiSchema, defaults });
 }
--- a/packages/common/src/groups/_internal/GroupUiSchemaCreateAssociation.ts
+++ b/packages/common/src/groups/_internal/GroupUiSchemaCreateAssociation.ts
@@ -1,4 +1,8 @@
-import { IUiSchema, UiSchemaRuleEffects } from "../../core/schemas/types";
+import {
+  IConfigurationValues,
+  IUiSchema,
+  UiSchemaRuleEffects,
+} from "../../core/schemas/types";
 import { IArcGISContext } from "../../ArcGISContext";
 import { EntityEditorOptions } from "../../core/schemas/internal/EditorOptions";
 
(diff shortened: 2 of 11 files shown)


packages/common/src/groups/_internal/GroupUiSchemaCreateFollowers.ts
```diff
@@ -115,3 +120,24 @@ export const buildUiSchema = async (
...
+ * @private
+ * constructs the default values for creating a followers group.
+ * This is used to pre-populate the form with specific default values
+ * that are different from the normal Group Schema defualts.
+ * @param i18nScope
+ * @param options
+ * @param context
+ * @returns
+ */
+export const buildDefaults = async (
```
juliannemarik, line 134: not that this is your responsibility, but I believe Ben's work merged, so if you wanted to hook up the `buildDefaults` functions in the `GroupUiSchemaCreateEdit` and `GroupUiSchemaCreateView`, that would be great.

juliannemarik (approved):
Left a few comments for consideration, but otherwise I think this is looking great! I like this new pattern for defaults - I do think it's a bit of a shame that we can't include some of the translation strings as defaults in the `buildDefaults` function - might be worth giving a bit more thought to, but I'm not seeing an easy way around it and am fine moving forward as-is

packages/common/src/core/schemas/internal/getEditorSchemas.ts
```diff
@@ -220,14 +226,23 @@ export async function getEditorSchemas(
         context
       );
 
+      // if we have the buildDefaults fn, then construct the defaults
+      if (groupModule.buildDefaults) {
```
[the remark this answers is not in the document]
jordantsanz: @juliannemarik Only thing with adding these is that it's now adding uncovered branches...and the "right" way to test them in `getEditorSchemas.test.ts` would be to add `buildDefault` functions to a module for each entity...or to pretend that the buildDefaults function exists on each of the entities sometimes...right?

packages/common/src/groups/_internal/GroupUiSchemaCreateAssociation.ts
```diff
@@ -0,0 +1,132 @@
...
+  };
+};
+
+export const buildDefaults = async (
+  i18nScope: string,
+  options: EntityEditorOptions,
+  context: IArcGISContext
+): Promise<IConfigurationValues> => {
+  return {
+    ...getWellKnownGroup("hubAssociationsGroup", context),
```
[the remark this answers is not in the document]
jordantsanz: If we passed the intl up as well into the `getEditorConfig` function, we probably could...but not sure that we'd want to actually do that. But I guess we could?

packages/common/src/associations/initEntityAssociationGroup.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Utility to initialize an entity association group that has just been created.
+ * This adds the entity-specific association keyword to the association group,
+ * and it adds the association definition to the original entity.
+ * @param entity
+ * @param group
+ * @param context
+ * @returns
+ */
+export async function initEntityAssociationGroup(
```
juliannemarik, line 16: nice, this looks great 👍

packages/common/src/groups/_internal/GroupUiSchemaCreateView.ts
```diff
@@ -126,3 +131,13 @@ export const buildUiSchema = async (
     ],
   };
 };
+
```
dbouwman, line 134: please add doc strings

packages/common/src/groups/_internal/GroupUiSchemaCreateEdit.ts
```diff
@@ -123,3 +128,13 @@ export const buildUiSchema = async (
     ],
   };
 };
+
+export const buildDefaults = async (
```
dbouwman, line 132: please add doc strings

packages/common/src/groups/_internal/GroupUiSchemaCreateAssociation.ts
```diff
@@ -0,0 +1,132 @@
...
+              },
+            ],
+          },
+        ],
+      },
+    ],
+  };
+};
+
+export const buildDefaults = async (
```
dbouwman, line 124: docs

packages/common/src/groups/HubGroups.ts
```diff
@@ -119,8 +119,20 @@ export async function createHubGroup(
     group,
     authentication: requestOptions.authentication,
   };
-  const result = await createGroup(opts);
-  return convertGroupToHubGroup(result.group, requestOptions);
+  const createdGroup = await createGroup(opts).then(async (res) => {
```
dbouwman, line 122: seems slightly odd to use `.then` chaining here when it'a an async function already. Consider usign `await`

packages/common/src/associations/initEntityAssociationGroup.ts
```diff
@@ -0,0 +1,43 @@
...
+  entity: HubEntity,
+  group: IHubGroup,
+  context: IArcGISContext
+): Promise<HubEntity> {
+  const type = getTypeFromEntity(entity);
+
+  // 1. Add the association keyword to the group entity
+  group.typeKeywords = [...group.typeKeywords, `${type}|${entity.id}`];
```
dbouwman, line 24: If you add a guard to check if the keyword exists, then this could also work for using an existing group for the association. (or maybe you guard that in the UI layer where the Group selection query precludes groups with this keyword. Consider renaming the function as well b/c it's not "init"-ing the group - it's adding the association definition, and part of that is the keyword... iirc :)

dbouwman (approved):
This looks good - left a few suggestions / clarifications that may be useful, but I'm not as deep in this as you are :)

packages/common/src/initiatives/_internal/InitiativeBusinessRules.ts
```diff
@@ -148,4 +149,10 @@ export const InitiativePermissionPolicies: IPermissionPolicy[] = [
     permission: "hub:initiative:manage",
     dependencies: ["hub:initiative:edit"],
   },
+  // permission to create an association group
+  {
+    permission: "hub:initiative:workspace:associationGroup:create",
+    dependencies: ["hub:initiative:workspace:projects", "hub:group:create"],
+    privileges: ["portal:user:addExternalMembersToGroup"],
```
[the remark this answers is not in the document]
jordantsanz: Good questions. From Jay:

_How broadly granted: That privilege is included in the default Publisher role, the one we recommend for using Hub

As for why it's useful — without that privilege, users can only create groups with
membershipAccess: "org". In theory that's not awful for association groups, though_

From that we've decided to update the permission to not rely on the priv, and instead, we'll update the UI to allow different options for `membershipAccess` depending on if the user has the priv. 

So this will be updated to remove the priv 👍

PR #1414, opened 2024-02-14
feat(hub-common): create feedback entity

juliannaeapicella:
affects: @esri/hub-common

ISSUES CLOSED: 8849

1. Description: Creates HubFeedback entity.

1. Instructions for testing:

1. Closes Issues: #8849


packages/common/src/feedback/HubFeedback.ts
```diff
@@ -0,0 +1,149 @@
...
+} from "..";
+import { HubItemEntity } from "../core/HubItemEntity";
+import { IHubFeedback } from "../core/types";
+import { FeedbackEditorType } from "./_internal/FeedbackSchema";
+import { getEditorConfig } from "../core/schemas/getEditorConfig";
+import { deleteFeedback } from "./edit";
+import { updateFeedback } from "./edit";
+import { enrichEntity } from "../core/enrichEntity";
+
+export class HubFeedback
```
dbouwman, line 19: Please add some JSDoc for the class :)

packages/common/src/feedback/_internal/FeedbackBusinessRules.ts
```diff
@@ -0,0 +1,67 @@
...
+  },
+  {
+    permission: "hub:feedback:delete",
+    authenticated: true,
+    dependencies: ["hub:feedback"],
+    entityOwner: true,
+  },
+  {
+    permission: "hub:feedback:workspace:dashboard",
+    dependencies: ["hub:feedback:view"],
```
dbouwman, line 53: the `:workspace` permissions should likely depend on `hub:feature:workspace` which is the permission connected to the "Workspace Preview" mode... consider looking at how this is done for Sites... where a `hub:site:workspace` permission is defined, which depends on `hub:feature:workspace`, and then all the other workspace permissions depend on `hub:site:workspace`

packages/common/src/feedback/fetch.ts
```diff
@@ -0,0 +1,69 @@
...
+import {
+  IHubFeedback,
+  IHubRequestOptions,
+  IModel,
+  fetchModelFromItem,
+  fetchSetting,
+  getDefaultEntitySettings,
+  getItemBySlug,
+  isGuid,
+} from "..";
```
dbouwman, line 11: Ensure you import from files - importing from folders or relative paths seems to cause gnarly problems with the actual exports

dbouwman (changes requested):
Generally looks great - please adjust the permissions so it connects into `hub:feature:workspace` (see comment) and review all files for imports that are not from files - this seems to be the cause (or at least a cure) for really gnary issues around missing exports. Thanks!

packages/common/src/feedback/fetch.ts
```diff
@@ -0,0 +1,69 @@
...
+    entitySettings = await fetchSetting({ id: item.id, ...requestOptions });
+  } catch (e) {
+    const defaultSettings = getDefaultEntitySettings("feedback");
+    entitySettings = {
+      id: null,
+      ...defaultSettings,
+    };
+  }
+  model.entitySettings = entitySettings;
+  const mapper = new PropertyMapper<Partial<IHubFeedback>, IModel>(
```
rweber-esri, line 64: In order to make the determination if a map question is configured for the survey, we'll need to fetch the form's configuration from within this method. The form configuration can live in a couple different places depending on whether we're working with a survey created using the Survey123 web application or the Survey123 Connect desktop application. A [getFormJSON](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L202) utility exists that abstracts the differences of retrieving the configuration for a form created in Survey123 web app vs Survey123 Connect. We should be able to avoid hoisting the [getInfoFile](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L290) utility used by `getFormJSON` in lieu of using the [getItemInfo](https://github.com/Esri/arcgis-rest-js/blob/main/packages/arcgis-rest-portal/src/items/get.ts#L378) RESTJS method that was added _after_ the survey feature was created.

I envision our code in this function looking something like:

```ts
model.formJSON = await getFormJson(item.id, requestOptions);
```

packages/common/src/feedback/_internal/computeProps.ts
```diff
@@ -0,0 +1,46 @@
...
+    requestOptions,
+    token
+  );
+
+  // Handle Dates
+  feedback.createdDate = new Date(model.item.created);
+  feedback.createdDateSource = "item.created";
+  feedback.updatedDate = new Date(model.item.modified);
+  feedback.updatedDateSource = "item.modified";
+  feedback.isDiscussable = isDiscussable(feedback);
```
rweber-esri, line 42: We'll need to compute a couple custom properties here, though after thinking about this more, we won't actually need the `questions` on the entity. We should make a minor tweak/update to the [Hub Class Model](https://lucid.app/lucidchart/066ce98b-9aad-43d1-b7bd-36f5e89d0293/edit?invitationId=inv_1138f4b7-7396-4d29-ae18-06f287ce7eb7&page=E7lKAams_xlW#), specifically replace the `questions` member with the property enumerated in the second bullet point below.

1. `hasMapQuestion` - value is `true` when a map question exists for the survey, `false` when no map question exists for the survey (will control whether the `Map display` section render in the `settings` workspace pane, per [figma](https://www.figma.com/file/1bvSoMV4ygNqHGXmkIBgJd/Survey-workspace?type=design&node-id=8-2102&mode=design&t=lJwoAoQJBO6OR96C-0))
2. `displayMode` - value is `"map"` when `"hubMapSurvey"` is present in `typeKeywords`, `"standard"` when `hubMapSurvey` is absent in typeKeywords. (will control which tile select controls is ticked/checked, per [figma]([figma](https://www.figma.com/file/1bvSoMV4ygNqHGXmkIBgJd/Survey-workspace?type=design&node-id=8-2102&mode=design&t=lJwoAoQJBO6OR96C-0)). I'm open to suggestions for naming here)

The `questions` array from the form configuration can include nested structures to support multiple pages of questions. [hasMapQuestion](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L767) [getMapQuestion](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L776), [isMapQuestion](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L754) & [isPageQuestion](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L90) utilities exist to determine if a map question exists and will need to be ported/hoisted as part of this effort.

A [isMapSurvey](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-hub-components/addon/utils/item-types/survey-utils.js#L1005) util exists that will also need ported/hoisted to hub.js but renamed/tweaked to return the expected values enumerated in the second bullet point above.

These ported/hoisted utils can likely live in a new `packages/common/feedback/utils` dir.

I envision the code in this file looking something like:

```ts
feedback.hasMapQuestion = hasMapQuestion(model.formJSON.questions);
feedback.displayMode = getDisplayMode(model.item);
```

packages/common/src/feedback/edit.ts
```diff
@@ -0,0 +1,68 @@
...
+  feedback.slug = await getUniqueSlug(
+    { slug: feedback.slug, existingId: feedback.id },
+    requestOptions
+  );
+  // update the status keyword
+  feedback.typeKeywords = setEntityStatusKeyword(
+    feedback.typeKeywords,
+    feedback.status
+  );
+  feedback.typeKeywords = setDiscussableKeyword(
```
rweber-esri, line 31: We'll need a similar `setDisplayModeKeyword` util here that either adds or removes `hubMapSurvey` based on the value of `feedback.displayMode` here

```ts
feedback.typeKeywords = setDisplayModeKeyword(
  feedback.typeKeywords,
  feedback.displayMode
);
```

packages/common/src/core/schemas/internal/getEditorSchemas.ts
```diff
@@ -217,6 +218,25 @@ export async function getEditorSchemas(
...
+      );
+      schema = cloneObject(FeedbackSchema);
+
+      const feedbackModule = await {
+        "hub:feedback:edit": () =>
+          import("../../../feedback/_internal/FeedbackUiSchemaEdit"),
+        "hub:feedback:settings": () =>
+          import("../../../feedback/_internal/FeedbackUiSchemaSettings"),
+      }[type as FeedbackEditorType]();
+      uiSchema = await groupModule.buildUiSchema(
```
rweber-esri, line 234: I believe we want to call `feedbackModule.buildUiSchema` here?

packages/common/src/feedback/_internal/FeedbackSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+  "hub:feedback:settings",
+] as const;
+
+/**
+ * defines the JSON schema for a Feedback entity's editable fields
+ */
+export const FeedbackSchema: IConfigurationSchema = {
+  ...HubItemEntitySchema,
+  properties: {
+    displayMap: {
```
rweber-esri, line 16: This property should be named whatever we settle on in my [comment](https://github.com/Esri/hub.js/pull/1414/files#r1491448517) above per the second bullet point. Again, open to suggestions on naming and/or values. Options are:

`displayMap: boolean`, `displayMode: "standard" | "map"`, `isMapFeedback: boolean`

packages/common/src/feedback/_internal/FeedbackSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+      type: "boolean",
+      enum: [true, false],
+      default: false,
+    },
```
rweber-esri, line 20: We may need to add the other property `hasMapQuestion` to the schema for the UISchema to make the determination if the `Map display` section renders per [figma](https://www.figma.com/file/1bvSoMV4ygNqHGXmkIBgJd/Survey-workspace?type=design&node-id=8-2102&mode=design&t=lJwoAoQJBO6OR96C-0)

packages/common/src/feedback/_internal/FeedbackUiSchemaSettings.ts
```diff
@@ -0,0 +1,44 @@
...
+export const buildUiSchema = async (
+  i18nScope: string,
+  options: EntityEditorOptions,
+  context: IArcGISContext
+): Promise<IUiSchema> => {
+  return {
+    type: "Layout",
+    elements: [
+      {
+        type: "Section",
```
rweber-esri, line 20: We'll likely want to add a `rule` here to conditionally render this section when `hasMapQuestion` from the schema is `true`. This will likely look something like:

```ts
        rule: {
          effect: UiSchemaRuleEffects.SHOW,
          condition: {
            scope: '/properties/hasMapQuestion',
            schema: { const: true }
          }
        },
```

Disclaimer: I've not worked with rules yet and have not tested ^. You can see some other examples of rule definitions [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/schemas/internal/metrics/StatCardUiSchema.ts#L329)

packages/common/src/feedback/edit.ts
```diff
@@ -0,0 +1,76 @@
...
+ * @private
+ * Remove a Hub Feedback object
+ * @param id
+ * @param requestOptions
+ */
+export async function deleteFeedback(
+  id: string,
+  requestOptions: IUserRequestOptions
+): Promise<void> {
+  const ro = { ...requestOptions, ...{ id } } as IUserItemOptions;
```
rweber-esri, line 73: I overlooked this until now, but there's a _lot_ of complexity related to successfully deleting a survey as compared to other entity types. This is because surveys are composed of up to 4 different items, which all live inside a folder in AGO, all of which need to be deleted (including the folder). Further complicating things, these items need to be deleted in a specific order, otherwise delete operations can fail. Additionally, there's the concept of "relationships" supported by the AGO platform which we lean on to identify/fetch the items. Also, items can be "protected" which means they can't be deleted without first issuing another XHR request to first remove that delete prevention. Given these complexities, we're likely going to need to port ~200 other lines of code to perform this seemingly simple task..

You can see what I mean by taking a look at the survey-service [removeItem](https://github.com/ArcGIS/opendata-ui/blob/master/packages/ember-arcgis-opendata-components/addon/services/survey-service.js#L741) method and the nightly E2E purge script's [deleteSurvey](https://github.com/ArcGIS/opendata-ui/blob/master/tests/hub-e2e/lib/helpers/survey.js#L40) method... 🤯 

To the best of my knowledge, we don't actually delete surveys from within the Hub UI anywhere, so I cannot speak to the reliability of the survey-service `removeItem` method, but I personally sunk a bunch of time into the nightly purge script's `deleteSurvey` to get that working as reliably as I could, though it looks like that is not removing delete prevention in it's current state (most likely b/c it's only targeting surveys created by E2E tests which wouldn't have delete prevention). We may want to consult with @dbouwman on whether we want/need to remove delete prevention here, maybe we don't need that piece given users need to explicitly add delete prevention from within the AGO UI...

rweber-esri: I verified we aren't currently handling the case in settings workspace panes where delete prevention is enabled for the entity item. The `delete` button is enabled, no call is made to remove delete prevention, and consequently the AGO API `delete` endpoint serves an error response, but the error is suppressed, and nothing happens in the UI. Options here include:

1. Disable `delete` button when entity's `protected` property is true + show a notice instructing the user to remove delete prevention
2. Call the endpoint to remove delete prevention before calling the `delete` endpoint

We may need direction from Dave and/or Sam on how to proceed. If the latter is preferred, we may need to create issues for the other entity types to update their delete logic.

We'll additionally need to consider how to approach for surveys given there are up to 4 items that could each have delete prevention enabled if the desire is the former.

dbouwman (approved):
Please take another look for imports that VSCode injected that ref folders vs files. Thanks!

packages/common/src/feedback/edit.ts
rweber-esri: Resolving this convo. From the sounds of Brian's comment [here](https://devtopia.esri.com/dc/hub/issues/8849#issuecomment-4595995) it sounds like we're only on the hook for deleting the `form` item, not the feature service, either views, or the containing folder. Also it sounds like folks are already considering a pattern related to protected items, so we probably don't need to be concerned about that at this time either.

packages/common/src/feedback/utils.ts/set-display-map-keyword.ts
```diff
@@ -0,0 +1,20 @@
+import { MAP_SURVEY_TYPEKEYWORD } from "../constants";
```
rweber-esri, line 1: The folder containing this file is named `utils.ts` but should probably be named just `utils`.

The `packages/common/index.ts` file does not export the contents from the `packages/common/feedback` dir.

We should probably consider consolidating the existing `surveys` and `feedback` directories into just `feedback`.
1. Move modules in `packages/common/surveys/utils` to `packages/common/feedback/utils` (move tests as necessary too)
2. Move `get-*` & `is-*` modules in `packages/common/surveys` to `packages/common/feedback/utils` (move tests as necessary too)
3. Update `packages/common/feedback/utils/index.ts` to export all modules from that dir
4. Update `packages/common/feedback/index.ts` to export `utils`
5. Update export in `packages/common/index.ts` from `"./surveys"` to `"./feedback"`

packages/common/src/core/types/IHubFeedback.ts
```diff
@@ -0,0 +1,13 @@
...
+import { IHubItemEntity, IHubItemEntityEditor } from "./IHubItemEntity";
+
+/**
+ * Defines properties of Hub Feedback object
+ * @internal
+ */
+export interface IHubFeedback
+  extends IHubItemEntity,
+    IWithPermissions,
+    IWithSlug {}
```
rweber-esri, line 11: We don't _currently_ support slugs for Feedback items, and since we won't be supporting creating Feedback items in Hub any longer, we can probably drop `IWithSlug` here

packages/common/src/feedback/_internal/FeedbackSchema.ts
```diff
@@ -0,0 +1,29 @@
+import { IConfigurationSchema } from "../../core";
+import { HubItemEntitySchema } from "../../core/schemas/shared/HubItemEntitySchema";
+import { hasMapQuestion } from "../../surveys/utils";
```
rweber-esri, line 3: Unused import?

packages/common/test/surveys/get-input-feature-service-model.test.ts
```diff
@@ -1,7 +1,7 @@
 /* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
  * Apache-2.0 */
 
-import { getInputFeatureServiceModel } from "../../src/surveys/get-input-feature-service-model";
+import { getInputFeatureServiceModel } from "../../src/feedback/utils/get-input-feature-service-model";
```
rweber-esri, line 4: Thanks for consolidating the `surveys` and `feedback` dirs, though we may also want to move these tests from `test/surveys` => `test/feedback/utils` to better conform to our organizational pattern.

rweber-esri (approved):
This looks pretty good to me pending:
1. Confirmation no imports from dirs per @dbouwman [comment](https://github.com/Esri/hub.js/pull/1414#pullrequestreview-1891245225)
2. Acknowledgement/acceptance from Brian per [my response](https://devtopia.esri.com/dc/hub/issues/8849#issuecomment-4598172) to his [question](https://devtopia.esri.com/dc/hub/issues/8849#issuecomment-4598080) regarding surveys having their own entity & workspace path
3. Successful `opendata-ui` E2E run with a PR-Preview of this branch to make sure we didn't break any imports/functionality from moving/consolidating things `surveys` and `feedback` dirs
4. Manual verification that feedback workspace works as expected w/ a PR-Preview of this branch
5. 100% test coverage

packages/common/src/surveys/_internal/SurveyUiSchemaSettings.ts
```diff
@@ -0,0 +1,51 @@
...
+          },
+        },
+        labelKey: `${i18nScope}.sections.settings.label`,
+        elements: [
+          {
+            labelKey: `${i18nScope}.fields.displayMap.label`,
+            scope: "/properties/displayMap",
+            type: "Control",
+            options: {
+              control: "hub-field-input-radio",
```
rweber-esri, line 35: I think we want `hub-field-tile-select` here vs radio

packages/common/src/surveys/_internal/SurveyUiSchemaSettings.ts
```diff
@@ -0,0 +1,52 @@
...
+              type: "radio",
+              labels: [
+                `{{${i18nScope}.fields.displayMap.enabled.label:translate}}`,
+                `{{${i18nScope}.fields.displayMap.disabled.label:translate}}`,
+              ],
+              descriptions: [
+                `{{${i18nScope}.fields.displayMap.enabled.description:translate}}`,
+                `{{${i18nScope}.fields.displayMap.disabled.description:translate}}`,
+              ],
+              icons: ["sidecard", "form-elements"],
```
rweber-esri, line 45: The first icon isn't rendering as expected, we want `sidecar` v `sidecard` here

packages/common/src/surveys/_internal/SurveyUiSchemaSettings.ts
```diff
@@ -0,0 +1,52 @@
...
+        },
+        labelKey: `${i18nScope}.sections.settings.label`,
+        elements: [
+          {
+            labelKey: `${i18nScope}.fields.displayMap.label`,
+            scope: "/properties/displayMap",
+            type: "Control",
+            options: {
+              control: "hub-field-input-tile-select",
+              type: "radio",
```
rweber-esri, line 36: Per figma, the tiles should have a horizontal layout. Think we just need to add `layout: 'horizontal',` to this `options` object

rweber-esri (commented):
We'll need to add some survey-specific logic to the `getRelativeEntityViewUrl` method in `packages/hub-components/src/utils/urls.ts` to build the appropriate URL to the live view of the survey. Changes should likely look something like:

```js
} else if (type === 'survey') {
    result = `/feedback/${type}s/${entity.id}`
  }
```

packages/common/src/core/types/IHubSurvey.ts
```diff
@@ -0,0 +1,10 @@
+import { IWithPermissions } from "../traits";
+import { IHubItemEntity, IHubItemEntityEditor } from "./IHubItemEntity";
+
+/**
+ * Defines properties of Hub Survey object
+ * @internal
+ */
+export interface IHubSurvey extends IHubItemEntity, IWithPermissions {}
```
rweber-esri, line 8: We'll want to define the `hasMapQuestion` and `displayMap` properties on `IHubSurvey` here.

```ts
export interface IHubSurvey extends IHubItemEntity, IWithPermissions {
  hasMapQuestion: boolean;
  displayMap: boolean;
}
```

packages/common/src/surveys/_internal/SurveyUiSchemaSettings.ts
```diff
@@ -0,0 +1,53 @@
...
+  i18nScope: string,
+  options: EntityEditorOptions,
+  context: IArcGISContext
+): Promise<IUiSchema> => {
+  return {
+    type: "Layout",
+    elements: [
+      {
+        type: "Section",
+        rule: {
```
rweber-esri, line 21: After looking more closely at [figma](https://www.figma.com/file/1bvSoMV4ygNqHGXmkIBgJd/Survey-workspace?type=design&node-id=0-1&mode=design&t=hAqQjRwSraKjf8OR-0), we don't want to conditionally render this section, we instead want to _always_ render this section, but _disable_ the tile-select when no map question is configured.  We also need to conditionally render a calcite notice when no map question is configured. This will involve:

1. Move this entire `rule` to the 0-index child element of this section that defines the tile-select control
2. Change the `effect` to `UiSchemaRuleEffects.DISABLE`
3. Define a `messages` array that is empty when `options.hasMapQuestion` is truthy, else an array containing a single message config object when `options.hasMapQuestion` is falsey.
4. Expanding the IUiSchemaMessage in `packages/common/src/core/schemas/types.ts` to define optional `title` and `titleKey` properties (IUiSchemaMessage currently only support defining the calcite-notice `message` slot value which is mapped from `label` or `labelKey` props on this interface). This will also require us to do a little bit more work in opendata-ui to support, adding specifics to your other PR.

```ts
// packages/common/src/surveys/_internal/SurveyUiSchemaSettings.ts

export const buildUiSchema = async (
  i18nScope: string,
  options: EntityEditorOptions,
  context: IArcGISContext
): Promise<IUiSchema> => {
  return {
    type: "Layout",
    elements: [
      {
        type: "Section",
        labelKey: `${i18nScope}.sections.settings.label`,
        elements: [
          {
            labelKey: `${i18nScope}.fields.displayMap.label`,
            scope: "/properties/displayMap",
            type: "Control",
            rule: {
              effect: UiSchemaRuleEffects.DISABLE,
              condition: {
                scope: "/properties/hasMapQuestion",
                schema: { const: false },
              },
            },
            options: {
              control: "hub-field-input-tile-select",
              type: "radio",
              labels: [
                `{{${i18nScope}.fields.displayMap.enabled.label:translate}}`,
                `{{${i18nScope}.fields.displayMap.disabled.label:translate}}`,
              ],
              descriptions: [
                `{{${i18nScope}.fields.displayMap.enabled.description:translate}}`,
                `{{${i18nScope}.fields.displayMap.disabled.description:translate}}`,
              ],
              icons: ["sidecar", "form-elements"],
              layout: "horizontal",
              messages: (options as IHubSurvey).hasMapQuestion
                ? []
                : [
                    {
                      type: UiSchemaMessageTypes.custom,
                      display: "notice",
                      keyword: "mapQuestion",
                      titleKey: `${i18nScope}.fields.displayMap.notice.title`,
                      labelKey: `${i18nScope}.fields.displayMap.notice.message`,
                      allowShowBeforeInteract: true,
                      alwaysShow: true,
                    },
                  ],
            },
          },
        ],
      },
    ],
  };
};
```

```ts
// packages/common/src/core/schemas/types.ts

export interface IUiSchemaMessage {
  type: UiSchemaMessageTypes;
  display?: "message" | "notice";
  keyword?: string;
  title?: string;
  titleKey?: string;
  label?: string;
  labelKey?: string;
  icon?: boolean | string;
  kind?: "brand" | "danger" | "info" | "success" | "warning";
  hidden?: boolean;
  condition?: IUiSchemaCondition;
  allowShowBeforeInteract?: boolean;
  alwaysShow?: boolean;
}
```

PR #1424, opened 2024-02-28
feat(): implement assertion conditions

jordantsanz:
1. Description:
Adds conditions to assertion logic. A condition dictates if an assertion should be checked or not. If the condition evaluates to false, the assertion can be ignored. 

We implement this logic at the checkAssertions level, as we want to determine if an assertion is relevant first before evaluating it. 

Use case: Groups with `isViewOnly: false` allow any group member to share content; groups with `isViewOnly: true` allow only group admins to share content. To use just one permission for sharing content, we use a condition on `entity.isViewOnly` to understand which assertion currently applies: `is-group-member` or `is-group-admin`. 

1. Instructions for testing:

1. Closes Issues: #9090 (if appropriate)

### DIFF
--- a/packages/common/src/permissions/_internal/checkAssertions.ts
+++ b/packages/common/src/permissions/_internal/checkAssertions.ts
@@ -20,11 +20,11 @@ export function checkAssertions(
     // iterate over the assertions, creating a check for each entry
     checks = policy.assertions.reduce(
       (acc: IPolicyCheck[], assertion: IPolicyAssertion) => {
-        let conditionResult = true;
+        let shouldCheckAssertion = true;
 
         // if conditions, check them first
         if (assertion.conditions?.length) {
-          conditionResult = assertion.conditions.every(
+          shouldCheckAssertion = assertion.conditions.every(
             (condition: IPolicyAssertion) =>
               checkAssertion(condition, entity, context).response === "granted"
           );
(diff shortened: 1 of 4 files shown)


packages/common/src/permissions/_internal/checkAssertions.ts
```diff
@@ -18,9 +18,29 @@ export function checkAssertions(
   // Only return a check if the policy is defined
   if (policy.assertions) {
     // iterate over the assertions, creating a check for each entry
-    checks = policy.assertions.map((assertion: IPolicyAssertion) => {
-      return checkAssertion(assertion, entity, context);
-    });
+    checks = policy.assertions.reduce(
+      (acc: IPolicyCheck[], assertion: IPolicyAssertion) => {
+        let conditionResult = true;
```
juliannemarik, line 23: minor, but I might suggest naming this something like `shouldCheckAssertion` to make it a bit more clear?

juliannemarik (approved):
This looks good to me and I _believe_ achieves what we need, but let's check with @dbouwman to make sure he's okay with this enhancement to permission assertions 👍

jordantsanz:
Requesting @dbouwman for review to confirm he's okay with these enhancements to assertions.

dbouwman (approved):
This looks like a good addition!

PR #1427, opened 2024-03-04
feat: org thumbnail

mjuniper:
1. Description:
- adds getOrgThumbnailUrl function
- adds orgThumbnailUrl to IArcGISContext

1. Instructions for testing: run the tests

### DIFF
--- a/packages/common/src/resources/get-portal-thumbnail-url.ts
+++ b/packages/common/src/resources/get-portal-thumbnail-url.ts
@@ -1,0 +1,26 @@
+import { IPortal } from "@esri/arcgis-rest-portal";
+
+/**
+ * Construct a the full url to a portal thumbnail
+ *
+ * - If the portal has a thumbnail, construct the full url
+ * - If the portal is not public, append on the token
+ * @param portalUrl
+ * @param portal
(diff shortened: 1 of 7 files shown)


packages/common/src/resources/get-portal-thumbnail-url.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns
+ */
+export function getPortalThumbnailUrl(
+  portalUrl: string,
+  portal: IPortal,
+  token?: string
+): string {
+  let thumbnailUrl = null;
+  if (portal?.thumbnail) {
+    thumbnailUrl = `${portalUrl}/portals/self/resources/${portal.thumbnail}`;
```
dbouwman, line 20: I _think_ the `/self` is ok, but since you're passing in the portal, maybe use `/${portal.id}` as this would then work across orgs? Also consider constructing the portalUrl from the portal in here b/c it's got the info needed - `getPortalUrl(portal)` will do the work... this makes it a little nicer to use elsewhere

dbouwman (approved):
Left some minor suggestions, but what's here is also fine

PR #1442, opened 2024-03-18
feat: add migration for web mapping application sites

juliannemarik:
[8160](https://devtopia.esri.com/dc/hub/issues/8160)

### Description:

### DIFF
--- a/packages/common/src/sites/_internal/migrateWebMappingApplicationSites.ts
+++ b/packages/common/src/sites/_internal/migrateWebMappingApplicationSites.ts
@@ -1,0 +1,20 @@
+import { getProp, getWithDefault } from "../../objects";
+import { IDraft, IModel } from "../../types";
+
+export function migrateWebMappingApplicationSites<T extends IModel | IDraft>(
+  model: T
+): T {
+  const isWebMappingApplication =
+    getProp(model, "item.type") === "Web Mapping Application";
+  const hasHubSiteKeyword = getWithDefault(
+    model,
+    "item.typeKeywords",
+    []
+  ).includes("hubSite");
+
+  if (isWebMappingApplication && hasHubSiteKeyword) {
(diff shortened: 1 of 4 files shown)


packages/common/src/sites/_internal/computeLinks.ts
```diff
@@ -0,0 +1,37 @@
...
+import { getHubRelativeUrl } from "../../content/_internal/internalContentUtils";
+
+/**
+ * Compute the links that get appended to a Hub Site
+ * search result and entity
+ *
+ * @param item
+ * @param requestOptions
+ */
+export function computeLinks(
```
juliannemarik, line 17: Took the opportunity to abstract this into a util like we're doing for other entities - this ensures that the links that are computed when we enrich site search results are the same as the ones we append to the entity in `computeProps`

tomwayson (approved):
mainly I wanted to verify the changes to fetching content entites, but I also glanced over the rest and all LGTM

packages/common/src/core/HubItemEntity.ts
```diff
@@ -236,10 +236,19 @@ export abstract class HubItemEntity<T extends IHubItemEntity>
    * Returns the followers group
    */
   async getFollowersGroup(): Promise<IGroup> {
-    return getGroup(
-      this.entity.followersGroupId,
-      this.context.userRequestOptions
-    );
+    let group = {} as IGroup;
```
dbouwman, line 239: I'm not super wild about this approach b/c consumers need to check more than just "is the returned object defined"... was there a reason to not let `getGroup` throw?

dbouwman (approved):
generally LGTM, other than returning `{} as IGroup` b/c that means the upstream code must do some check other than `if (group) {...` If there's a really compelling reason to do this, I can live with it, but feels like an anti-pattern

PR #1445, opened 2024-03-21
feat: add canEditAccess permission to all entities

juliannemarik:
[9580](https://devtopia.esri.com/dc/hub/issues/9580)

### Description:

- extends the permission assertion types to support "contains-some"
- adds a new `hub:{entity}:canChangeAccess` to all entities based on the following logic to determine whether the current user can edit the access of the entity:

    **For item entities:**
    1. user has the (`portal:admin:shareToOrg` OR `portal:admin:shareToPublic`) OR (`portal:user:shareToOrg` OR `portal:user:shareToPublic`) privileges
    AND
    3. user has admin privileges over the item (`itemControl:admin`)
    
    **For group entities:**
    1. user has the `portal:admin:updateGroups` privilege
    OR
    4. user has admin or owner membership in the group

### DIFF
--- a/packages/common/src/content/_internal/ContentBusinessRules.ts
+++ b/packages/common/src/content/_internal/ContentBusinessRules.ts
@@ -17,6 +17,7 @@ export const ContentPermissions = [
   "hub:content:delete",
   "hub:content:edit",
   "hub:content:view",
+  "hub:content:canEditAccess",
   "hub:content:workspace",
   "hub:content:workspace:overview",
   "hub:content:workspace:dashboard",
@@ -33,6 +34,10 @@ export const ContentPermissions = [
  * @private
  */
 export const ContentPermissionPolicies: IPermissionPolicy[] = [
+  {
+    permission: "hub:content",
+    services: ["portal"],
+  },
   {
     permission: "hub:content:create",
     services: ["portal"],
@@ -55,6 +60,28 @@ export const ContentPermissionPolicies: IPermissionPolicy[] = [
     authenticated: true,
     services: ["portal"],
     entityEdit: true,
+  },
+  {
+    permission: "hub:content:canEditAccess",
+    dependencies: ["hub:content"],
+    authenticated: true,
+    assertions: [
+      {
+        property: "context:currentUser.privileges",
+        type: "contains-some",
+        value: [
+          "portal:admin:shareToPublic",
+          "portal:admin:shareToOrg",
+          "portal:user:shareToPublic",
+          "portal:user:shareToOrg",
+        ],
+      },
+      {
+        property: "entity:itemControl",
+        type: "eq",
+        value: "admin",
+      },
+    ],
   },
   {
     permission: "hub:content:workspace",
--- a/packages/common/src/content/_internal/ContentBusinessRules.ts
+++ b/packages/common/src/content/_internal/ContentBusinessRules.ts
@@ -17,7 +17,7 @@ export const ContentPermissions = [
   "hub:content:delete",
   "hub:content:edit",
   "hub:content:view",
-  "hub:content:canEditAccess",
+  "hub:content:canChangeAccess",
   "hub:content:workspace",
   "hub:content:workspace:overview",
   "hub:content:workspace:dashboard",
@@ -62,7 +62,7 @@ export const ContentPermissionPolicies: IPermissionPolicy[] = [
     entityEdit: true,
   },
   {
-    permission: "hub:content:canEditAccess",
+    permission: "hub:content:canChangeAccess",
     dependencies: ["hub:content"],
     authenticated: true,
     assertions: [
--- a/packages/common/src/discussions/_internal/DiscussionBusinessRules.ts
+++ b/packages/common/src/discussions/_internal/DiscussionBusinessRules.ts
@@ -11,6 +11,7 @@ export const DiscussionPermissions = [
   "hub:discussion:delete",
   "hub:discussion:edit",
   "hub:discussion:view",
+  "hub:discussion:canEditAccess",
   "hub:discussion:workspace:overview",
   "hub:discussion:workspace:dashboard",
   "hub:discussion:workspace:details",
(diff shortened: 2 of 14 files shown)


packages/common/src/groups/_internal/GroupBusinessRules.ts
```diff
@@ -80,6 +81,23 @@ export const GroupPermissionPolicies: IPermissionPolicy[] = [
...
     entityOwner: true,
   },
+  {
+    permission: "hub:group:canEditAccess",
+    dependencies: ["hub:group"],
+    authenticated: true,
+    assertions: [
+      {
+        property: "context:currentUser.privileges",
+        type: "contains-some",
```
jordantsanz, line 91: Could we use "contains" here, since we only have one thing we're checking for ("portal:admin:updateGroups")?

packages/common/src/groups/_internal/GroupBusinessRules.ts
```diff
@@ -80,6 +81,23 @@ export const GroupPermissionPolicies: IPermissionPolicy[] = [
...
+        value: ["portal:admin:updateGroups"],
+      },
+      {
+        property: "context:currentUser",
+        type: "is-group-admin",
+        value: "entity:id",
+      },
```
jordantsanz, line 98: I may be reading the issue wrong, so totally correct me if I'm off here -- but Jay said this towards the bottom: 

> Hence, I think group sharing should be limited to (userMembership.membershipType: "admin" OR userMembership.membershipType: "owner" OR portal:admin:updateGroups)

I think, with how we handle assertions, your addition here currently reads: 
`(userMembership.membershipType: "admin" OR userMembership.membershipType: "owner") AND portal:admin:updateGroups`

Again, might be totally off here, but I think we'd need to combine these assertions in a way so that we have them joined by or's. Which all gets really complex 😅 I'm not sure that would be possible right now without adding something new, like an operator for all of the assertions....or creating separate permissions. 

Can a user even have the `updateGroups` permission without being an admin or an owner?

jordantsanz (commented):
Looks good, nice job with the `contains-some` idea! I left two comments around the group implementation -- one is smaller, but one could be a bit more of an issue if I'm understanding things correctly 😅 let me know if you want to hop on a call to discuss!

packages/common/src/groups/_internal/GroupBusinessRules.ts
juliannemarik: Oh man, nice catch - I definitely missed that last `OR`. And that does significantly complicate this 😅 
I'm trying to think if I could use conditions in any way on these assertions, but I don't _think_ that gets us all the way. Hmmm, being able to specify an operator is a good thought, but that does seem like a heavier lift here. 

> Can a user even have the updateGroups permission without being an admin or an owner?
I think the answer is technically yes, even though that doesn't make too much sense, but I think we have to account for it. 

creating separate permissions is probably the easiest solution for now, but I don't love that - it would be nice if all entities had this same permission to key off of. I'll keep pondering

juliannemarik: just spitballing, so I could be way off here, but with conditions, wondering if we'd have to do something like this: 

```ts
{
  permission: "hub:group:canEditAccess",
  dependencies: ["hub:group"],
  authenticated: true,
  assertions: [
    {
      conditions: [
        {
          property: "context:currentUser",
          type: "is-not-group-admin", // this type doesn't currently exist but would be easy to add and would basically just check if you're a "member"
          value: "entity:id",
        },
      ],
      property: "context:currentUser.privileges",
      type: "contains",
      value: ["portal:admin:updateGroups"],
    },
    {
      conditions: [
        {
          property: "context:currentUser.privileges",
          type: "without",
          value: ["portal:admin:updateGroups"],
        }
      ],
      property: "context:currentUser",
      type: "is-group-admin",
      value: "entity:id",
    },
  ],
},
```

Still parsing this myself, and it does seem kind of redundant, but does this give us what we need? 😅

jordantsanz: I think so -- that was along the idea I was playing around with yesterday as well! 

The first assertion is: if someone isn't an admin, do they have the updateGroups priv?
The second assertion is: if someone doesn't have the updateGroups priv, are they an admin?

If someone is an admin && has the updateGroups priv, neither assertion would be "checked"...but I think that's okay, because in that case, the permission would come back with access: true, since all assertions (none) pass...right?

juliannemarik: yes, that's how I read it as well and think that all sounds correct!

packages/common/src/content/_internal/ContentBusinessRules.ts
```diff
@@ -17,6 +17,7 @@ export const ContentPermissions = [
   "hub:content:delete",
   "hub:content:edit",
   "hub:content:view",
+  "hub:content:canEditAccess",
```
dbouwman, line 20: not a big thing, but I was scratching my head for a second to understand what this was actually doing... maybe `:canChangeAccess`?

packages/common/src/discussions/_internal/DiscussionBusinessRules.ts
```diff
@@ -58,6 +59,28 @@ export const DiscussionPermissionPolicies: IPermissionPolicy[] = [
...
   },
+  {
+    permission: "hub:discussion:canEditAccess",
+    dependencies: ["hub:discussion"],
+    authenticated: true,
+    assertions: [
+      {
+        property: "context:currentUser.privileges",
+        type: "contains-some",
+        value: [
```
dbouwman, line 70: I assume the permission is used to show/hide a component, which then looks at the privs to decide if "Org" and "Everyone" are enabled...

dbouwman (approved):
This all looks 💯 - only suggestion might be to consider changing the permission to `:canChangeAccess` as `"Edit"` feels overloaded... but it's also not 😱 - your call. Thanks!

packages/common/src/discussions/_internal/DiscussionBusinessRules.ts
juliannemarik: yes, that is correct 👍

PR #1447, opened 2024-03-25
feat: post with enrichments

drspacemanphd:
1. Description: This PR modifies the `IHubSearchResult` created for discussion posts to include other post and channel data.

2. Instructions for testing:
- Use the preview to search discussions 

3. Closes Issues: 8936

4. [x] Updated meaningful TSDoc to methods including Parameters and Returns, see [Documentation Guide](https://esri.github.io/hub-components/storybook/?path=/story/guides-documentation--page)

5. [x] used semantic commit messages
  
6. [x] PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

7. [x] updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.
- I dont believe this was necessary


packages/common/src/search/_internal/hubSearchItemsHelpers/ogcItemToDiscussionPostResult.ts
```diff
@@ -30,5 +30,41 @@ export async function ogcItemToDiscussionPostResult(
     rawResult: ogcItem.properties,
     access: null,
     family: null,
+
+    // Post properties
+    channelId: ogcItem.properties.channelId,
```
sonofflynn89, line 35: I don't know that we want to expose all these top level properties on the interface, especially since they only belong to a single entity type. Is there a problem with having UI layer just use `rawResult.<propName>` instead? I'm curious if @dbouwman feels the same.

dbouwman: Since `IHubSearchResult` has `[key: string]: any;` it will allow any props to be added. By I'd agree with @sonofflynn89 that we'd be better off copying `ogcItem.properties` to the returned object as `.rawResult`. So we've have `result.rawResult.postType` === `ogcItem.properties.postType`

dbouwman: @drspacemanphd in hindsight, there may have been a better pattern to utilize, but the hubSearch + gallery have evolved over time, and this is where we find ourselves. This is the short version of the history...

The initial design was to have `hubSearch` return objects optimized for display in the client, and abstracted from their implementation - thus `IHubSearchResult`. Then there was a request to allow the system to return arbitrary additional metadata - you've heard this from Andrew as "percent cloud cover" etc - so we added `[key: string]: any;` to the search result to allow for that. Later on, we got to a place where different galleries needed to exert more specific control over what's displayed in the cards in the gallery (implemented as a callback fn that can be passed to the gallery), and to do this, we needed to expose more of the underlying source object - thus `.rawResult`.

packages/common/src/search/_internal/hubSearchItemsHelpers/ogcItemToDiscussionPostResult.ts
```diff
@@ -23,11 +23,11 @@ export async function ogcItemToDiscussionPostResult(
     updatedDateSource: "properties.updatedAt",
     type: ogcItem.properties.postType,
```
sonofflynn89, line 24: Let's change this to be a constant `type: 'post'` so functions down the line can use the field as a discriminator. Then functions that specialize in posts can access `rawResult.properties.postType` as needed`

drspacemanphd: As you wish!

sonofflynn89 (approved):
Thank you for your work, Mark! 🙏

PR #1461, opened 2024-04-05
fix: metric/stat card schema sourceLink does url validation

benstoltz:
1. Description: Adds url validation to stat card schema for sourceLink

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/internal/metrics/MetricSchema.ts
+++ b/packages/common/src/core/schemas/internal/metrics/MetricSchema.ts
@@ -66,6 +66,10 @@ export const MetricSchema: IConfigurationSchema = {
     },
     sourceLink: {
       type: "string",
+      if: { minLength: 1 },
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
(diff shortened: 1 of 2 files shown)


packages/common/src/core/schemas/internal/metrics/MetricSchema.ts
```diff
@@ -66,6 +66,10 @@ export const MetricSchema: IConfigurationSchema = {
     },
     sourceLink: {
       type: "string",
+      if: { minLength: 1 },
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
juliannemarik, line 71: I know you're just following the existing pattern, but I don't love that we'll need to include this `ts-ignore` every time we want to use the hub-defined `url` format. I'm wondering if we can somehow extend our `IConfigurationSchema` (which extends the `JSONSchema`) to allow this format? 

Anyways, that's outside the scope of this issue, but may be something we want to consider.

juliannemarik (approved):

PR #1467, opened 2024-04-10
feat(hub-common): add facade methods related to entity sharing and ac…

rweber-esri:
…cess in support of extending co

affects: @esri/hub-common

ISSUES CLOSED: 8473

1. Description:

Adds facade methods related to sharing and access to extend collaborators workspace support for events 3

1. Instructions for testing:

1. Closes Issues: [#8473](https://devtopia.esri.com/dc/hub/issues/8473)


rweber-esri:
This is ready for review @juliannaeapicella & @dbouwman. I've tested these changes extensively against my [opendata-ui PR-13385](https://github.com/ArcGIS/opendata-ui/pull/13385) (`?uiVersion=PR-13385`). Specifically:
* changing access works as expected across all entities (tested Groups, Initiatives, Projects & Events workspaces)
* managing groups works as expected across all entity workspaces that have collaborators panes (tested Initiatives, Projects & Events)

packages/common/src/events/_internal/shareEventWithGroups.ts
```diff
@@ -0,0 +1,77 @@
...
+import { poll } from "../../utils/poll";
+
+/**
+ * Shares an event with one or more groups
+ * @param groupIds An array of group IDs to share the event with
+ * @param entity An IHubEvent object
+ * @param context An IArcGISContext object
+ * @returns a promise that resolves IHubItemEntity
+ */
+export async function shareEventWithGroups(
```
dbouwman, line 15: consider adding more comments about what's going on in this - I'm not following why we're polling a search for groups etc, so my guess is that others on the team, including "Future Randy" will also be a little O_o when they need to get into this code :)

rweber-esri: Will do. For context, _new_ groups aren't _immediately_ available in the AGO search index. This prevents the need for us to add polling potentially in many places of the app where we create new groups and then want to immediately share an item to that group. In 99% of cases, this should only fire a single request. The collaborators pane had implemented this polling before sharing the entity with 1 or more groups, but had a bug where `num` wasn't specified on the polling request. This was problematic in the scenario where the user selected 11+ groups to share the entity with. The validation function expected 11 results, but AGO search would only return 10 (the default num of results when `num` param is omitted).

packages/common/src/utils/poll.ts
```diff
@@ -11,14 +11,16 @@
  * @param validationFn
  * @param opts
  */
```
dbouwman, line 13: ❤️❤️❤️❤️ Only as is that the polling also implement an incremental back-off - so if it's not valid after 300ms, the next timeout is 600ms. This is what AGO asked us to do if we use a polling strategy b/c the api is not deterministic

rweber-esri: I thought about this too. Will push up a quick change to add incremental backoff

dbouwman (approved):
This looks great! Thanks for adding the polling - only ask is to add "incremental back-off" to `poll` and add a little more description of what's happening in the share/unshare functions

PR #1471, opened 2024-04-11
feat: scheduler added to content uischema

abp6318:
1. Description:

- [x] move types and interfaces to hub.js
- [x] add section to uiSchema for scheduler
- [x] fetch schema for content entity OR assign it default value
- [x] render scheduler w/ default value
- [x] update schedule value if user hits save button

Screenshot 2024-04-16 at 5 59 12 PM

1. Instructions for testing:
- pull this PR and the od-ui PR -- https://github.com/ArcGIS/opendata-ui/pull/13422
- copy your build of hub.js into od-ui
- run your ember server (this will not work in the harness given the permission / feature flagging)
- go to any dev site and view the scheduler in the settings pane
- OR go to any prod or qa site and view it by attaching the permission `pe=hub:content:workspace:settings:schedule`
  - here is the one I've been testing on: https://download-test-qa-pre-a-hub.hubqa.arcgis.com:4200/workspace/content/88b077a232224e52b684eacd2ecdda5f/settings?pe=hub:content:workspace:settings:schedule

also, thank you Dave for helping get the wonky permissions going!

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/2068

### DIFF
--- a/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
+++ b/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
@@ -17,8 +17,59 @@ export const buildUiSchema = async (
 ): Promise<IUiSchema> => {
   const uiSchema: IUiSchema = {
     type: "Layout",
-    elements: [],
+    label: "Scheduling",
+    elements: [
+      // TODO: wrap this element in a feature flag / permission
+      {
+        type: "Section",
+        label: "Scheduling", // TODO: i18n
+        elements: [
+          {
+            type: "Control",
+            scope: "/properties/schedule",
+            label:
+              "Hub will automatically try to determine the best time to generate a download file for this item. However, you can choose to manually set when you want the download file to be updated.", // TODO: i18n
+            options: {
+              type: "Control",
+              control: "hub-field-input-scheduler",
+              format: "radio",
+              inputs: [
+                { label: "Default", type: "automatic" }, // TODO: i18n
+                {
+                  label: "Daily",
+                  type: "daily",
+                  expandedHelperText: "Each day at",
+                }, // TODO: i18n
+                {
+                  label: "Weekly",
+                  type: "weekly",
+                  expandedHelperText: "Each week on",
+                }, // TODO: i18n
+                {
+                  label: "Monthly",
+                  type: "monthly",
+                  expandedHelperText: "Each month on",
+                }, // TODO: i18n
+                {
+                  label: "Yearly",
+                  type: "annually",
+                  expandedHelperText: "Each year on",
+                }, // TODO: i18n
+                {
+                  label: "Manual",
+                  type: "manual",
+                  helperActionIcon: "information-f",
+                  helperActionText:
+                    "Use this option to manually update the search index and cached download files for this item.",
+                }, // TODO: i18n
+              ],
+            },
+          },
+        ],
+      },
+    ],
   };
+
   if (isHostedFeatureServiceEntity(options as IHubEditableContent)) {
     uiSchema.elements.push({
       type: "Section",
--- a/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
+++ b/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
@@ -56,9 +56,9 @@ export const buildUiSchema = async (
                 expandedHelperText: `option.monthly.expandedHelperText`,
               },
               {
-                label: `option.annually.label`,
-                type: "annually",
-                expandedHelperText: `option.annually.expandedHelperText`,
+                label: `option.yearly.label`,
+                type: "yearly",
+                expandedHelperText: `option.yearly.expandedHelperText`,
               },
               // uncomment this when the manual option is available
               // {
--- a/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
+++ b/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
@@ -29,8 +29,10 @@ export const buildUiSchema = async (
             scope: "/properties/schedule",
             labelKey: `${i18nScope}.sections.schedule.helperText`,
             options: {
+              // The scope for the options is the scope of the scheduler field, not the config editor
               type: "Control",
               control: "hub-field-input-scheduler",
+              labelKey: "fieldHeader",
               format: "radio",
               inputs: [
                 { label: `option.default.label`, type: "automatic" },
--- a/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
+++ b/packages/common/src/content/_internal/ContentUiSchemaSettings.ts
@@ -22,47 +22,45 @@ export const buildUiSchema = async (
       // TODO: wrap this element in a feature flag / permission
       {
         type: "Section",
-        label: "Scheduling", // TODO: i18n
+        labelKey: `${i18nScope}.sections.schedule.label`,
         elements: [
           {
             type: "Control",
             scope: "/properties/schedule",
-            label:
-              "Hub will automatically try to determine the best time to generate a download file for this item. However, you can choose to manually set when you want the download file to be updated.", // TODO: i18n
+            labelKey: `${i18nScope}.sections.schedule.helperText`,
             options: {
               type: "Control",
               control: "hub-field-input-scheduler",
               format: "radio",
               inputs: [
-                { label: "Default", type: "automatic" }, // TODO: i18n
+                { label: `option.default.label`, type: "automatic" },
                 {
-                  label: "Daily",
+                  label: `option.daily.label`,
                   type: "daily",
-                  expandedHelperText: "Each day at",
-                }, // TODO: i18n
+                  expandedHelperText: `option.daily.expandedHelperText`,
+                },
                 {
-                  label: "Weekly",
+                  label: `option.weekly.label`,
                   type: "weekly",
-                  expandedHelperText: "Each week on",
-                }, // TODO: i18n
+                  expandedHelperText: `option.weekly.expandedHelperText`,
+                },
                 {
-                  label: "Monthly",
+                  label: `option.monthly.label`,
                   type: "monthly",
-                  expandedHelperText: "Each month on",
-                }, // TODO: i18n
+                  expandedHelperText: `option.monthly.expandedHelperText`,
+                },
                 {
-                  label: "Yearly",
+                  label: `option.annually.label`,
                   type: "annually",
-                  expandedHelperText: "Each year on",
-                }, // TODO: i18n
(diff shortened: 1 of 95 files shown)


packages/common/src/content/edit.ts
```diff
@@ -168,6 +169,12 @@ export async function updateContent(
     }
   }
 
+  if (content.schedule.mode === "automatic") {
+    await deleteSchedule(item, requestOptions);
+  } else if (content.schedule !== (await getSchedule(item, requestOptions))) {
+    await setSchedule(item, content.schedule, requestOptions);
+  }
```
sonofflynn89, line 176: Couple things to consider:
- What does `deleteSchedule()` do if no schedule is actually set? If it throws a 4xx error we'll need to account for that
- `content.schedule` is an object, so `!==` probably won't work the way you're expecting. You'll either need to create a deep compare function _or_ just always save the schedule.

packages/common/src/content/manageSchedule.ts
```diff
@@ -0,0 +1,95 @@
...
+  const url = `${getHubApiUrl(requestOptions)}/api/download/v1/items/${
+    item.id
+  }/schedule`;
+  const options = {
+    method: "DELETE",
+    headers: {
+      accept: "application/json",
+    },
+  };
+  await fetch(url, options);
```
sonofflynn89, line 94: Same here by just awaiting fetch, the consumer has no idea whether the delete succeeded or not.

sonofflynn89 (commented):
Great first past! Can't wait to get this into the app 🙌 I've left a few comments for you, but I'm happy to jump on a call if you'd prefer to talk synchronously.

packages/common/src/content/_internal/ContentUiSchemaSettings.ts
```diff
@@ -19,6 +20,60 @@ export const buildUiSchema = async (
...
+    uiSchema.elements.push({
+      type: "Section",
+      labelKey: `${i18nScope}.sections.schedule.label`,
+      elements: [
+        {
+          type: "Control",
+          scope: "/properties/schedule",
+          labelKey: `${i18nScope}.sections.schedule.helperText`,
+          options: {
+            // The scope for the options is the scope of the scheduler field, not the config editor
```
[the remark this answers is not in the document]
tomwayson: @sonofflynn89 is that the pattern we do for other sections and schemas?

packages/common/src/content/edit.ts
```diff
@@ -168,6 +169,12 @@ export async function updateContent(
     }
   }
 
+  if (content.schedule.mode === "automatic") {
+    await deleteSchedule(item, requestOptions);
```
tomwayson, line 173: should we first fetch the schedule, and then only delete the schedule when one exists and the mode is automatic?

packages/common/src/content/fetch.ts
```diff
@@ -264,13 +267,17 @@ export const fetchHubContent = async (
...
+  const enrichments: IHubEditableContentEnrichments = {};
   if (isHostedFeatureServiceItem(item)) {
     enrichments.server = await getService({
       ...requestOptions,
       url: parseServiceUrl(item.url),
     });
   }
+
+  // fetch schedule and add it to enrichments if it exists in schedule API
+  enrichments.schedule = await getSchedule(item, requestOptions);
```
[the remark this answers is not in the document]
tomwayson: I wonder if we should include the enterprise check into `getSchedule()`? othewise we'll need to repeat it in a few other places.

tomwayson (commented):
Mostly "what he said" to Caleb's existing comments.

packages/common/src/content/_internal/ContentUiSchemaSettings.ts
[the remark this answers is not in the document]
sonofflynn89: @tomwayson it's a good question, I don't know. I was under the impression that all other configuration, editor fields depend on external (configuration editor) translation strings, not hard coded strings within the field itself. @abp6318 do you know of any other configuration editor fields whose options use static translation paths like you have here?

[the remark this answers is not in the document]
abp6318: @tomwayson @sonofflynn89 

It seems weird, but I took this approach considering the strings will most likely get reused, and rather than repeating strings in the already cluttered JSON file for the config editor, you can store fewer, shared strings in the input locales itself.

This was the original plan back in November and December, and there was even talk of a string audit by design/platform, but I never saw anything come out of it. 

If either of you think I should change this, I will, but to avoid having to change half a dozen strings in one file when it can be done with only one, I'd like to keep it this way.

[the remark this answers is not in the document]
abp6318: Discussed a better way to do this with Julianne and Jordan and settled on using default labels managed in opendata-ui, OR you can pass in a label+expandedHelperText from another scope!

packages/common/test/content/manageSchedule.test.ts
```diff
@@ -0,0 +1,220 @@
...
+        statusCode: 404,
+      }
+    );
+
+    const response = await maybeUpdateSchedule(content, MOCK_HUB_REQOPTS);
+    expect(response).toEqual(false);
+    expect(fetchMock.calls().length).toBe(1);
+  });
+
+  it("maybeUpdateSchedule: no schedule is set, and updatinng to scheduled is needed", async () => {
```
sonofflynn89, line 134: small typo `updatinng` -> `updating`

packages/common/src/content/manageSchedule.ts
```diff
@@ -44,52 +40,102 @@ export const getSchedule = async (
...
+ */
+export const maybeUpdateSchedule = async (
+  content: IHubEditableContent,
+  requestOptions: IRequestOptions
+) => {
+  const currentSchedule = await getSchedule(content.id, requestOptions);
+
+  // if no schedule is set and incoming schedule is automatic, do nothing
+  if (content.schedule.mode === "automatic" && currentSchedule === null) {
+    return false;
```
sonofflynn89, line 103: Whenever we are working in typescript, we want to make sure that we are very deliberate with our return types. I notice that we don't have an explicit return type on this function and the `deleteSchedule`, `setSchedule`, etc. have a return type of `any`.

Try and come up with a consistent pattern (e.g., always returning a boolean, returning an `IMaybeUpdateScheduleResponse`, etc.) and let's talk when you do.

sonofflynn89 (commented):
Great work so far! Just a few things we'll want to tighten up before we merge 👌

packages/common/src/content/manageSchedule.ts
```diff
@@ -44,52 +40,102 @@ export const getSchedule = async (
...
+ *
+ * Note: After a discussion w/ Dave, we decided this check is preferred to passing the context
+ * into functions that would need checkPermission
+ * @param requestOptions The request options needed to get the HubApiUrl
+ * @returns Whether or not the scheduling feature is available
+ */
+export const isDownloadSchedulingAvailable = (
+  requestOptions: IRequestOptions
+): boolean => {
+  return requestOptions.portal?.includes("arcgis.com");
```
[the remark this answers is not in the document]
abp6318: Yup, although if you think it should be different, I can message him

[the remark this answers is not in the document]
sonofflynn89: All good! Thanks for responding 🙏

tomwayson (approved):
Great work @abp6318

packages/common/src/content/fetch.ts
```diff
@@ -264,13 +266,21 @@ export const fetchHubContent = async (
...
-  const enrichments: IItemAndIServerEnrichments = {};
+  const enrichments: IHubEditableContentEnrichments = {};
   if (isHostedFeatureServiceItem(item)) {
     enrichments.server = await getService({
       ...requestOptions,
       url: parseServiceUrl(item.url),
     });
   }
+
+  if (isDownloadSchedulingAvailable(requestOptions)) {
```
sonofflynn89, line 277: Shoot, last big thing. Aren't schedules only available for _public_ items? Double check with San, but if so we should probably bake that check into `isDownloadSchedulingAvailable`. 

On that note, I recommend we move the `|| { mode: 'automatic' }` fallback within this if statement so private items (and enterprise items) will have an undefined `schedule`.

packages/common/src/content/manageSchedule.ts
```diff
@@ -0,0 +1,155 @@
...
+      "Content-Type": "application/json",
+      accept: "application/json",
+    },
+    body: JSON.stringify({
+      ...body,
+      itemId,
+    }),
+  };
+  const response = await fetch(url, options);
+  return (await response.json()) as IHubScheduleResponse;
```
sonofflynn89, line 75: Won't the API return `itemId` in the response? And won't we need to add `mode` like you do in `getSchedule`?

packages/common/src/content/manageSchedule.ts
```diff
@@ -0,0 +1,155 @@
...
+): Promise<IHubScheduleResponse> => {
+  const url = getSchedulerApiUrl(itemId, requestOptions);
+  const options = {
+    method: "DELETE",
+    headers: {
+      accept: "application/json",
+    },
+  };
+  const response = await fetch(url, options);
+  return (await response.json()) as IHubScheduleResponse;
```
sonofflynn89, line 95: Same here, won't the API return itemId in the response? And won't we need to add mode like you do in getSchedule?

packages/common/test/content/manageSchedule.test.ts
```diff
@@ -0,0 +1,235 @@
...
+import { MOCK_HUB_REQOPTS } from "../mocks/mock-auth";
+import { IHubEditableContent } from "../../src/core/types/IHubEditableContent";
+import * as fetchMock from "fetch-mock";
+
+describe("manageSchedule", () => {
+  afterEach(() => {
+    fetchMock.restore();
+  });
+
+  it("getSchedule: returns null if no schedule is set", async () => {
```
sonofflynn89, line 20: We should probably update the test names since we no longer return null. Same goes for the other tests in this file.

sonofflynn89 (commented):
Heroic work man! I do have one bigger question I'd like for us to answer before we merge, all the other comments are minor.

packages/common/src/content/manageSchedule.ts
sonofflynn89: Yes, I would do just that. Remember, these functions are supposed to return an `IHubScheduleResponse`. Our duty as developers is to make sure that callers of this function get a reponse object _always matches the return type_. Right now we're breaking the contract.

sonofflynn89 (approved):
Good job seeing this through!

PR #1482, opened 2024-04-23
feat(hub-common): add new utils for a unified download flow

sonofflynn89:
1. Description:

Part of https://devtopia.esri.com/dc/hub/issues/9627
Introduces new functions to enable a unified interface for downloading via:
- createReplica (Hub API)
- paging jobs (Hub API)
- item export
- image export

Note that the logic for item / image export has not been fully fleshed out at this time, but it will be in a future PR.


packages/common/src/content/fetch.ts
```diff
@@ -249,13 +249,13 @@ export const fetchHubContent = async (
...
 ): Promise<IHubEditableContent> => {
   // NOTE: b/c we have to support slugs we use fetchContent() to get the item
-  // by telling it to not fetch any enrichments
-  // which we then fetch as needed after we have the item
+  // by telling it to not fetch any enrichments which we then fetch as needed after we have the item
+  // EXCEPTION: We ask it to fetch the metadata enrichment so we don't have to parse xml on the client
   const options = {
     ...requestOptions,
-    enrichments: [],
+    enrichments: ["metadata"],
```
sonofflynn89, line 256: I have a few questions here:
1) Do we only want formal metadata for public items?
2) Do we want to fetch formal metadata from our search index or from AGO directly (and have to parse XML ourselves)?
3) Should we rely on the metadata fetching logic of `fetchContent` or should we fetch it in `fetchHubContent`?

packages/common/src/core/types/IItemEnrichments.ts
```diff
@@ -30,4 +31,9 @@ export interface IItemEnrichments {
    * see https://github.com/ArcGIS/hub-indexer/blob/master/docs/errors.md#response-formatting-for-errors
    */
   errors?: IEnrichmentErrorInfo[];
+
+  /**
+   * Additional resources specified in the formal item metadata
+   */
+  additionalResources?: IHubAdditionalResource[];
```
sonofflynn89, line 38: `IItemEnrichments` already has a `metadata` member, should we really add an `additionalResources` member too? I feel like it would through off types for `fetchContent`, but adding it here was the most expedient thing during development.

sonofflynn89: Since Aaron created `IHubEditableContentEnrichments`, this was no longer a problem.

tannerjt (approved):
LGTM pending jsdoc and test coverage.  Very nice work @sonofflynn89!

packages/common/src/content/fetch.ts
tomwayson: > Do we want to fetch formal metadata from our search index or from AGO directly (and have to parse XML ourselves)?

I think we should get it from AGO directly, and parse it on the client. That's what we do in enterprise already, and  that means we don't have to worry about only fetching metadata if not on enterprise or for public items.

> Should we rely on the metadata fetching logic of fetchContent or should we fetch it in fetchHubContent?

We should fetch it in `fetchHubContent` using the `enrichments` system below.

packages/common/src/content/fetch.ts
```diff
@@ -251,13 +251,16 @@ export const fetchHubContent = async (
...
   } as IFetchContentOptions;
-  const { item, access } = await fetchContent(identifier, options);
+  const { item, access, additionalResources } = await fetchContent(
```
tomwayson, line 260: so `additionalResources` is only available from the v3 API? If not, let's fetch it like any other enrichment below.

packages/common/src/core/types/IHubEditableContent.ts
```diff
@@ -19,6 +20,16 @@ export interface IHubEditableContent
    * capability enabled. This is a pre-requisite for Hosted Downloads to work.
    */
   serverExtractCapability?: boolean;
+  /**
+   * If the item represents a hosted feature service with "Extract enabled", shows the formats that
+   * can be extracted from the service via the "createReplica" operation.
+   */
+  serverExtractFormats?: string[];
+  /**
+   * links to additional resources specified in the formal item metadata
```
tomwayson, line 29: if this comes from formal item metadata, but needs to be exposed at the top level, then let's just make a getter on the class for `this.metadata?.path.to.additionalResources`

packages/common/src/downloads/index.ts
```diff
@@ -1 +1,4 @@
-export * from "./build-existing-exports-portal-query";
```
tomwayson, line 1: Looks like you took the old module and split it up into new modules. That's fine as long as everything that was previously exported from the old module is still exported from the new ones (so that there are no breaking changes). I assume you were conscious of that, but I just want to confirm.

packages/common/src/downloads/helpers/canCreateExportItem.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param _entity
+ * @param _context
+ * @returns
+ */
+export function canCreateExportItem(
+  _entity: IHubEditableContent,
+  _context: IArcGISContext
+) {
+  // TODO: port over logic from the download-service
```
tomwayson, line 18: confirming that you didn't intend to do this in this PR

packages/common/src/downloads/helpers/canUseHubDownloadApi.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Determines if the Hub Download API can be used for the given entity.
+ * @param entity entity to check if Hub Download API can be used
+ * @param context ArcGIS context
+ * @returns whether the Hub Download API can be used
+ */
+export function canUseHubDownloadApi(
+  entity: IHubEditableContent,
+  context: IArcGISContext
+): boolean {
+  // TODO: use permission instead.
```
tomwayson, line 17: ditto

packages/common/src/downloads/helpers/file-url-fetchers/fetchExportImageDownloadFileUrl.ts
```diff
@@ -0,0 +1,59 @@
+import { request } from "@esri/arcgis-rest-request";
+import {
+  DownloadOperationStatus,
+  IFetchDownloadFileUrlOptions,
+} from "../types";
+
+/**
+ * @private
```
tomwayson, line 8: please consider moving any net new functions that we do not intend to expose externally to an `_internal` folder (or module) so that they do not get accidentally re-exported later.

packages/common/src/downloads/helpers/file-url-fetchers/fetchExportItemDownloadFileUrl.ts
```diff
@@ -0,0 +1,109 @@
...
+  const result: IExportParameters = {};
+
+  if (layers) {
+    result.layers = layers.map((id) => ({ id, where }));
+  }
+
+  return result;
+}
+
+async function pollForJobCompletion(
```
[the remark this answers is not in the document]
tomwayson: seems to me like this is a good use case for that. I feel like download times increase exponentially w/ the size of the data.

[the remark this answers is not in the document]
sonofflynn89: Good callout! I'd be curious to hear what product thinks. Since we have to display a progress indicator with a percentage, I'd wager we want a consistent timeout. I suspect that the exponential timeouts would cause users to think that the download operation has frozen (particularly for larger downloads)

packages/common/src/downloads/fetchDownloadFileUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+export async function fetchDownloadFileUrl(
+  options: IFetchDownloadFileUrlOptions
+): Promise<string> {
+  let fetchingFn;
+  if (canUseHubDownloadApi(options.entity, options.context)) {
+    fetchingFn = fetchHubApiDownloadFileUrl;
+  } else if (canUseExportItemFlow(options.entity)) {
+    fetchingFn = fetchExportItemDownloadFileUrl;
+  } else if (canUseExportImageFlow(options.entity)) {
+    fetchingFn = fetchExportImageDownloadFileUrl;
```
tomwayson, line 24: consider using dynamic `import()` for these, for example:

`fetchingFn = (await import('./helpers/file-url-fetchers/fetchHubApiDownloadFileUrl')).fetchHubApiDownloadFileUrl`

packages/common/src/downloads/fetchDownloadFormats.ts
```diff
@@ -0,0 +1,50 @@
...
+): Promise<IDownloadFormat[]> {
+  const { entity, context, layers } = options;
+  // fetch base formats for the item
+  let baseFormats: IDownloadFormat[] = [];
+  if (canUseHubDownloadApi(entity, context)) {
+    baseFormats = getHubDownloadApiFormats(entity);
+  } else if (canUseExportItemFlow(entity)) {
+    baseFormats = await fetchExportItemFormats(entity, context, layers);
+  } else if (canUseExportImageFlow(entity)) {
+    baseFormats = getExportImageDownloadFormats();
```
tomwayson, line 30: consider using dynamic `import()` for these as well

packages/common/src/downloads/helpers/format-fetchers/fetchExportItemFormats.ts
```diff
@@ -0,0 +1,43 @@
...
+  //   : fetchAvailableExportItemFormats(entity, context, layers);
+}
+
+// function canCreateExport(
+//   _entity: IHubEditableContent,
+//   _context: IArcGISContext
+// ): boolean {
+//   // TODO: port over logic from ember's download-service
+//   return true;
+// }
```
tomwayson, line 43: delete commented code?

tomwayson (commented):
Heroic work @sonofflynn89 

I didn't dig into the actual business logic too much - I trust you on that.

packages/common/src/content/fetch.ts
sonofflynn89: If we are good parsing the XML on the client, then we can absolutely fetch it below and not rely on the v3 api. We already have util functions to convert metadata JSON into the additional resources.

packages/common/src/core/types/IHubEditableContent.ts
sonofflynn89: Good idea! Unfortunately, there is some conversion that needs to be made so we can't use a single getter. See this compose helper [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/content/_internal/internalContentUtils.ts#L466).

packages/common/src/downloads/index.ts
sonofflynn89: Absolutely, you can see the updated exports associated with this on line 1 for the function, and line 2 for the `PORTAL_EXPORT_TYPES` const

packages/common/src/downloads/helpers/canCreateExportItem.ts
sonofflynn89: That is correct. I thought about deleting all the code for the export item flow (including this helper), but since I had gone so far down the path, I figured it would be better to check the work in progress stuff in and flesh it out in a follow on PR.

packages/common/src/downloads/helpers/canUseHubDownloadApi.ts
sonofflynn89: I'm actually curious what the best path is to take here. I don't think a permission could do such complex validation on the entity types, but I guess we could create a permission that reports whether the download API is available in general. I'll take this on in a follow-on PR.

packages/common/src/downloads/helpers/file-url-fetchers/fetchExportImageDownloadFileUrl.ts
sonofflynn89: Great idea, I'll do that now

packages/common/src/core/types/IHubEditableContent.ts
```diff
@@ -19,6 +20,16 @@ export interface IHubEditableContent
    * capability enabled. This is a pre-requisite for Hosted Downloads to work.
    */
   serverExtractCapability?: boolean;
+  /**
+   * If the item represents a hosted feature service with "Extract enabled", shows the formats that
+   * can be extracted from the service via the "createReplica" operation.
+   */
+  serverExtractFormats?: string[];
```
tannerjt, line 27: Does it make any sense to specify that the type is replica format array (eg `createReplicaFormat[]`)?

packages/common/src/downloads/_internal/_types.ts
```diff
@@ -0,0 +1,63 @@
...
+  ServiceDownloadFormat.GIF,
+  ServiceDownloadFormat.JPG,
+  ServiceDownloadFormat.JPG_PNG,
+  ServiceDownloadFormat.LERC,
+  ServiceDownloadFormat.PNG,
+  ServiceDownloadFormat.PNG24,
+  ServiceDownloadFormat.PNG32,
+  ServiceDownloadFormat.PNG8,
+  ServiceDownloadFormat.TIFF,
+] as const;
```
tannerjt, line 34: Cool, I didn't realize `as const` makes this work like a read-only tuple.

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFileUrl.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @param options options for refining / filtering the resulting download file
+ * @returns a download api url that can be polled
+ */
+function getDownloadApiRequestUrl(options: IFetchDownloadFileUrlOptions) {
+  const { entity, format, context, layers, geometry, where } = options;
+
+  const searchParams = new URLSearchParams({
+    redirect: "false", // Needed to get the download URL instead of the file itself
+    layers: layers[0].toString(),
```
tannerjt, line 67: The interface for `IFetchDownloadFileUrlOptions` shows `layers` as being optional, so it's possible `layers[0].toString()` may throw an error.

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFileUrl.ts
```diff
@@ -0,0 +1,222 @@
...
+  });
+
+  if (geometry) {
+    const geometryJSON = geometry.toJSON();
+    // Not sure why type isn't included in the toJSON() output, but our API expects it
```
tannerjt, line 72: agreed, i don't know why type is excluded here, but a common problem

tannerjt (approved):
Re-approved.  Epic 🚀

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFileUrl.ts
tannerjt: Perfect, makes sense, thanks!

packages/common/src/core/types/IHubEditableContent.ts
sonofflynn89: @tannerjt It's a good question, and I'm torn. The unfortunate reality is that `CreateReplicaFormat` is an incomplete list, it's just the ones we know about. My thought was that we would leave it as `string[]` here to indicate our uncertainty of what will actually be returned back.

The next question may be "why have `CreateReplicaFormat` at all if we're not sure they are the right values?" Even though we are uncertain, I think it is worth having a list within the downloads sub-package that developers deep in the bowels of downloads can use as a sanity check. But what do you think?

tannerjt: It seems perfectly okay to leave it as-is for now.

tomwayson (approved):
Great work @sonofflynn89!

PR #1483, opened 2024-04-23
feat(): remove default groups

jordantsanz:
1. Description:

1. Instructions for testing:


packages/common/src/projects/HubProject.ts
```diff
@@ -198,32 +198,8 @@ export class HubProject
         )) as IHubProjectEditor)
       : (cloneObject(this.entity) as IHubProjectEditor);
 
-    // 2. on project creation, pre-populate the sharing field
-    // with the core + collaobration groups if the user has the
-    // appropriate shareToGroup portal privilege
+    // 2. editor._groups handling
```
juliannemarik, line 201: yay! I'm so glad we can remove this. Couple questions:
1. do we still need to set `editor._groups` to an empty array or can we just remove this step entirely? 
2. I wonder if we should remove these options from the `IEntityEditorContext` [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/types/HubEntityEditor.ts#L12-L13)? 
3. do we also want to remove this from the initiative creation workflow? I don't think I see that in the AC, but I imagine we would? If the answer is no, then disregard number 2, as we'd still need those properties.

PR #1484, opened 2024-04-24
fix: update fetchHubGroup ro and follow card permission

vivzhang:
1. Description:

- change the ro parameter that `fetchHubGroup` takes from `userRequestOptions` to `hubRequestOptions`, so anonymous users can fetch public groups from public sites
- add "hub-basic" license to follow card permission so hub basic users have access to follow cards

1. Instructions for testing:

### DIFF
--- a/packages/common/src/groups/HubGroups.ts
+++ b/packages/common/src/groups/HubGroups.ts
@@ -142,7 +142,7 @@ export async function createHubGroup(
  */
 export async function fetchHubGroup(
   identifier: string,
-  requestOptions: IUserRequestOptions
+  requestOptions: IHubRequestOptions
 ): Promise<IHubGroup> {
   const group = await getGroup(identifier, requestOptions);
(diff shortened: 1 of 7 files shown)


packages/common/src/groups/HubGroups.ts
```diff
@@ -142,7 +142,7 @@ export async function createHubGroup(
  */
 export async function fetchHubGroup(
   identifier: string,
-  requestOptions: IUserRequestOptions
+  requestOptions: IHubRequestOptions
```
juliannemarik, line 145: is this technically a breaking change? wondering if we need to release it as such?

juliannemarik (approved):

packages/common/src/groups/HubGroups.ts
vivzhang: I did check with Dave and we decided this was something small and not widely used and would have been caught by Typescript if the type was incorrect.

PR #1489, opened 2024-04-29
fix: export IWithFollowers

vivzhang:
affects: @esri/hub-common

1. Description:

Export `IWithFollowers` so we can assign it to certain hub entities as only entities extend `IWithFollowers` have access to the `followersGroupId`

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/traits/index.ts
+++ b/packages/common/src/core/traits/index.ts
@@ -7,3 +7,4 @@ export * from "./IWithCatalog";
 export * from "./IWithViewSettings";
 export * from "./IWithDiscussions";
 export * from "./ICapabilityConfig";
+export * from "./IWithFollowers";


juliannemarik (approved):
Just curious why this is needed? 
Currently we extend `IWithFollowers` in the `IHubItemEntity` (see [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/types/IHubItemEntity.ts#L21)) meaning all item-based entities have access to a `followersGroupId`. Do we want to change that as part of this work as well?

vivzhang:
> Just curious why this is needed? Currently we extend `IWithFollowers` in the `IHubItemEntity` (see [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/types/IHubItemEntity.ts#L21)) meaning all item-based entities have access to a `followersGroupId`. Do we want to change that as part of this work as well?

We used to get the entity's followers group id by calling the `getEntityFollowersGroupId` fn in `arcgis-hub-follow-card`, which we used the [same approach](https://github.com/Esri/hub.js/blob/44bb118f3a5e4372712be88429862936b041a4d5/packages/common/src/items/follow.ts#L23-L27), because `HubEntity` also includes `IHubGroup` which doesn't extend `IHubItemEntity`, so if we need to access an item's `IWithFollowers` prop we will need to assign `IWithFollowers` as the type otherwise we will get a TypeScript error. Exporting it now because we are switching to fetching the entity and then getting the `followersGroupId` in `arcgis-hub-follow-card` instead of calling `getEntityFollowersGroupId`, because we want to log the entity's `access` telemetry per product's request. I just pushed a [new commit](https://github.com/ArcGIS/opendata-ui/pull/13376/commits/4b4758543d305550684864e29088cbcacc8c853f) in od-ui for this change.

PR #1492, opened 2024-04-30
fix: updates well known types to fix gallery query

MarvinPerry:
affects: @esri/hub-common

1. Description:
    - updates hub Initiative query to include  `typekeyword: hubInitiativeV2` 
1. Closes Issues: #[10065](https://devtopia.esri.com/dc/hub/issues/10065) (if appropriate)

### DIFF
--- a/packages/common/src/search/_internal/portalSearchItems.ts
+++ b/packages/common/src/search/_internal/portalSearchItems.ts
@@ -347,6 +347,7 @@ export const WellKnownItemPredicates: IWellKnownItemPredicates = {
   $initiative: [
     {
       type: "Hub Initiative",
+      typekeywords: "hubInitiativeV2",
     },
   ],
   $experience: [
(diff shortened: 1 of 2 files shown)


packages/common/src/search/_internal/portalSearchItems.ts
```diff
@@ -347,6 +347,7 @@ export const WellKnownItemPredicates: IWellKnownItemPredicates = {
   $initiative: [
     {
       type: "Hub Initiative",
+      typekeywords: "hubInitiativeV2",
```
vivzhang, line 350: Will this affect the current initiative?

MarvinPerry: @vivzhang Jay and I discussed that in our 1:1 yesterday, and he didn't think it was an issue. With that said, that doesn't mean there isn't some case we missed. I could add a new entry, if we think that's better.

vivzhang: I see, that's fine with me, especially since we are close to the release!

vivzhang (approved):

PR #1494, opened 2024-05-01
feat: scheduler force update

abp6318:
1. Description:
- adds manual option to scheduler on content settings
- adds tile select below scheduler on content settings

Waiting for San's PR to merge before you can actually save the manual option -- https://github.com/ArcGIS/hub-download-api/pull/34

1. Instructions for testing:
- open network tab on https://localui.arcgis.com:5173/workspace/content/88b077a232224e52b684eacd2ecdda5f/settings?pe=hub:content:workspace:settings:schedule
- test saving as manual option
- test that reharvest is triggered when save button is pressed ONLY when tile select is checked/true

1. Closes Issues: [2175]( https://devtopia.esri.com/dc/hub/issues/2175)


packages/common/src/content/_internal/ContentSchema.ts
```diff
@@ -23,6 +23,15 @@ export const ContentSchema: IConfigurationSchema = {
     },
     schedule: {
       type: "object",
+      properties: {
+        _forceUpdate: {
+          type: "array",
+          items: {
+            type: "boolean",
+            enum: [true],
```
sonofflynn89, line 31: Do we want to restrict this to only allow true? And how come we're using an array instead of a const?

packages/common/src/content/_internal/ContentUiSchemaSettings.ts
```diff
@@ -45,13 +45,26 @@ export const buildUiSchema = async (
...
+                helperActionIcon: "information-f",
+                helperActionText: `{{${i18nScope}.fields.schedule.manual.helperActionText:translate}}`,
+              },
+            ],
+          },
+        },
+        // force update checkbox -- TODO: replace with button once available
+        {
+          type: "Control",
+          scope: "/properties/schedule/properties/_forceUpdate",
```
sonofflynn89, line 59: Wait, we need to specify `/properties/_forceUpdate`? I would have figured that `/schedule/_forceUpdate` would have been enough.

sonofflynn89 (commented):
Looking good, just a couple questions.

packages/common/src/content/_internal/ContentSchema.ts
juliannemarik: Yeah, I agree this feels a little strange, but it has to do with how the tile-select field works - when it's in "checkbox" mode (e.g. multiselect), we emit an array of values. 

Not to change things, but I did just have another idea which would be to use the switch field instead. This is configured to simply emit a `true`/`false` boolean rather than an array: 

Screenshot 2024-05-02 at 1 44 57 PM
Screenshot 2024-05-02 at 1 46 20 PM

Not sure how design/Tomas would feel about this, but from the dev perspective it's a bit easier/more consistent to deal with

packages/common/src/content/edit.ts
```diff
@@ -174,6 +176,14 @@ export async function updateContent(
   }
 
   if (isDownloadSchedulingAvailable(requestOptions, content.access)) {
+    // if schedule has "Force Update" checked and clicked save, initiate an update
+    if (deepEqual(content._forceUpdate, [true])) {
+      // [true]
+      await forceUpdateContent(item.id, requestOptions);
+    }
+
+    delete (content.schedule as any)._forceUpdate;
```
sonofflynn89, line 185: Make sure we delete `content._forceUpdate` instead

sonofflynn89 (approved):
Don't merge quite yet, just one small change to be made. Once it is, you're good to go. Thank you for all this Aaron!

PR #1517, opened 2024-05-16
feat: add a partners case to well known item catalog

benstoltz:
1. Description: Adds a partners case to wellKnownItemCatalog and uses it over in getAvailableToRequestAssociationCatalogs

1. Instructions for testing:

### DIFF
--- a/packages/common/src/associations/wellKnownAssociationCatalogs.ts
+++ b/packages/common/src/associations/wellKnownAssociationCatalogs.ts
@@ -161,15 +161,22 @@ export const getAvailableToRequestAssociationCatalogs = (
     associationType
   )?.filters;
   // Get Community org id. If the current org is a community org, use the first trusted org id, otherwise fall back to the communityOrgId
-  const communityOrgId = context?.isCommunityOrg
-    ? context?.trustedOrgIds[0]
-    : context?.communityOrgId;
+  // First pull out the community org / enterprise org relationship
+  const communityTrustedOrgRelationship = context.trustedOrgs?.find(
+    (org) => org.from.orgId === context.currentUser.orgId
+  );
+  // If we are in a community org, and there is a trusted org relationship, use the orgId from there (which would be the enterprise org id)
+  // Otherwise, use the community org id
+  // If neither exist, the communityOrgId will be undefined (which is fine, as it will not be added as a catalog in that case)
+  const communityOrgId =
+    context.isCommunityOrg && communityTrustedOrgRelationship
+      ? communityTrustedOrgRelationship.to.orgId
+      : context.communityOrgId;
   // Get trusted orgs that aren't the current user's org or the community org
   const trustedOrgIds =
-    context?.trustedOrgIds?.filter((orgId) => {
+    context.trustedOrgIds?.filter((orgId) => {
       return (
-        orgId !== context?.currentUser?.orgId &&
-        orgId !== context?.communityOrgId
+        orgId !== context.currentUser?.orgId && orgId !== context.communityOrgId
       );
     }) || [];
   // Default catalogs to include
--- a/packages/common/src/associations/wellKnownAssociationCatalogs.ts
+++ b/packages/common/src/associations/wellKnownAssociationCatalogs.ts
@@ -160,18 +160,27 @@ export const getAvailableToRequestAssociationCatalogs = (
     entity,
     associationType
   )?.filters;
+  // Get Community org id. If the current org is a community org, use the first trusted org id, otherwise fall back to the communityOrgId
+  const communityOrgId = context?.isCommunityOrg
+    ? context?.trustedOrgIds[0]
+    : context?.communityOrgId;
   // Get trusted orgs that aren't the current user's org or the community org
   const trustedOrgIds =
-    context.trustedOrgIds &&
-    context?.trustedOrgIds.filter((orgId) => {
+    context?.trustedOrgIds?.filter((orgId) => {
       return (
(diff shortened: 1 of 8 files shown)


packages/common/src/associations/wellKnownAssociationCatalogs.ts
```diff
@@ -160,17 +160,40 @@ export const getAvailableToRequestAssociationCatalogs = (
...
+      return orgId !== context.currentUser?.orgId && orgId !== communityOrgId;
+    }) || [];
+  // Default catalogs to include
+  const catalogNames: WellKnownCatalog[] = ["myContent", "organization"];
+  // If there is a community org id, include the community catalog
+  if (communityOrgId) {
+    catalogNames.push("community");
+  }
+  // If there are trusted orgs, include the partners catalog
+  if (trustedOrgIds && trustedOrgIds.length > 0) {
```
juliannemarik, line 187: minor: I think you can just check `if(trustedOrgIds.length) {...}` since you're defaulting `trustedOrgIds` to an empty array above.

juliannemarik (approved):
The association side of things looks good to me, and I think your addition to context makes sense, but you may want to get verification from platform on adding that (if you haven't already).

packages/common/src/associations/wellKnownAssociationCatalogs.ts
```diff
@@ -160,17 +160,40 @@ export const getAvailableToRequestAssociationCatalogs = (
     entity,
     associationType
   )?.filters;
-  const catalogNames: WellKnownCatalog[] = [
-    "myContent",
-    "favorites",
-    "organization",
-    "world",
-  ];
+  // Get Community org id. If the current org is a community org, use the first trusted org id, otherwise fall back to the communityOrgId
```
juliannemarik, line 163: Okay, I gave all of this a bit more thought after our conversation about the `od-ui` side of things. I'm now thinking that some of this logic should be moved around. From what I can tell, none of the logic here is specific to associations - e.g. anytime we want a "My partners" or "My community" facet, all of this pre-processing logic will remain the same. Same with the "My community" label logic that you're overriding in `od-ui`. 

For this reason, I'd propose that all of this logic should live in the `wellKnownCatalog` utils themselves. This way, all the consumer has to do is call `getWellKnownCatalog` with "partners" or "community", and the util will take care of the rest. This will ensure consistency. 

I did a quick prototype off of your branch that you can see [here](https://github.com/Esri/hub.js/compare/poc/jmk-prototoype?expand=1). I quickly tested it against the app, and everything _appears_ to work the same, but I could certainly be missing something. 

We can follow-up tomorrow morning, but interested to hear your thoughts.

benstoltz: And 'lo the refactor is done.

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -40,6 +43,8 @@ export interface IGetWellKnownCatalogOptions {
   collectionNames?: WellKnownCollection[];
   /** additional filters to apply to the catalog scope */
   filters?: IFilter[];
+  /** optional context */
+  context?: IArcGISContext;
```
juliannemarik, line 47: I'm wondering if this should be pulled out into one of the top-level args for the `getWellKnownCatalog` function. Anything in `options` should really be optional since we're using the same interface for all catalogs...and `context` is technically required for the "community" + "partner" catalogs. As an aside, I also think `context` removes the need to pass in `user` as an option (which also doesn't _really_ appear to be optional). 

We could also try to do something with discriminated unions depending on the `catalogName` 🤔 but in any case, I still think `context` should be a required arg for this function. Unfortunately even if we pull it out into a top-level arg, we'll still have to make it optional to avoid a breaking change. Maybe we can leave a note to change this in the next breaking change though.

packages/common/src/search/_internal/buildCatalog.ts
```diff
@@ -15,7 +15,8 @@ export function buildCatalog(
   catalogName: string,
   filters: IFilter[],
   collections: IHubCollection[],
-  targetEntity: EntityType
+  targetEntity: EntityType,
+  title?: string
```
juliannemarik, line 19: let's add a `@param` comment for `title` above

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -148,6 +154,55 @@ function getWellknownItemCatalog(
         "item"
       );
       break;
+    case "partners":
+      validateUserExistence(catalogName, options);
```
juliannemarik, line 158: do we actually need to do this?

packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -148,6 +154,55 @@ function getWellknownItemCatalog(
...
+            },
+            ...additionalFilters,
+          ],
+          collections,
+          "item"
+        );
+      }
+      break;
+    case "community":
+      validateUserExistence(catalogName, options);
```
juliannemarik, line 188: same here - we don't actually use `options.user` to construct these catalogs

juliannemarik (commented):
Thanks for moving stuff around! I think it's looking really good. Just a few more minor comments

packages/common/src/search/wellKnownCatalog.ts
benstoltz: As discussed I've added two TODO's to the interface and to the actual function.

juliannemarik (approved):

PR #1518, opened 2024-05-16
feat(hub-common): add isHubHome prop

juliannaeapicella:
1. Description: Adds the isHubHome prop so we can determine if a site is an org site or not.

1. Instructions for testing:

1. Closes Issues: #10379

### DIFF
--- a/packages/common/src/core/types/IHubSite.ts
+++ b/packages/common/src/core/types/IHubSite.ts
@@ -81,7 +81,7 @@ export interface IHubSite
   legacyTeams: string[];
 
   /**
-   * Whether or not the site is an org site--should be removed when no longer necessary
+   * True when the site is a "Hub Home" site
    */
   isHubHome: boolean;
 }
--- a/packages/common/src/core/types/IHubSite.ts
+++ b/packages/common/src/core/types/IHubSite.ts
@@ -79,6 +79,11 @@ export interface IHubSite
 
   /** Legacy teams - list of ids */
   legacyTeams: string[];
+
+  /**
+   * Whether or not the site is an org site--should be removed when no longer necessary
+   */
+  isHubHome: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


packages/common/src/core/types/IHubSite.ts
```diff
@@ -79,6 +79,11 @@ export interface IHubSite
 
   /** Legacy teams - list of ids */
   legacyTeams: string[];
+
+  /**
+   * Whether or not the site is an org site--should be removed when no longer necessary
```
rweber-esri, line 84: I would maybe slightly amend this comment to read something like: `True when the site is a "Hub Home" site`. We probably wouldn't want to remove this, as it'd technically be a breaking change and other features might want/need to cue off of this in the future.

packages/common/src/sites/_internal/computeProps.ts
```diff
@@ -42,6 +42,10 @@ export function computeProps(
   site.updatedDate = new Date(model.item.modified);
   site.updatedDateSource = "item.modified";
   site.isDiscussable = isDiscussable(site);
+
+  // set isHubHome
+  site.isHubHome = Boolean(model.item.properties?.isHubHome);
```
rweber-esri, line 47: Upon closer look, sites has it's own [property mapper](https://github.com/Esri/hub.js/blob/10379/hide-dashboard-pane/packages/common/src/sites/_internal/getPropertyMap.ts). Maybe it'd make more sense to add an entry there to facilitate this mapping rather than manually doing it here.

```ts
  map.push({
    entityKey: "isHubHome",
    storeKey: "item.properties.isHubHome",
  });
```

Though that wouldn't coerce `undefined` values to `false`, but I _think_ that's OK.

rweber-esri (approved):
LGTM. Added 1 minor note, approved pending successful CI run.

PR #1532, opened 2024-05-29
fix(hub-common): remove addresses from event, add more event location…

velveetachef:
… properties

affects: @esri/hub-common

1. Description: Remove `addresses`, `geometry`, and `catalog` from `IEvent`, `ICreateEvent`, `IUpdateEvent`.
  Add `addNum`, `city`, `cntryName`, `nbrhd`, `placeAddr`, `placeName`, `postal`, `region`, `stDir`, `stName`, `stType`, `subRegion` to `event.location`

1. Instructions for testing:


brittneyjb (approved):
lgtm

packages/common/src/events/api/orval/api/orval-events.ts
```diff
@@ -277,6 +267,26 @@ export interface IOnlineMeeting {
   url: string;
 }
 
+export type ILocationGeometriesItem = { [key: string]: any };
```
rweber-esri, line 270: Minor/optional, I _believe_ we could alternatively use `Record<string, any>` here and for `ICreateLocationGeometriesItem`

packages/common/src/events/api/orval/api/orval-events.ts
```diff
@@ -388,15 +366,67 @@ export interface ICreateOnlineMeeting {
   url: string;
 }
 
+export type ICreateLocationGeometriesItem = { [key: string]: any };
```
rweber-esri, line 369: I find it a little strange/confusing to have `Item` in the name of this interface and `ILocationGeometriesItem`. Would it be more intuitive to name these something like `*Collection` vs `*Item`? Unsure if this would require API changes and/or complexity when updating the interfaces in the future given that we're using a generated client.

packages/common/src/events/api/types.ts
```diff
@@ -4,21 +4,18 @@ export {
...
   ICreateOnlineMeeting,
-  ICreateAddress,
   ICreateEvent,
-  ICreateEventGeometry,
   ICreateRegistration,
   IEvent,
   IEventPermission,
-  IEventCatalogItem,
-  IEventGeometry,
+  ILocation,
```
rweber-esri, line 15: Some of these interface/enum names are a bit ambiguous, while others are named more explicitly with `Event` in their name. We're not currently exporting anything from the event's `api` dir b/c the event `SortOrder` enum conflicts with one of the same name that's exported from the `discussions` subdir. Not sure the LOE, or if it's even possible given we're using an auto-generated client, to make some of these more explicit that they deal specifically with events? E.g. `IEventLocation`, `EventSortOrder`, etc?

packages/common/src/events/_internal/PropertyMapper.ts
```diff
@@ -63,7 +63,7 @@ export class EventPropertyMapper extends PropertyMapper<
     } else {
       obj.attendanceType = HubEventAttendanceType.Online;
     }
-    obj.inPersonCapacity = store.addresses?.[0]?.capacity ?? null;
+    obj.inPersonCapacity = store.inPersonCapacity ?? null;
```
rweber-esri, line 66: Minor, but most of what's in this file deals w/ assigning/deriving values that can't be mapped verbatim from the event record (e.g. mapping the capacity from the 0-index record from onlineMeetings array to the `onlineCapacity` property implemented by the entity POJO, or deriving a `boolean` for the entity POJO's `canChangeStatus` from the event record's `permission` hash). It seems we could remove this line, and simply add the string `"inPersonCapacity"` to the [getPropertyMap fn's commonPropNames array](https://github.com/Esri/hub.js/blob/f/9564-remove-address-model/packages/common/src/events/_internal/getPropertyMap.ts#L10) to achieve the same thing.

rweber-esri (approved):
Overall this looks good to me. One minor note relating to mapping the `inPersonCapacity` from the event record to the event entity POJO, and a couple minor comments/questions about naming of some of the interfaces/enums.

packages/common/src/events/edit.ts
```diff
@@ -96,6 +95,7 @@ export async function updateHubEvent(
     editGroups: model.editGroups,
     endDate: model.endDate,
     endTime: model.endTime,
+    inPersonCapacity: model.inPersonCapacity,
```
rweber-esri, line 98: We may want to stub out default values for the `inPersonCapacity` in the `packages/common/src/events/defaults.ts` module's `buildDefaultEventEntity` and `buildDefaultEventRecord` methods.

packages/common/src/events/api/orval/api/orval-events.ts
velveetachef: For generic types, this is how orval exports. I'm working on a next task to improve the validation on the geometries array, and this will be more explicit

packages/common/src/events/api/orval/api/orval-events.ts
velveetachef: This is another byproduct of the type generator. Maybe we can customize the suffix, I'll check. But also this will be more explicit with my next task

packages/common/src/events/api/types.ts
velveetachef: This can be adjusted

PR #1537, opened 2024-06-03
feat(hub-common): add register button tooltips

juliannaeapicella:
1. Description: Adds tooltips for entity hero action buttons.

1. Instructions for testing:

1. Closes Issues: #9320

### DIFF
--- a/packages/common/src/events/_internal/PropertyMapper.ts
+++ b/packages/common/src/events/_internal/PropertyMapper.ts
@@ -94,11 +94,19 @@ export class EventPropertyMapper extends PropertyMapper<
 
     const heroActions: HubActionLink[] = [];
     if (store.allowRegistration) {
+      const eventHasEnded = obj.endDateTime < new Date();
+      let tooltip;
+      if (obj.isCanceled) {
+        tooltip = "{{tooltip:register.isCancelled:translate}}";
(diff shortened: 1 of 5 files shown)


packages/common/src/events/_internal/PropertyMapper.ts
```diff
@@ -94,11 +94,19 @@ export class EventPropertyMapper extends PropertyMapper<
 
     const heroActions: HubActionLink[] = [];
     if (store.allowRegistration) {
+      const eventHasEnded = obj.endDateTime < new Date();
```
rweber-esri, line 97: Given that the `arcgis-hub-event-about.tsx` component will need to know if the event occurred in the past in order to render one of the `calcite-notice` per [this comment](https://github.com/ArcGIS/opendata-ui/pull/13599#discussion_r1622878118), it may be better to define an additional `isPast` property on the `IHubEvent` POJO and set it's value to the result of this expression somewhere higher in this method outside this if-condition (maybe after assigning `obj.isRemoved`). That would eliminate the need for us to duplicate this date comparison logic here and in that component.

rweber-esri (approved):
LGTM, one minor note that could slightly simplify things in the front-end concerning determining if an even occurred in the past.

PR #1541, opened 2024-06-05
feat(hub-common): adds map settings configuration to initiative entity

tannerjt:
affects: @esri/hub-common

ISSUES CLOSED: 10462

1. Description: Adds map configuration settings to Initiative UI Schema Settings.  This is used when constructing the map settings section on an initiative which provides the ability to select an existing web map or scene.  Future work will be done to support additional map configuration options.

1. Instructions for testing:


packages/common/src/core/schemas/shared/subschemas.ts
```diff
@@ -53,6 +53,22 @@ export const ENTITY_LOCATION_SCHEMA = {
...
 };
 
+/**
+ * defines the JSON schema for the map configuration settings
+ */
+export const ENTITY_MAP_SCHEMA = {
+  type: "object",
+  properties: {
+    itemId: {
+      type: "array",
```
sonofflynn89, line 63: Out of curiosity, why are we making this an array?

packages/common/src/core/schemas/shared/subschemas.ts
```diff
@@ -53,6 +53,22 @@ export const ENTITY_LOCATION_SCHEMA = {
   default: { type: "none" },
 };
 
+/**
+ * defines the JSON schema for the map configuration settings
+ */
+export const ENTITY_MAP_SCHEMA = {
+  type: "object",
+  properties: {
+    itemId: {
```
sonofflynn89, line 62: I understand that the full path to this property is `initiative.mapSettings.itemId`, but just a heads up that I was pretty confused when I first saw this. Out of context, I didn't know what kind of item we were talking about (and whether it was a self reference or not).

What if we made it something more explicit like `basemapItemId`?

sonofflynn89 (approved):
Looking good Josh! Just some comments for readability

packages/common/src/core/schemas/shared/subschemas.ts
tannerjt: I see.  I'm not sure `basemapItemId` is accurate since it's any web map or web scene item, not necessarily a basemap.  We could do something like `portalItemId`?

packages/common/src/core/schemas/shared/subschemas.ts
tannerjt: Mainly just for possible future use cases, such as a map switcher where we would store multiple items.  As current, we just defined maxItems and treat it like a single item id.

packages/common/src/core/schemas/shared/subschemas.ts
sonofflynn89: Or maybe `baseViewItemId`?

packages/common/src/core/types/IHubMapSettings.ts
```diff
@@ -3,5 +3,5 @@
  */
 export interface IHubMapSettings {
   // the item id for the map or scene
-  baseViewItemId: string;
+  baseViewItemId: string[];
```
sonofflynn89, line 6: Should we add a comment so others know why we're making this an array?

sonofflynn89 (approved):
Fantastic work Josh!

PR #1547, opened 2024-06-11
feat: new env permissions && permission tweaks

benstoltz:
1. Description: Adds new env permissions and slight tweaks to projects initiative workspace/tab permissions.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/permissions/HubPermissionPolicies.ts
+++ b/packages/common/src/permissions/HubPermissionPolicies.ts
@@ -109,6 +109,18 @@ const SystemPermissionPolicies: IPermissionPolicy[] = [
     environments: ["devext", "qaext"],
   },
   {
+    permission: "hub:environment:qaext",
+    environments: ["qaext"],
+  },
+  {
+    permission: "hub:environment:devext",
+    environments: ["devext"],
+  },
+  {
+    permission: "hub:environment:production",
+    environments: ["production"],
+  },
+  {
     // When enabled, the new follow card will be loaded
     // instead of the old follow initiative card.
     // To enable it, we need to pass the feature flag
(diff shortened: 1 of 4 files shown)


juliannemarik (approved):
These changes look good to me...wasn't there something we wanted to do in the `wellKnownCatalog` module to swap out the "project" collection with the "projectAndInitiative" collection? Or are we not doing that yet?

benstoltz:
We aren't doing that yet, these are still QA gated. Next pass through then we should be doing that swap I believe.

packages/common/src/permissions/HubPermissionPolicies.ts
```diff
@@ -108,6 +108,18 @@ const SystemPermissionPolicies: IPermissionPolicy[] = [
...
+    environments: ["qaext"],
+  },
+  {
+    permission: "hub:environment:devext",
+    environments: ["devext"],
+  },
+  {
+    permission: "hub:environment:production",
+    environments: ["production"],
+  },
```
jordantsanz, line 122: I'd maybe add a quick comment here saying to use these only if necessary -- we'd rather people make permissions where they specify the environment, but if that can't happen (i.e. in the case of notices where there isn't an entity type to write permissions for), then we can utilize these.

dbouwman: I agree with @jordantsanz - these should be limited use, and we definitely don't want to have them used in the dependencies arrays. But I do like using permissions >> `context.environment === "qaext"` sort of stuff

dbouwman (approved):

packages/common/src/permissions/HubPermissionPolicies.ts
benstoltz: Added :)

PR #1561, opened 2024-06-21
feat(hub-common): disable categories field and provide notice when no options are available

sonofflynn89:
1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/10382

Highlights:
- Abstracts the creation of the categories ui schema field into a helper function to maintain consistency.
- Adds a notice to all categories schema fields when no categories are available within the user's org
- Adds a `link` configuration object to `IUiSchemaMessage` that can only be used when rendering notices

### DIFF
--- a/packages/common/src/core/schemas/internal/fetchCategoriesUiSchemaElement.ts
+++ b/packages/common/src/core/schemas/internal/fetchCategoriesUiSchemaElement.ts
@@ -4,7 +4,7 @@ import {
   IUiSchemaMessage,
   UiSchemaMessageTypes,
 } from "../types";
-import { getCategoryItems } from "./getCategoryItems";
+import { fetchCategoryItems } from "./fetchCategoryItems";
 
 /**
  * Returns the UI schema element needed to render
@@ -18,7 +18,7 @@ export async function fetchCategoriesUiSchemaElement(
   i18nScope: string,
   context: IArcGISContext
 ): Promise<IUiSchemaElement> {
-  const categoryItems = await getCategoryItems(
+  const categoryItems = await fetchCategoryItems(
     context.portal.id,
     context.hubRequestOptions
   );
--- a/packages/common/src/core/schemas/internal/fetchCategoriesUiSchemaElement.ts
+++ b/packages/common/src/core/schemas/internal/fetchCategoriesUiSchemaElement.ts
@@ -55,7 +55,6 @@ export async function fetchCategoriesUiSchemaElement(
             "{{shared.fields.categories.noCategoriesNotice.link:translate}}",
           href: "https://doc.arcgis.com/en/arcgis-online/reference/content-categories.htm",
           target: "_blank",
-          icon: "launch",
         },
         allowShowBeforeInteract: true,
         alwaysShow: true,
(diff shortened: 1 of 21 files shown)


packages/common/src/core/schemas/types.ts
```diff
@@ -250,6 +250,13 @@ export interface IUiSchemaMessage {
   labelKey?: string;
   icon?: boolean | string;
   kind?: "brand" | "danger" | "info" | "success" | "warning";
+  link?: {
```
juliannemarik, line 253: I'd suggest leveraging the existing [`HubActionLink`](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/types/ActionLinks.ts) to type this property...or the more specific `IHubExternalActionLink` interface if we want to restrict it to external links. 

It doesn't look like we have an `iconEnd` property on the interface (just an `icon` property), but you can go ahead and add one to `IHubBaseActionLink` if necessary.

There's also no `labelKey` property on the interface, but we shouldn't need one - we should be able to just have a `label` and use the `{{some.key:translate}}` pattern we use elsewhere as long as we're interpolating the translations when we load the `uiSchema`. 

_Aside: my hope in the future is that we'd create a `hub-action-link` component that would accept a `link: HubActionLink` and handle all things around rendering, telemetry, etc. If this were the case, we could have logic that checks if the link is external + has a "blank" target and automatically include the launch icon at the end (as I believe this should always be shown in this case)...this would effectively remove the immediate need for an `iconEnd` in the interface, but I think it's fine to add for now_

packages/common/src/core/schemas/internal/fetchCategoriesUiSchemaElement.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Returns the UI schema element needed to render
+ * the categories editing control for an entity.
+ *
+ * @param i18nScope i18n scope for the entity translations
+ * @param entity The entity to build the UI schema for
+ * @returns the UI schema element for thumbnail editing
+ */
+export async function fetchCategoriesUiSchemaElement(
```
juliannemarik, line 17: minor: I might suggest naming this `buildCategoriesUiSchemaElement` or `getCategoriesUiSchemaElement`. In my mind, `fetch...` implies an XHR call

sonofflynn89: Great callout. This was also a point of confusion for me when I started working here, partially because the underlying `getCategoryItems()` function should really be named `fetchCategoryItems()`. I've renamed the helper accordingly so devs know that an XHR is required.

packages/common/src/core/schemas/types.ts
sonofflynn89: Done! Great suggestion @juliannemarik!

abp6318 (approved):
I agreed with the comments that Julianne left, and tested things locally (including some a11y checks). Looks good! I left a comment on the other PR about repeated strings that we might be able to reduce before you merge this.

juliannemarik (approved):
This all looks good to me as well - thanks for making the updates and for the addition to the interface!

PR #1585, opened 2024-07-09
feat: add embed types/interfaces

juliannemarik:
[10885](https://devtopia.esri.com/dc/hub/issues/10885)

### Description:
Scaffolds a new `HubEmbed` type for use within the entity class system as well as our frontend view/configuration components

### DIFF
--- a/packages/common/src/core/traits/IWithViewSettings.ts
+++ b/packages/common/src/core/traits/IWithViewSettings.ts
@@ -4,6 +4,7 @@ import {
   IMetricDisplayConfig,
   HubActionLink,
   IHubMapSettings,
+  HubEmbed,
 } from "../types";
 
 /**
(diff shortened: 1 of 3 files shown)


benstoltz (approved):

vivzhang (approved):
Looks good!

packages/common/src/core/traits/IWithViewSettings.ts
```diff
@@ -58,4 +59,7 @@ export interface IWithViewSettings {
    * array of actions for action links
    */
   heroActions?: HubActionLink[];
+
+  /** array of embedded content */
+  embeds?: HubEmbed[];
```
dbouwman, line 64: Is the assumption that the order of the embeds === order injected into the view? Or should we have a `key` prop on `IHubEmbedBase` so that we can fetch specific embed's out of the array?

dbouwman (approved):
This looks fine to start, but just wondering how the array of embeds lands in the view, and to consider that when we lay down these interfaces. I suspect we can just add something to the `IHubEmbedBase` down the line, but if there's more we can add now, lets do that.

PR #1595, opened 2024-07-18
feat: moving types and utils

abp6318:
Turned this into a brief follow-on ticket for sprint tracking sake: https://devtopia.esri.com/dc/hub/issues/10952

1. Description:
- Follow up PR for https://devtopia.esri.com/dc/hub/issues/10835 and https://devtopia.esri.com/dc/hub/issues/10834
- Previous OD-UI PR: https://github.com/ArcGIS/opendata-ui/pull/13819

1. Instructions for testing:

- Unavailable service: https://download-test-qa-pre-a-hub.hubqa.arcgis.com/maps/7cbf5f77f5c2463ab5c42619e79a11fd/about
- Slow service: https://santest-qa-pre-hub.hubqa.arcgis.com/datasets/9c58597535c24d26ae0e979f090c317e_0/explore?location=33.293344%2C-100.000000%2C4.75

1. Closes Issues: #10835 and #10834


packages/common/src/content/contentUtils.ts
```diff
@@ -437,3 +440,52 @@ const getContentRelativeUrl = (
...
   );
 };
+
+// Tests can be found in packages/common/test/content/content.test.ts
+/**
+ * Get the status of a content item
+ * @param entity the content item
+ * @returns the status of the content item
+ */
+export async function getContentStatus(
```
tomwayson, line 450: this really feels to me like it should be `getServiceStatus(url: string, requestOptions: IRequestOptions)`
 
it also feels like it should optionally allow users to override the default timeout.

So probably the best thing to do is extend [IGetLayerOptions](https://esri.github.io/arcgis-rest-js/api/feature-layer/IGetLayerOptions/) like this:

```ts
interface IGetServiceStatusOptions extends IGetLayerOptions {
  /* number of milliseconds to wait for the service response, defaults to 3000 */
  timeout?: number
}

export async function getServiceStatus(options: IGetServiceStatusOptions) {
  const { timeout, ...requestOptions } = options;
  // ....
}
````

File: packages/common/src/content/contentUtils.ts [hunk 2]
```diff
@@ -437,3 +440,52 @@ const getContentRelativeUrl = (
...
+
+  if (entity.url) {
+    // set up our two promises: one to get the service definition and one to sleep for 3 seconds
+    const definitionPromise: Promise<IFeatureServiceDefinition> = getService({
+      url: entity.url,
+      authentication: requestOptions.authentication,
+    });
+    const timeoutPromise = (ms: number) => {
+      return new Promise((resolve) => setTimeout(resolve, ms));
+    };
```
tomwayson, line 469: Looks like there are a couple of places that could use this, so I think it is worth moving to it's own util and ideally updating those places to use it (or at least adding a comment that they should use it):

```bash
packages/common/src/downloads/_internal/file-url-fetchers/fetchExportItemDownloadFileUrl.ts:
  129      progressCallback && progressCallback(DownloadOperationStatus.PROCESSING);
  130:     await new Promise((resolve) => setTimeout(resolve, pollInterval));
  131      return pollForJobCompletion(

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFileUrl.ts:
  131    // Operation still in progress, poll again
  132:   await new Promise((resolve) => setTimeout(resolve, pollInterval));
  133    return pollDownloadApi(requestUrl, pollInterval, progressCallback);

packages/common/src/items/_internal/_wait-for-item-ready.ts:
  7  function delay(milliseconds: number) {
  8:   return new Promise((resolve) => setTimeout(resolve, milliseconds));
  9  }

packages/common/src/utils/poll.ts:
  33      const results = await new Promise<T>((resolve) => {
  34:       setTimeout(async () => {
  35          const res = await requestFn();
```

packages/common/test/content/content.test.ts
```diff
@@ -1410,4 +1414,147 @@ describe("content: ", () => {
...
+        createdDateSource: "",
+        updatedDate: new Date(),
+        updatedDateSource: "",
+        type: "",
+        orgUrlKey: "",
+      };
+
+      fetchMock.once(
+        "https://hubqa.arcgis.com/api/v3/connectors/test/file-geojson/rest/services/slowpoints/FeatureServer/0?stop=false",
+        new Promise((resolve) => {
```
tomwayson, line 1470: As we discussed, it would be best to use the built-in [delay](https://www.wheresrhys.co.uk/fetch-mock/api/mocking/parameters/options/#delay) option, though I know it's a hassle b/c of the outdated types.

tomwayson (commented):
All in all this is good. I left a few ideas on how it can be improved if you have time.

packages/common/src/content/contentUtils.ts
abp6318: After further investigation, this is wayyyy more repeated than what I realized. 

Made a spin off and assigned it to platform -- https://devtopia.esri.com/dc/hub/issues/10953

tomwayson: Thanks for creating that issue, but it's not the sort of thing that a developer on any thread (least of all Platform) will have an easy time pulling into their sprint. Instead, it is the sort of thing that we should try to fold into our work, in little steps if needed. The first step, and the minimum amount of work is that you should do in this PR, is create the hub.js fn and use it in the new code you are introducing. You might be able to avoid writing a test for it by using it in at least one of the other places that already has coverage. If it's too big a lift to update all the other places, that's fine, but the next step is to at least add a comment to the other places to say that they should be refactored to use new fn. Finally, you can scope this to hub.js and hub-components; no need to update the Ember code at this point. 
So ideally this PR will:
- add the new fun
- use that fn in your code
- get coverage for that fn either through using in an existing fn w/ coverage or writing a test
- either replace or add comments to other hub.js locations

And then your PR that updates opendata-ui to use the fn you introduce in this PR would:
- either replace or add comments to the hub-components locations

packages/common/src/content/types.ts
```diff
@@ -93,4 +92,7 @@ export interface IHubOtherContentStatus extends IHubBaseContentStatus {
   kind: "other";
   // TODO: expand in the future if needed
 }
+export interface IGetServiceStatusOptions extends IGetLayerOptions {
+  timeout?: number;
+}
```
tomwayson, line 97: I think this should live next to `getServiceStatus` in `contentUtils.ts` and be exported from there

packages/common/src/content/contentUtils.ts
```diff
@@ -437,3 +438,69 @@ const getContentRelativeUrl = (
...
+ * @returns the status of the content item
+ */
+export async function getServiceStatus(
+  entity: IHubEditableContent,
+  options: IGetServiceStatusOptions
+): Promise<IHubServiceBackedContentStatus> {
+  // get the request options for the `getService` call, and set a default timeout if one is not provided
+  const { timeout = 3000, ...requestOptions } = options;
+
+  if (entity.url) {
```
tannerjt, line 478: As outcome of current [opendata-ui PR](https://github.com/ArcGIS/opendata-ui/pull/13855) to address non-service backed items, we should add this fix here as well.

tannerjt (approved):
LGTM, pending porting over checking of entity URL to make sure it's a service backed entity.

packages/common/src/content/contentUtils.ts
abp6318: I think I've made all the updates needed. @tomwayson, can you confirm this PR looks good to you before I merge?

packages/common/src/content/contentUtils.ts
```diff
@@ -453,44 +466,37 @@ const getContentRelativeUrl = (
...
-    const status = Promise.race([definitionPromise, timeoutPromise(timeout)])
-      .then((result) => {
-        return {
-          kind: "service",
-          service: {
-            // if we got a result, the service is available, otherwise it's slow
-            availability: result ? "available" : "slow",
-          },
-        };
+    })
```
tomwayson, line 479: Any reason we wouldn't use the rest of `requestOptions` and not just `authentication`?

```ts
const { url } = entity;
const definitionPromise = getService(...requestOptions, url);
```

packages/common/src/content/contentUtils.ts
```diff
@@ -22,11 +22,11 @@ import { getFamily } from "./get-family";
...
 import {
   DatasetResource,
-  IGetServiceStatusOptions,
+  IHubContentStatus,
   IHubServiceBackedContentStatus,
 } from "./types";
-import { IFeatureServiceDefinition } from "@esri/arcgis-rest-types";
-import { getService } from "@esri/arcgis-rest-feature-layer";
+import { getService, IGetLayerOptions } from "@esri/arcgis-rest-feature-layer";
+import { isService } from "../resources/_internal/_validate-url-helpers";
```
tomwayson, line 29: did you want to add a commit to this PR that exposes this as `isServiceUrl()` for use in hub-components?

tomwayson (approved):
Two small suggestions, but otherwise this looks good.

PR #1619, opened 2024-08-08
feat: adding a third content service status

abp6318:
1. Description:
related od-ui PR: https://github.com/ArcGIS/opendata-ui/pull/13913

1. Instructions for testing:

- copy this build into opendata-ui with this branch https://github.com/ArcGIS/opendata-ui/pull/13913
- here is my test item for "unknown" status: https://download-test-qa-pre-a-hub.hubqa.arcgis.com:4200/maps/b1a9d8f10ac9467a8d4f5593ba84ef51/about
- here is San's test item for "unavailable" status: https://qa-pre-hub.hubqa.arcgis.com:4200/maps/d35ec2e90db247a396763a785c53f864/about

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/10973


packages/common/src/content/contentUtils.ts
```diff
@@ -487,9 +487,14 @@ export async function getServiceStatus(
...
-      .catch(() => {
+      .catch(async (error) => {
         // if the service is not returned, we consider it unavailable
-        return availability("unavailable");
+        // if the service responds with a 403, we consider it unknown
+        const response = await error.response.json();
+
+        return response.error.code === 403
+          ? availability("unknown")
+          : availability("unavailable");
```
tannerjt, line 497: I have a test item, and am observing a `499` error code.

https://coffee-culture-qa-pre-a-hub.hubqa.arcgis.com/maps/33e105ebafb24bd497da4491f15cc0d6/about

I think we may need to account for this case here.

tannerjt (commented):
Added one comment to test against my sample item and check HTTP status code.

packages/common/src/content/contentUtils.ts
abp6318: @tannerjt + @sonofflynn89 + @drspacemanphd 

So I tried resolving this today but I'm running into several odd cases that are awkward to cover correctly:

1. 403 example ([link](https://rochester-moose-qa-pre-a-hub.hubqa.arcgis.com:4200/maps/b1a9d8f10ac9467a8d4f5593ba84ef51/about)) - my test item
Screenshot 2024-08-12 at 11 22 19 AM

2. 499 example ([link](https://coffee-culture-qa-pre-a-hub.hubqa.arcgis.com:4200/maps/33e105ebafb24bd497da4491f15cc0d6/about)) - Josh's item
Screenshot 2024-08-12 at 11 23 37 AM
**note**: the caught error does not have a `response` property :/

3. 500 example ([link](https://qa-pre-hub.hubqa.arcgis.com:4200/maps/d35ec2e90db247a396763a785c53f864_0/about)) ([link to stop San's server](https://hubqa.arcgis.com/api/v3/connectors/test/file-geojson/rest/services/multipoints/FeatureServer?stop=true)) -- is San's stopped server that doesn't mimic server responses necessarily
Screenshot 2024-08-12 at 11 15 31 AM

During our afterparty at stand-up today, we were trying to figure out which bases need covering, and what counts as `unknown` and what counts as `unavailable`.

In my head, I was thinking 403 and 499 --> `unknown`; otherwise, `unavailable`. Do any of you know of other responses we can expect?

sonofflynn89: @abp6318 thank you for documenting this, there are a couple of things at play:

> note: the caught error does not have a response property

You correctly noted that the response structures look different between errors. That's because there are two types of error objects that you can receive:
- Vanilla http error objects (no response object)
- Rest.js ArcGIS Errors (ArcGISRestRequestError, ArcGISAuthError, etc.)

We will need to account for both of them. Just assume that if a "response" object is present, you're dealing with an ArcGIS***Error. 

> In my head, I was thinking 403 and 499 --> unknown; otherwise, unavailable. Do any of you know of other responses we can expect?

There are many other status codes that we can receive during this operation. Just look at this list that we have in the indexer (which are all probably codes that we've received at some point):

image

Here's my gut reaction to what we should probably do:
- 404 (not found) -> **unavailable**, because the resource does not exist
- 5xx -> **unavailable**, because the server is having internal problems that need to be rectified
- 401 (unauthorized), 403 (forbidden) -> **unknown**, because we do not have credentials to check
- 499 -> **unknown**, because this seems to be an AGO / Enterprise specific code that is returned when the system cannot validate the token provided
- Other 4xx status codes -> **unavailable**, because something else went wrong and the **unknown** notice might not apply

packages/common/src/content/contentUtils.ts
```diff
@@ -487,9 +487,19 @@ export async function getServiceStatus(
...
+        // see interface IHubServiceBackedContentStatus for possible
+        // availability values and what each signifies
+        if (error.response) {
+          const statusCode = error.response.error.code;
+          const requiresAuth = [401, 403, 499].includes(statusCode);
+          return requiresAuth
+            ? availability("auth-required")
+            : availability("unavailable");
+        } else {
+          // sometimes, the 499 status code is returned as a "failed to fetch" error
```
sonofflynn89, line 500: This is fine for now, but I wonder if we will run into an issue down the line where an unrelated "failed to fetch" error will incorrectly show the "auth-required" message. I think we are OK to proceed, but I'm curious if @tannerjt has any thoughts on this.

sonofflynn89 (approved):
Tentatively approving with one comment. Thank you so much for doing this!

PR #1634, opened 2024-08-23
feat(): scaffold user workspace

jordantsanz:
1. Description:

1. Instructions for testing:


packages/common/src/users/_internal/UserUiSchemaSettings.ts
```diff
@@ -0,0 +1,45 @@
...
+  context: IArcGISContext
+): Promise<IUiSchema> => {
+  return {
+    type: "Layout",
+    elements: [
+      {
+        type: "Section",
+        labelKey: `${i18nScope}.sections.userSettings`,
+        elements: [
+          {
```
juliannemarik, line 27: Not entirely sure _where_ the best place to put a note/comment would be (maybe in the harness caveats section?):
Screenshot 2024-08-30 at 11 16 16 AM

But I went to verify functionality of this toggle in the harness, and I don't _think_ it works there because the user hub settings don't appear to be populated in the harness? Not entirely sure what's going on (maybe it's just me), but everything appears to work correctly on QA.

packages/common/src/users/HubUser.ts
```diff
@@ -0,0 +1,125 @@
...
+   * Save the HubUser to the backing store
+   * TODO: fill this out
+   */
+  async save(): Promise<void> {
+    if (this.isDestroyed) {
+      throw new Error("HubUser is already destroyed.");
+    }
+
+    // 1. update user hub settings
+    await this.context.updateUserHubSettings(this.entity.settings);
```
juliannemarik, line 57: I think this is perfectly fine as-is, but now it just has me wondering about how we're doing things for other entities...basically I'm wondering if this should happen in `save` or in `fromEditor` 🤔 

In projects, for example, we make the request to upsert/remove the configured featured image directly in `fromEditor` before calling `save`.

In general (and as we've discussed), I think there's a lot of ambiguity in this whole `EntityEditor` system, so I'm not actually sure what to suggest or if one way is better than the other.

juliannemarik (approved):
This all looks good to me! As noted though, probably worth getting another look from Dave/Tom since you're scaffolding a new entity class

packages/common/src/users/_internal/computeProps.ts
```diff
@@ -0,0 +1,38 @@
...
+import { IRequestOptions, request } from "@esri/arcgis-rest-request";
+
+export async function computeProps(
+  model: IUser,
+  user: IHubUser,
+  context: IArcGISContext
+): Promise<IHubUser> {
+  user.settings = context.userHubSettings;
+
+  const signinSettings = await getPortalSignInSettings(context);
```
dbouwman, line 15: consider not even fetching this if the user lacks privs to actually update them

dbouwman (approved):
DM'ed a bunch of suggestions to @jordantsanz - mainly around adding more comments in various sections of the code, not loading the org settings if the user lacks platform privs to update the settings. Also hiding the UX for those under same conditions. Otherwise, the code looks good :)

PR #1650, opened 2024-09-16
fix: revert 11360 changes until we hear from AGO

abp6318:
Reverts Esri/hub.js#1642

Reverting until we can confirm with AGO an accurate way of determining if a service is a secure proxy service based in the downloads API due to a related bug


sonofflynn89 (approved):
Thank you!

PR #1655, opened 2024-09-18
feat: user settings org section & notice

benstoltz:
1. Description: Adds new org settings section to User config && a notice.

1. Instructions for testing:


vivzhang (approved):
LGTM!

packages/common/src/users/_internal/UserUiSchemaSettings.ts
```diff
@@ -16,6 +17,31 @@ export const buildUiSchema = async (
...
+    // set the community org name
+    communityOrgName = context.trustedOrgs.find(
+      (org) => org.to.orgId === context.communityOrgId
+    ).to.name;
+    // add the community org action
+    orgNoticeActions.push({
+      ariaLabel: `{{${i18nScope}.notice.actions.goToCommunityOrg:translate}}`,
+      label: `{{${i18nScope}.notice.actions.goToCommunityOrg:translate}}`,
+      icon: "launch",
+      href: `${context.communityOrgUrl}/home/organization.html?tab=general#settings`,
```
juliannemarik, line 41: I might be completely mis-remembering our conversation, but wasn't this causing issues to link out to settings directly?

juliannemarik (changes requested):
A few other general comments after reviewing this on QA: I'm noticing a couple unexpected things/strings we may want to tweak:

1. When I go to a community org, I'm not seeing the attribution for the employee org even if one exists. Sorry if this wasn't clear, but we want the "Your employee/community organization is: <e-org.name/c-org.name> from _both_ directions. So here, I'd expect to see "Your employee organization is: QA Premium Alpha Hub" below the initial message in the notice and a second link out to the employee org settings. 
   Screenshot 2024-09-20 at 12 05 36 PM
2. Since the original notice message refers to "staff and community", I'm wondering if we should replace all instances of "employee organization" with "staff organization"? I'm not actually sure what's known to customers, but being consistent is probably a good idea.
3. The link out to org settings in the screenshot above inaccurately says "Go to employee organization settings" even though it should be linking out to the community org settings since that's the context we're in. I'm wondering if the first link should just drop "employee/community" altogether, and just say "Go to organization settings". The second link can then switch between saying "Go to community organization settings"/"Go to staff organization settings"

packages/common/src/users/_internal/UserUiSchemaSettings.ts
benstoltz: It does, I haven't discussed the issue with Ike yet :)

packages/common/src/users/_internal/UserUiSchemaSettings.ts
```diff
@@ -39,6 +82,66 @@ export const buildUiSchema = async (
...
+              },
+            },
+          },
+        ],
+      },
     ],
   };
 };
+
+async function _getCommunityOrEnterpriseUrl(
```
juliannemarik, line 127: Two thoughts/suggestions: 
1. Maybe call this `getCommunityOrEnterpriseAGOUrl`?
2. Just wondering if this should be abstracted out into a reusable util?

packages/common/src/users/_internal/UserUiSchemaSettings.ts
```diff
@@ -16,6 +18,47 @@ export const buildUiSchema = async (
   options: any,
   context: IArcGISContext
 ): Promise<IUiSchema> => {
+  let associatedOrgName;
+  let noticeMessage = `{{${i18nScope}.notice.message:translate}}`;
+  // default Notice action
```
juliannemarik, line 23: maybe make this more explicit: "default notice action - "Go to organization settings""

packages/common/src/users/_internal/UserUiSchemaSettings.ts
```diff
@@ -16,6 +18,47 @@ export const buildUiSchema = async (
...
+  const orgNoticeActions = [
+    {
+      ariaLabel: `{{${i18nScope}.notice.actions.goToOrg:translate}}`,
+      label: `{{${i18nScope}.notice.actions.goToOrg:translate}}`,
+      icon: "launch",
+      href: `${context.portalUrl}/home/organization.html?tab=general#settings`,
+      target: "_blank",
+    },
+  ];
+  // If there is a community org relationship, or we are in a community org with an enterprise org relationship
```
juliannemarik, line 33: And then here, extend this to _something_ like: 

```ts
// If there is a community org relationship, or we are in a community org with an enterprise org relationship, show another action that links out to the corresponding relationship org ("Go to community organization" or "Go to staff organization")
```

juliannemarik (approved):
This all looks good to me - left a few more minor suggestions for you to take/leave, but otherwise I think this is good to go. Nice work on this craziness!

PR #1657, opened 2024-09-19
feat(hub-common): allow force updates of cache files as well as cache status reporting

sonofflynn89:
1. Description:

Part of https://devtopia.esri.com/dc/hub/issues/11445

This PR is meant to reflect new realities in our Download API, namely:
- If the download api serves us a cached file, we propagate the status of that cache to the caller
- Allow callers of `fetchDownloadFile()` to update the cache by providing the "updateCache" option

As an aside, we decided to remove the implicit requirement that `fetchDownloadFile()` would call `progressCallback` with a completed status. It's now up to the callers of `fetchDownloadFile()` to inform interested parties that the download is complete.


packages/common/src/downloads/_internal/file-url-fetchers/fetchExportImageDownloadFile.ts
```diff
@@ -53,7 +53,6 @@ export async function fetchExportImageDownloadFile(
   // }
 
   const blob: Blob = await request(`${entity.url}/exportImage`, requestOptions);
-  progressCallback && progressCallback(DownloadOperationStatus.COMPLETED);
```
sonofflynn89, line 56: As part of this PR, we made the decision that `fetchDownloadFile()` delegates are no longer responsible for emitting the "complete" event. It is up to the caller of `fetchDownloadFile()` to notify that the download is complete.

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFile.ts
```diff
@@ -97,44 +114,77 @@ function getDownloadApiRequestUrl(options: IFetchDownloadFileOptions) {
...
     return {
       type: "url",
       href: resultUrl,
+      cacheStatus,
     };
   }
 
-  // Operation still in progress, poll again
+  // Operation still in progress. Report progress if a callback was provided and poll again.
+  progressCallback && progressCallback(operationStatus, progressInPercent);
```
sonofflynn89, line 172: Moving this for the same reason as above. We don't want to emit the "complete" event from within this function.

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFile.ts
```diff
@@ -97,44 +114,77 @@ function getDownloadApiRequestUrl(options: IFetchDownloadFileOptions) {
...
   await new Promise((resolve) => setTimeout(resolve, pollInterval));
```
abp6318, line 173: we have a `wait(...)` helper function that does this I think

abp6318 (approved):
Also looks good to me. Brain very tired so hopefully I didn't miss anything on this one.

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFile.ts
```diff
@@ -97,44 +114,77 @@ function getDownloadApiRequestUrl(options: IFetchDownloadFileOptions) {
...
   requestUrl: string,
   pollInterval: number,
+  cacheQueryParam?: CacheQueryParam,
   progressCallback?: downloadProgressCallback
 ): Promise<IFetchDownloadFileResponse> {
-  const response = await fetch(requestUrl);
+  // If requested, append the cache query param to the request URL
+  let withCacheQueryParam = requestUrl;
+  if (cacheQueryParam) {
+    withCacheQueryParam = `${requestUrl}&${cacheQueryParam}=true`;
```
tannerjt, line 134: Are we assuming that existing query params are present?  I'm just wondering `?` vs `&` is going to cause an issue, eg:

`${requestUrl}?${cacheQueryParam}=true` vs `${requestUrl}&${cacheQueryParam}=true`

tannerjt (approved):
LGTM, one small question otherwise good to go!

packages/common/src/downloads/_internal/file-url-fetchers/fetchHubApiDownloadFile.ts
sonofflynn89: Good question @tannerjt! As of now, the Download API requires a `?layers` query param in order to function:

`/v1/:itemId/:format?layers=<:layers>`

That's why we can reliably just add &.

PR #1668, opened 2024-09-25
feat(hub-common): add site URL to settings schema

tomwayson:
affects: @esri/hub-common

1. Description:

1. Instructions for testing:


packages/common/src/core/types/IHubSite.ts
```diff
@@ -7,12 +7,30 @@ import { IWithCatalog, IWithCatalogs } from "../traits/IWithCatalog";
 import { IWithLayout } from "../traits/IWithLayout";
 import { IWithPermissions } from "../traits/IWithPermissions";
 
+export interface IHubSiteUrlInfo {
```
dbouwman, line 10: maybe add a comment that while not used for any other entity (thus why it's in this file) this is extracted into an interface to ensure IHubSiteEditor shares a consistent structure (this took a me little 🤔)

packages/common/src/sites/_internal/SiteSchema.ts
```diff
@@ -17,11 +17,17 @@ export const SiteEditorTypes = [
...
+  ({
+    $async: true,
+    ...HubItemEntitySchema,
+    properties: {
+      ...HubItemEntitySchema.properties,
+      _discussions: ENTITY_IS_DISCUSSABLE_SCHEMA,
+      telemetry: PRIVACY_CONFIG_SCHEMA,
+      _urlInfo: {
+        type: "object",
+        isUniqueDomain: { siteId },
```
dbouwman, line 30: I'm far from a json schema expert, but how does this work when creating a new site, when there is no siteId?

packages/common/test/sites/_internal/SiteuUiSchemaTelemetry.test.ts
```diff
@@ -1,15 +1,44 @@
 import { buildUiSchema } from "../../../src/sites/_internal/SiteUiSchemaSettings";
 import { MOCK_CONTEXT } from "../../mocks/mock-auth";
 
-describe("buildUiSchema: site telemetry", () => {
-  it("returns the full site telemetry uiSchema", async () => {
-    const uiSchema = await buildUiSchema("some.scope", {} as any, MOCK_CONTEXT);
+describe("buildUiSchema: site settings", () => {
```
dbouwman, line 4: Since you renamed the test, should the file name also change?

dbouwman (approved):
Code looks good - but make sure this works w/ the "create site" flow b/c I'm not clear how getSiteSchema works if there is no siteId...

packages/common/src/sites/_internal/SiteSchema.ts
tomwayson: that's a good question.

short answer: it's fine

long answer: the keyword validator that this schema is passed into only uses the `siteId` if it was passed in. Here's the logic:

```ts
  // siteId is only passed in when editing an existing site
  const siteId = schema?.siteId;
  let isValid;
  try {
    const domain = await lookupDomain(hostname, context.hubRequestOptions);
    // if lookupDomain does NOT throw, we found a domain record
    // which is only valid if it is the one associated with the site we are editing
    isValid = !!siteId
      ? domain.siteId === siteId
      : false;
  } catch (_) {
    // no domain found, so it's available
    isValid = true;
  }
  return isValid;
```

packages/common/test/sites/_internal/SiteuUiSchemaSettings.test.ts
```diff
@@ -0,0 +1,55 @@
+import { buildUiSchema } from "../../../src/sites/_internal/SiteUiSchemaSettings";
```
juliannemarik, line 1: minor: looks like there's still a small typo in the file name? Should "SiteuUiSchemaSettings.test.ts" should be "SiteUiSchemaSettings.test.ts"?

juliannemarik (approved):
This all looks good to me!

PR #1676, opened 2024-10-03
feat: showInfoBanner for user org settings

benstoltz:
1. Description: Adds a showInfoBanner section to User Org Settings. Also adds the save logic.

1. Instructions for testing:

### DIFF
--- a/packages/common/src/utils/internal/updateCommunityOrgSettings.ts
+++ b/packages/common/src/utils/internal/updateCommunityOrgSettings.ts
@@ -1,4 +1,3 @@
-import { _isOrgAdmin } from "../..";
 import { IArcGISContext } from "../../ArcGISContext";
 import { IHubUserOrgSettings } from "../../core/types/IHubUser";
 import { request } from "@esri/arcgis-rest-request";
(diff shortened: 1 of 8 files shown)


packages/common/src/utils/internal/updateCommunityOrgSettings.ts
```diff
@@ -1,4 +1,4 @@
-import { _isOrgAdmin, getPortalUrl } from "../..";
+import { _isOrgAdmin } from "../..";
```
jordantsanz, line 1: This is my bad I think, but can we update the import to point to the specific file here?

packages/common/src/utils/internal/updatePortalOrgSettings.ts
```diff
@@ -0,0 +1,44 @@
+import { _isOrgAdmin, cloneObject } from "../..";
```
jordantsanz, line 1: Same thing with updating import here

packages/common/src/utils/internal/updateCommunityOrgSettings.ts
benstoltz: Actually we can just remove it altogether since we are checking for org admin off context.

PR #1684, opened 2024-10-10
feat(hub-common): add location props

juliannaeapicella:
1. Description: Adds props for location name.

1. Instructions for testing:

1. Closes Issues: #11062


packages/common/test/events/_internal/EventSchemaEdit.test.ts
```diff
@@ -80,7 +80,17 @@ describe("EventSchemaEdit", () => {
             default: HubEventCapacityType.Unlimited,
           },
           location: {
-            type: "object",
+            ...ENTITY_LOCATION_SCHEMA,
+            allOf: [
+              {
```
rweber-esri, line 85: Are these changes needed? I see on your opendata-ui PR that you created a separate schema and uiSchema for the location field. This schema & uiSchema has to do with the event data structure though. Given that, `type` would never have a value of `custom`, `org` or `none`, but rather something like `Event`. The `name` property you're attempting to make required here is the entity `name`, not the location name... If my understanding is correct, we can close this PR and your opendata-ui PR doesn't need to pull in these changes.

rweber-esri: I misread the code when I left the above comment, I now see that the `allOf` here is a descendant of the `location` property; I thought this was added at the root Event schema level. Jules confirmed w/ Julianne this is needed.

rweber-esri (approved):
LGTM

PR #1756, opened 2024-12-04
feat(hub-common): add fetchItem and call it from fetchPage

tomwayson:
1. Description:

- adds `fetchItem()` which:
  - uses a new internal `parseIdentifier()` to parse the id or slug out of the identifier
  - allows callers to pass in the site's org key as a fallback for slugs w/o an org key prefix
  - encapsulates the logic for calling the appropriate function to fetch via id or slug
- `fetchPage()` now defers to `fetchItem()` to handle all the above (we will do the same for other item-backed entities in a future PR)

1. Instructions for testing:

### DIFF
--- a/packages/common/src/items/_internal/slugs.ts
+++ b/packages/common/src/items/_internal/slugs.ts
@@ -1,3 +1,10 @@
+import { isGuid } from "../../utils/is-guid";
+
+const SLUG_ID_SEPARATOR = "~";
+
+// TODO: use this const in other utils
+const SLUG_ORG_KEY_SEPARATOR = "::";
+
 const TYPEKEYWORD_MAX_LENGTH = 256;
 
 export const TYPEKEYWORD_SLUG_PREFIX = "slug";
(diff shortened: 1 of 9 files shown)


packages/common/src/items/_internal/slugs.ts
```diff
@@ -40,3 +47,42 @@ export function getSlugMaxLength(orgKey: string) {
...
+  /* org key */
+  orgKey?: string;
+}
+
+/**
+ * parse out the item's id, slug, and org key out of an identifier
+ * @param identifier
+ * @returns
+ */
+export const parseIdentifier = (identifier: string): IParsedIdentifier => {
```
dbouwman, line 65: This seems like it could also work for an Events, where it would return the CUID vs a GUID... Wondering if we can/should hoist this out of /items

packages/common/src/items/fetch.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  siteOrgKey?: string;
+}
+
+/**
+ * Fetch an item by its identifier
+ * @param identifier item id or slug, which may or may not include the org key
+ * @param requestOptions
+ */
+export function fetchItem(
```
dbouwman, line 22: Can/should fetchEvent have similar logic

dbouwman (approved):
This looks good for items, left few comments re: use for other entities

packages/common/src/items/_internal/slugs.ts
tomwayson: tbh - I don't know how fetching events work at all, but the great part of this being internal is that we can refactor as needed later.

PR #1788, opened 2025-01-30
fix(hub-common): update spatial element from default dcat us template

sansth1010:
1. Description: This PR updates `spatial` field of default DCAT US feed template to use default transformer `computeSpatialProperty` with `extent` value. [`computeSpatialProperty`](https://github.com/ArcGIS/hub-feeds-api/blob/main/src/default-feed-templates/dcat-us/template-transforms.js#L164-L189) computes valid value from extent so should be used with `extent` by default.

1. Instructions for testing:

1. Closes Issues: [#12271]( https://devtopia.esri.com/dc/hub/issues/12271)

### DIFF
--- a/packages/common/src/sites/feeds/_internal/defaults.ts
+++ b/packages/common/src/sites/feeds/_internal/defaults.ts
@@ -31,7 +31,6 @@ const DCAT_US_1X_DEFAULT = {
     fn: "{{owner}}",
     hasEmail: "{{orgContactEmail}}",
   },
-  spatial: "{{extent}}",
 };
 
 const DCAT_US_3X_DEFAULT = {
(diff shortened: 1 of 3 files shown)


packages/common/src/sites/feeds/_internal/defaults.ts
```diff
@@ -31,7 +31,7 @@ const DCAT_US_1X_DEFAULT = {
     fn: "{{owner}}",
     hasEmail: "{{orgContactEmail}}",
   },
-  spatial: "{{extent}}",
+  spatial: "{{extent:computeSpatialProperty}}",
```
sonofflynn89, line 34: This will work for _new_ sites, or sites that don't have a dcat configuration previously saved. What do we want to do about sites that have "spatial: "{{extent}}" hard coded in their saved templates?

sansth1010: > What do we want to do about sites that have "spatial: "{{extent}}" hard coded in their saved templates?

We will have to request those customers to update their saved templates manually through feed editor. The sites affected by the bug are the ones that don't have dcat configurations saved.

sonofflynn89 (approved):
Great work San! Thank you!

PR #1808, opened 2025-02-20
feat: add IHubOrganization with fetch & hubSearch support

dbouwman:
1. Description:
- Adds `IHubOrganization` as a type of `HubEntity`
-  Adds support for "organization" to `fetchHubEntity`, which delegates to `fetchOrganization` fn
- adds super limited support to `hubSearch` for targetEntity "organization"; Since there is no portal search api, it only respects the `id` predicate, and simply batch fetches the specified ids. If the backend returns an error that is simply omitted in the IHubSearchResponse. This was added to simplify the use of the `arcgis-hub-gallery-layout-compact`

1. Instructions for testing:
- run tests

1. Closes Issues: #12466

1. [x] Updated meaningful TSDoc to methods including Parameters and Returns, see [Documentation Guide](https://esri.github.io/hub-components/storybook/?path=/story/guides-documentation--page)

1. [x] used semantic commit messages
  
1. [x] PR title follows semantic commit format (**CRITICAL** if the title is not in a semantic format, the release automation will not run!)

1. [ ] updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.

PR #1813, opened 2025-02-25
fix: revert "refactor: minimize use of rest-types package"

benstoltz:
Reverts Esri/hub.js#1809

This is being reverted due to breaking issues in ci / running hub-components from either #1808 or #1809, they went out together.


rweber-esri (approved):
LGTM

PR #1833, opened 2025-03-18
feat(hub-common): remove all collection, update default collection definitions, start persisting IHubCatalog to sites

sonofflynn89:
affects: @esri/hub-common

BREAKING CHANGE:
getCatalogFromSiteModel() no longer includes the "all" collection

1. Description:

We need to stop returning the "all" collection when calculating a site catalog as we do not ever intend to persist it in the data.json. FE and BE consumers of site catalogs that need to display an "all" collection will need to add one as a display-only collection.

We also changed the default site collections to stop using `IQuery.collection` and leverage well-known predicates instead (i.e., `$apps`).

Finally, we want to start actually persisting an IHubCatalog to site.data.catalogV2. This will be maintained in conjunction with the legacy catalog properties (site.data.catalog, site.data.values.searchCategories)


abp6318 (approved):

packages/common/src/sites/site-schema-version.ts
```diff
@@ -1 +1 @@
-export const SITE_SCHEMA_VERSION = 1.8;
+export const SITE_SCHEMA_VERSION = 1.9;
```
abp6318, line 1: Are there any future concerns to having to manually update this version? Or on the other hand, do we have code documentation for this file?

packages/common/src/sites/upgrade-site-schema.ts
```diff
@@ -14,6 +14,7 @@ import { migrateBadBasemap } from "./_internal/migrateBadBasemap";
 import { ensureBaseTelemetry } from "./_internal/ensureBaseTelemetry";
 import { migrateWebMappingApplicationSites } from "./_internal/migrateWebMappingApplicationSites";
 import { _migrateLinkUnderlinesCapability } from "./_internal/_migrate-link-underlines-capability";
+import { _migrateToV2Catalog } from "./_internal/_migrate-to-v2-catalog";
```
abp6318, line 17: Maybe I don't fully understand the limits of the `_internal` folder, but could we reduce these imports with a rolled up `index.ts` file exporting each of these files?

abp6318 (approved):
Two comments of very little significance. Looks good to me.

packages/common/src/sites/site-schema-version.ts
sonofflynn89: This is a well-known practice, thought it could be documented better. Definitely something to follow up with the platform team on.

packages/common/src/sites/upgrade-site-schema.ts
sonofflynn89: It's a good question! I don't think there's anything stopping us from making an index.ts file for `_internal`, but I also don't know if it would cause more dependency problems due to the file structure. We could likely ask Tom.

sonofflynn89:
@tomwayson, your cherry-picked commit looks good to me. I'm okay to merge with it.

tomwayson:
thanks @sonofflynn89, but that's only one of a few cherry picked commits that I'm going to add. I'm pushing each individually to see if tests pass, etc for exactly that commit. I'll let you know when they are all pushed up.

packages/common/src/sites/searchCategoriesToCollections.ts
```diff
@@ -0,0 +1,90 @@
...
+import { defaultSiteCollectionKeys } from "./defaultSiteCollectionKeys";
+
+/**
+ * Convert searchCategories to collections. If no searchCategories are provided,
+ * the default collections will be returned.
+ *
+ * @param searchCategories searchCategories from site model (data.values.searchCategories)
+ * @returns IHubCollections that respect the searchCategories configuration
+ */
+export function searchCategoriesToCollections(
```
sonofflynn89, line 13: I needed to expose this function, so the content library could use it. There's nothing new in the function, it's just copied out out `applyDefaultCollectionMigration()`. I didn't add tests because this function is covered by the existing tests in `applyDefaultCollectionMigration()`.

abp6318 (approved):
I think this looks good. Curious to test it on QA / with a uiVersion.

PR #1871, opened 2025-04-16
feat: add well-known group and event catalogs

juliannemarik:
- [12732](https://devtopia.esri.com/dc/hub/issues/12732)

### Description
- adds new well-known group & event catalogs
  - groups: "myGroups", "orgGroups", "communityGroups", "publicGroups"
  - events: "myEvents", "orgEvents", "worldEvents"
- adds a new `getWellKnownCatalogs` util which returns an array of requested well-known catalogs
- exports internal `getTagItems` util

### DIFF
--- a/packages/common/src/search/wellKnownCatalog.ts
+++ b/packages/common/src/search/wellKnownCatalog.ts
@@ -6,21 +6,42 @@ import { buildCatalog } from "./_internal/buildCatalog";
 import { getProp } from "../objects";
 import type { IArcGISContext } from "../types/IArcGISContext";
 
-/**
- * This is used to determine what IHubCatalog definition JSON object
- * can be created
- */
+/** well known item catalogs */
+export const WELL_KNOWN_ITEM_CATALOGS = [
+  "myContent",
+  "favorites",
+  "organization",
+  "world",
+  "partners",
+  "community",
+  "livingAtlas",
+] as const;
+export type WellKnownItemCatalog = (typeof WELL_KNOWN_ITEM_CATALOGS)[number];
+
+/** well known group catalogs */
+export const WELL_KNOWN_GROUP_CATALOGS = [
+  "editGroups",
+  "viewGroups",
+  "allGroups",
+  "myGroups",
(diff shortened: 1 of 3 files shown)


packages/common/src/search/wellKnownCatalog.ts
```diff
@@ -62,6 +84,51 @@ export function dotifyString(i18nScope: string): string {
...
+  const validCatalogNames = {
+    item: [...WELL_KNOWN_ITEM_CATALOGS] as WellKnownCatalog[],
+    group: [...WELL_KNOWN_GROUP_CATALOGS] as WellKnownCatalog[],
+    event: [...WELL_KNOWN_EVENT_CATALOGS] as WellKnownCatalog[],
+  }[targetEntity as "item" | "group" | "event"];
+  if (
+    !catalogNames.every((name: WellKnownCatalog) =>
+      validCatalogNames.includes(name)
+    )
+  ) {
```
vivzhang, line 115: Minor but I think it's better to use `filter` here so we can avoid duplicated checks - 

```
catalogNames.filter(
  (name: WellKnownCatalog) => !validCatalogNames.includes(name)
);
```

vivzhang (approved):
LGTM! Just one minor comment.

packages/common/src/search/wellKnownCatalog.ts
juliannemarik: Let me know if I'm misunderstanding your comment here, but the reason I'm using `every` is because I want it to throw an error if the user passes in a set of inconsistent catalog names. Using `filter` would only filter the array down to a set of consistent catalog names. Additionally, `every` will return `false` _immediately_ if any of the options don't pass the check, so I _think_ it should avoid duplicate checks.

vivzhang: I see. I misunderstood it earlier. I think `every` totally works fine then! Or maybe with `some` for better readability, but it's totally your call! 

`if (catalogNames.some(name => !validCatalogNames.includes(name))) {}`

PR #1901, opened 2025-05-14
fix: configuration editor now allows an empty value for the slug field

vivzhang:
affects: @esri/hub-common

ISSUES CLOSED: 13184

1. Description:

Update slug validation logic so it allows an empty value for the slug field

1. Instructions for testing:

### DIFF
--- a/packages/common/src/core/schemas/internal/addDynamicSlugValidation.ts
+++ b/packages/common/src/core/schemas/internal/addDynamicSlugValidation.ts
@@ -26,13 +26,29 @@ export const addDynamicSlugValidation = (
   _slug.maxLength = getSlugMaxLength(orgUrlKey);
   // add conditional validation to ensure slug is unique
   const allOf = cloneObject(schema.allOf) || [];
+
   const { pattern } = _slug;
+  // remove pattern from base schema so the form doesn't always apply the pattern
+  // validation. We will add it back to the allOf conditionally
+  delete _slug.pattern;
+
   const { id } = options as HubEntity;
-  allOf.push({
-    // only do async isUniqueSlug check if the slug is valid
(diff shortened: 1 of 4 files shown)


mjuniper (approved):
Nice work!

packages/common/src/core/schemas/internal/addDynamicSlugValidation.ts
```diff
@@ -26,13 +26,29 @@ export const addDynamicSlugValidation = (
...
+
   const { id } = options as HubEntity;
-  allOf.push({
-    // only do async isUniqueSlug check if the slug is valid
-    if: { properties: { _slug: { pattern } } },
-    then: { properties: { _slug: { isUniqueSlug: { id, orgUrlKey } } as any } },
-  });
+
+  allOf.push(
+    {
```
juliannemarik, line 38: Maybe this doesn't work for some reason, but I'm wondering if this should actually get moved up into the `SLUG_SCHEMA` definition [here](https://github.com/Esri/hub.js/blob/master/packages/common/src/core/schemas/shared/subschemas.ts#L185): 

```ts
export const SLUG_SCHEMA: JSONSchema = {
  type: "string",
  allOf: [
    {
      // if a slug has been entered, ensure it is lower case
      // alpha-numeric characters any '-' only
      if: { properties: { _slug: { minLength: 1 } } },
      then: { properties: { _slug: { pattern: "^[a-z0-9]+(?:-[a-z0-9]+)*$" } } },
    },
  ]
};
```

This would prevent you from having to delete the `_slug.pattern` on line 33 above as well

vivzhang: This makes sense and works!

PR #1909, opened 2025-05-21
fix: change group capabilities toggles to inline and add tooltips in group creation

vivzhang:
affects: @esri/hub-common

ISSUES CLOSED: 13236

1. Description:

Change group capabilities toggles to inline and add tooltips in group creation

1. Instructions for testing:


packages/common/src/groups/_internal/GroupUiSchemaCreate.ts
```diff
@@ -109,9 +109,13 @@ export const buildUiSchema = async (
             type: "Control",
             options: {
               control: "hub-field-input-switch",
+              layout: "inline-space-between",
               helperText: {
                 labelKey: `${i18nScope}.fields.isSharedUpdate.helperText`,
               },
+              tooltip: {
```
juliannemarik, line 116: looking at the actual tooltip text you configured in `opendata-ui`, it appears we only want to render a tooltip when the toggle is disabled. I _believe_ the way you have these configured now, they will always show up. Am I reading that correctly? 

It should be possible to only include the `tooltip` configuration if the relevant permissions are not met for the first two toggles with something like: 

```ts
...(checkPermission("platform:portal:admin:createUpdateCapableGroup", context).access && { tooltip: {
  labelKey: `${i18nScope}.fields.isSharedUpdate.tooltip`
}})
```
however, the last toggle looks like it's more complex and depends on `access` which is an editable value in the form itself, so I'm not sure how we handle that...

Off the top of my head: 
- short-term/fastest solution: change the tooltip text to be more generic. Something like "If the toggle is disabled, you don't have permission to enable this setting. Contact your organization’s administrator to adjust your account privileges."
- longer-term solution: enhance the config editor/rule interface to accept a tooltip configuration that renders if the rule condition(s) are not met

vivzhang: That's a good catch, I was a little confused too from what I read in the last tooltip and I just confirmed with Klara, we only want to show the tooltip when the toggle is disabled.

PR #1931, opened 2025-06-25
fix: internalize service flag processing in context manager

dbouwman:
1. Description:

Previously, service flags (`?sol=discussions&sna=events`) were processed and applied to the HubServiceStatus hash in the host application. However, as we added logic to ensure context got the service hash for Enterprise when appropriate, this logic became complex and located in several places.

To address that, this PR centralizes the processing - the Context Manager now accepts an `IServiceFlags` object, which the host app is responsible for constructing from the query params. If passed in, it applies those overrides to the service hash that's appropriate for the runtime environment (AGO or Enterprise)


packages/common/src/ArcGISContextManager.ts
```diff
@@ -432,7 +438,25 @@ export class ArcGISContextManager {
           ];
           break;
         case "serviceStatus":
-          this._serviceStatus = result as HubServiceStatus;
+          // eslint-disable-next-line no-case-declarations
+          const fetchedStatus = result as HubServiceStatus;
```
dbouwman, line 442: this is kinda gross, but not using this cast made the subsequent lines inscruable for all the needed type-casting.

packages/common/src/ArcGISContextManager.ts
```diff
@@ -520,9 +544,9 @@ export class ArcGISContextManager {
  * @param portalUrl
  * @returns
  */
-function getServiceStatus(portalUrl: string): Promise<HubServiceStatus> {
+async function getServiceStatus(portalUrl: string): Promise<HubServiceStatus> {
   let status = HUB_SERVICE_STATUS;
-  const isPortal = portalUrl.indexOf("arcgis.com") === -1;
+  const isPortal = !/^https?:\/\/([^.]+\.)*arcgis\.com(\/|$)/i.test(portalUrl);
```
tomwayson, line 549: seems like this should be centralized into a `isEnterprisePortalUrl()` util

PR #1967, opened 2025-08-11
fix(hub-common): page maps location property

juliannaeapicella:
1. Description: Page entities now properly map location property.

1. Instructions for testing:

1. Closes Issues: #13866

***CRITICAL** 

### DIFF
--- a/packages/common/src/content/_internal/getPropertyMap.ts
+++ b/packages/common/src/content/_internal/getPropertyMap.ts
@@ -21,11 +21,6 @@ export function getPropertyMap(): IPropertyMap[] {
   map.push({ entityKey: "permissions", storeKey: "data.permissions" });
 
   map.push({
-    entityKey: "location",
-    storeKey: "item.properties.location",
-  });
-
-  map.push({
     entityKey: "licenseInfo",
     storeKey: "item.licenseInfo",
   });
--- a/packages/common/src/discussions/_internal/getPropertyMap.ts
+++ b/packages/common/src/discussions/_internal/getPropertyMap.ts
@@ -17,10 +17,6 @@ export function getPropertyMap(): IPropertyMap[] {
     storeKey: "data.prompt",
   });
   map.push({
-    entityKey: "location",
-    storeKey: "item.properties.location",
-  });
-  map.push({
     entityKey: "catalog",
     storeKey: "data.catalog",
   });
--- a/packages/common/src/initiatives/_internal/getPropertyMap.ts
+++ b/packages/common/src/initiatives/_internal/getPropertyMap.ts
@@ -28,11 +28,6 @@ export function getPropertyMap(): IPropertyMap[] {
   map.push({ entityKey: "metrics", storeKey: "item.properties.metrics" });
 
   map.push({
-    entityKey: "location",
-    storeKey: "item.properties.location",
-  });
-
-  map.push({
     entityKey: "followersGroupId",
     storeKey: "item.properties.followersGroupId",
   });
--- a/packages/common/src/projects/_internal/getPropertyMap.ts
+++ b/packages/common/src/projects/_internal/getPropertyMap.ts
@@ -25,10 +25,6 @@ export function getPropertyMap(): IPropertyMap[] {
     entityKey: "features",
     storeKey: "data.settings.features",
   });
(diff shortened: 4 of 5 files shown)


packages/common/src/core/_internal/getBasePropertyMap.ts
```diff
@@ -70,6 +70,10 @@ export function getBasePropertyMap(): IPropertyMap[] {
     entityKey: "boundary",
     storeKey: "item.properties.boundary",
   });
+  map.push({
```
tommybramble, line 73: From my testing this looks good and works correctly for me, but I have a couple questions so I understand:

1. Why are we adding this here instead of at `/packages/common/src/pages/_internal/getPropertyMap.ts`?
   - I see the version of this file for other entities (e.g. sites, projects) adds this line in their respective getPropertMap.ts file.
   - Should the equivalent line be removed from the other entity specific property mapper logic since it's now included in the core file?
1. Does adding this at the core/base version affect any other entities? Will this expose the location properties to entities that shouldn't have it?

tommybramble (approved):
This LGTM and can't find any regressions testing other entity location pickers. I have a few questions so I understand, but I'm approving assuming those questions don't prompt any changes.

packages/common/src/core/_internal/getBasePropertyMap.ts
juliannaeapicella: @tommybramble All entities have a location property, though not all of them would map to `item.properties.location`. `getBasePropertyMap` seems to be designed for item-backed entities already, though, so I'm not sure why this property isn't included in the base map to begin with.
It is probably safer to just add it to `/packages/common/src/pages/_internal/getPropertyMap.ts`, though... worth checking with @rweber-esri to see if we should keep it contained to pages like you mentioned or consider moving the logic from the other entities into this file.

rweber-esri: I think it makes sense to add this here; it's used by all AGO-backed entities, _most_ of which have a location picker form field in their workspace details pane form (exception is `templates` and `initiativeTemplates`). I suspect support for locations on entities were implemented over time and devs opted to redundantly add this common property mapping to the entity's property mapper rather than in the base property mapper they all utilize. If we're adding this here, we should probably remove the redundant definitions in:

* [common/src/discussions/_internal/getPropertyMap.ts](https://github.com/Esri/hub.js/blob/13866/custom-location-not-removable/packages/common/src/discussions/_internal/getPropertyMap.ts#L19)
* [common/src/content/_internal/getPropertyMap.ts](https://github.com/Esri/hub.js/blob/13866/custom-location-not-removable/packages/common/src/content/_internal/getPropertyMap.ts#L23)
* [common/src/projects/_internal/getPropertyMap.ts](https://github.com/Esri/hub.js/blob/13866/custom-location-not-removable/packages/common/src/projects/_internal/getPropertyMap.ts#L28)
* [common/src/initiatives/_internal/getPropertyMap.ts](https://github.com/Esri/hub.js/blob/13866/custom-location-not-removable/packages/common/src/initiatives/_internal/getPropertyMap.ts#L30)
* [common/src/sites/_internal/getPropertyMap.ts](https://github.com/Esri/hub.js/blob/13866/custom-location-not-removable/packages/common/src/sites/_internal/getPropertyMap.ts#L62)

As mentioned above, adding this to the base property mapper will result in `templates` and `initiativeTemplates` getting this mapping now too. I suspect that should be completely transparent to the users, but please verify there are no deltas in UI or behavior (_both_ workspace _and_ entity view) as a result of this change using a PR-Preview of this branch and OD-UI uiVersion.

* [initiative-template](https://qa-pre-a-hub.hubqa.arcgis.com/workspace/initiativeTemplates/7993c71c03b44956a9d86768d14267c5/details) as `juliana_pa`
* [template](https://qa-pre-a-hub.hubqa.arcgis.com/workspace/templates/9f5d5af106e74d2bb808c37e81c98192/details) as `juliana_pa`

rweber-esri (approved):
LGTM. Left some feedback regarding redundant property mappings that can be purged. Please verify no regressions to `templates` or `initiativeTemplates` as a result of these changes using your OD-UI uiVersion.

PR #1974, opened 2025-08-19
fix: page slugify regex

abp6318:
1. Description: Slugs generated based off of a new page's provided title would incorrectly allow for leading hyphens and underscores

1. Instructions for testing: TBA

1. Closes Issues: https://devtopia.esri.com/dc/hub/issues/13433

***CRITICAL** 

### DIFF
--- a/packages/common/src/utils/slugify.ts
+++ b/packages/common/src/utils/slugify.ts
@@ -2,8 +2,9 @@
  * Perform the following operations on a string to make it slug-friendly:
  * 1. trim it
  * 2. convert to lowercase
- * 3. remove any character not a-z, 0-9, or _
+ * 3. remove any character not a-z or 0-9
  * 4. dasherize it
+ * 5. remove leading and trailing dashes
  * @param {String} value String to slugify
  */
 export function slugify<T>(value: T): T {
@@ -12,9 +13,9 @@ export function slugify<T>(value: T): T {
     return value
       .trim()
       .toLowerCase()
-      .replace(/ +/g, "-")
-      .replace(/[^\w-]/g, "")
-      .replace(/-+/g, "-");
+      .replace(/[^a-z0-9]+/g, "-") // replace runs of anything NOT a-z or 0-9 with "-"
+      .replace(/-+/g, "-") // collapse multiple "-" into one
+      .replace(/^-+|-+$/g, ""); // trim leading/trailing "-"
   } else {
     return value;
   }


packages/common/src/utils/slugify.ts
```diff
@@ -1,20 +1,24 @@
...
     // @ts-ignore
     return value
-      .trim()
-      .toLowerCase()
-      .replace(/ +/g, "-")
-      .replace(/[^\w-]/g, "")
-      .replace(/-+/g, "-");
+      .trim() // remove leading/trailing spaces
+      .toLowerCase() // convert to lowercase
+      .replace(/['"]/g, "") // remove single and double quotes
```
abp6318, line 18: So this one isn't a problem for pages, but is a needed change on it's own line in order to prevent a regression for event title slugification

sonofflynn89 (approved):
Good work on this Aaron! Thank you!

packages/common/src/utils/slugify.ts
```diff
@@ -12,9 +13,9 @@ export function slugify<T>(value: T): T {
     return value
       .trim()
       .toLowerCase()
-      .replace(/ +/g, "-")
-      .replace(/[^\w-]/g, "")
-      .replace(/-+/g, "-");
+      .replace(/[^a-z0-9]+/g, "-") // replace runs of anything NOT a-z or 0-9 with "-"
```
knbhagat: My only question is how we want to handle non-ASCII characters. For example, "Déjà Vu" would become "dj-vu". Are titles allowed to include non-ASCII characters? If not, this implementation should be fine. The previous code didn’t handle this case either, so this isn’t introducing a new issue — just something to confirm.

packages/common/src/items/slugs.ts
```diff
@@ -22,7 +22,14 @@ export function constructSlug(title: string, orgKey = ""): string {
...
   // slug|qa-pre-a-hub|some-really-really-...-really-100
   const paddingEnd = 4;
-  return truncateSlug(slugify(title), orgKey.toLowerCase(), paddingEnd);
+  const slugified = slugify(title);
+  if (!slugified) {
+    // If slugify returns an empty string, return an empty string for the complete slug as well.
+    // This is a rather strange edge case where in the process of auto-slugging a title for a Page or Event,
+    // we end up with an invalid slug because the title did not include any valid characters for a slug [a-z0-9].
+    return "";
+  }
```
abp6318, line 31: If a user enters a valid title that would result in a slug of `""` or `"-"`, we no longer set the slug at all. 

We'd rather have no slug than an invalid slug, and we can't change the pattern of the title validation input on creation to require a-z0-9 for the slug alone.

To avoid losing out on the slug entirely, we'd need to complete https://devtopia.esri.com/dc/hub/issues/13959 as a follow-up.

PR #1981, opened 2025-08-25
feat(hub-common): add group workspace discussions settings permissions

juliannaeapicella:
1. Description: Adds permissions for group discussion settings workspace pane

1. Instructions for testing:

1. Closes Issues: #13805

***CRITICAL** 

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -1,3 +1,10 @@
+# @esri/hub-common [20.2.0](https://github.com/Esri/hub.js/compare/@esri/hub-common@20.1.1...@esri/hub-common@20.2.0) (2025-08-29)
+
+
+### Features
+
+* **hub-common:** Prevent items list from being resorted on save event ([#1982](https://github.com/Esri/hub.js/issues/1982)) ([1925493](https://github.com/Esri/hub.js/commit/19254931ce3256d5f2978e51271ddaab23f613cc))
+
 ## @esri/hub-common [20.1.1](https://github.com/Esri/hub.js/compare/@esri/hub-common@20.1.0...@esri/hub-common@20.1.1) (2025-08-28)
 
 
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@esri/hub-common",
-  "version": "20.1.1",
+  "version": "20.2.0",
   "description": "Common TypeScript types and utility functions for @esri/hub.js.",
   "main": "dist/node/index.js",
   "module": "dist/esm/index.js",
--- a/packages/common/src/core/schemas/internal/getTagItems.ts
+++ b/packages/common/src/core/schemas/internal/getTagItems.ts
@@ -60,7 +60,7 @@ export async function getTagItems(
      * any empty values and convert them to the IUiSchemaComboboxItem format
      */
     const entityTags = tags || [];
-    return [...new Set([...tagsAgg.values.map((t) => t.value), ...entityTags])]
+    return [...new Set([...entityTags, ...tagsAgg.values.map((t) => t.value)])]
       .filter((t) => t)
       .map((t) => ({ value: t }));
   } catch (e) {
--- a/packages/common/src/utils/array.ts
+++ b/packages/common/src/utils/array.ts
@@ -1,0 +1,19 @@
+/**
+ * Determines if two arrays are shallowly equal.
+ *
+ * @param a - The first array to compare.
+ * @param b - The second array to compare.
+ * @returns True if both arrays are the same reference or contain identical elements in the same order; otherwise, false.
+ */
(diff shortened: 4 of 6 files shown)


packages/common/src/groups/_internal/GroupSchema.ts
```diff
@@ -11,14 +11,14 @@ export type GroupEditorType = (typeof GroupEditorTypes)[number];
 export const GroupEditorTypes = [
   "hub:group:edit",
   "hub:group:settings",
-  "hub:group:discussions",
```
rweber-esri, line 14: Removing line 14 changes all the other subsequent values' indices which is related to the test failures you're observing. I would recommend replacing the original line 14 with the line you added on 21

packages/common/test/core/schemas/internal/getEditorSchemas.test.ts
```diff
@@ -143,7 +142,6 @@ describe("getEditorSchemas: ", () => {
 
     { type: GroupEditorTypes[0], module: GroupBuildEditUiSchema },
     { type: GroupEditorTypes[1], module: GroupBuildSettingsUiSchema },
-    { type: GroupEditorTypes[2], module: GroupBuildDiscussionsUiSchema },
```
rweber-esri, line 146: Rather than removing, this should be updated to something like:

```js
    { type: GroupEditorTypes[2], module: EntityBuildDiscussionSettingsUiSchema },
```

vivzhang (changes requested):
These changes look good and are a solid step forward. One thing still missing for full support of allowed channels is creating and updating the `discussionSettings` and `entitySettingsId` in the entity. Without that, the allowed channels disappear after a page refresh.

I'd suggest starting the investigation in the [updateHubGroup](https://github.com/Esri/hub.js/blob/master/packages/common/src/groups/HubGroups.ts#L158-L181) call(we should make similar changes in the [createHubGroup](https://github.com/Esri/hub.js/blob/master/packages/common/src/groups/HubGroups.ts#L101-L134) call as well). You may refer to the [updateContent](https://github.com/Esri/hub.js/blob/master/packages/common/src/content/edit.ts#L130-L264) and [createContent](https://github.com/Esri/hub.js/blob/master/packages/common/src/content/edit.ts#L71-L122) methods for an example of how content is handled, though the details may not be exactly the same.

packages/common/src/groups/_internal/GroupBusinessRules.ts
```diff
@@ -248,13 +248,13 @@ export const GroupPermissionPolicies: IPermissionPolicy[] = [
...
-    services: ["discussions"],
+    permission: "hub:group:workspace:settings",
     dependencies: ["hub:group:workspace"],
   },
   {
-    permission: "hub:group:workspace:settings",
-    dependencies: ["hub:group:workspace"],
+    permission: "hub:group:settings:discussions",
+    services: ["discussions"],
+    dependencies: ["hub:group:workspace:settings"],
```
rweber-esri, line 257: we should add `licenses: ["hub-basic", "hub-premium"]` to this permission policy definition, else the `Settings => Discussions` workspace pane will render in the group workspace in enterprise-sites (similar to [14010](https://devtopia.esri.com/dc/hub/issues/14010) & [PR-1984](https://github.com/Esri/hub.js/pull/1984)).

juliannaeapicella:
> These changes look good and are a solid step forward. One thing still missing for full support of allowed channels is creating and updating the `discussionSettings` and `entitySettingsId` in the entity. Without that, the allowed channels disappear after a page refresh.
> 
> I'd suggest starting the investigation in the [updateHubGroup](https://github.com/Esri/hub.js/blob/master/packages/common/src/groups/HubGroups.ts#L158-L181) call(we should make similar changes in the [createHubGroup](https://github.com/Esri/hub.js/blob/master/packages/common/src/groups/HubGroups.ts#L101-L134) call as well). You may refer to the [updateContent](https://github.com/Esri/hub.js/blob/master/packages/common/src/content/edit.ts#L130-L264) and [createContent](https://github.com/Esri/hub.js/blob/master/packages/common/src/content/edit.ts#L71-L122) methods for an example of how content is handled, though the details may not be exactly the same.

@vivzhang Thanks for catching this! That should be working now.

packages/common/src/groups/HubGroups.ts
```diff
@@ -113,6 +114,15 @@ export async function createHubGroup(
     hubGroup.typeKeywords,
     hubGroup.isDiscussable
   );
+
+  // create or update entity settings
+  const entitySetting = await createOrUpdateEntitySettings(
+    hubGroup,
+    requestOptions
```
vivzhang, line 121: I'm not sure if this will actually work, since `createOrUpdateEntitySettings` takes an `IHubRequestOptions`, but the `requestOptions` here is an `IUserRequestOptions`.

PR #1987, opened 2025-09-02
fix(hub-common): fix issue where existing download export requests were not serialized correctly

sonofflynn89:
affects: @esri/hub-common, @esri/hub-downloads

1. Description:
Part of https://devtopia.esri.com/dc/hub/issues/14008. Turns out that in rest.js 4, the QueryBuilder class stopped automatically wrapping values in double quotes, which causes major issues for the enterprise download flow.

***CRITICAL** 


tomwayson (approved):
good catch

abp6318 (approved):
LGTM; verified on the enterprise machine you spun up

PR #1995, opened 2025-09-09
fix: 13240 add location to initiative template

vivzhang:
1. Description:

1. Instructions for testing:

1. [ ] Updated meaningful TSDoc to methods including Parameters and Returns, see [Documentation Guide](https://friendly-adventure-7w1eyl2.pages.github.io/storybook/?path=/story/guides-documentation--page)

1. [ ] Updated `peerDependencies` as needed. **CRITICAL** our automated release system can **not** be counted on to update `peerDependencies` so we _must_ do it manually in our PRs when needed. See the [updating peerDependencies](/RELEASE.md#Updating-peerDependencies) section of the release instructions for more details.

1. [ ] These changes have been verified by QA using a `?uiVersion` that includes a PR-Preview of this branch and the `v_req` label has been applied to the issue.

1. [ ] OD-UI E2E tests pass against these changes using a `?uiVersion` that includes a PR-Preview of this branch
 
***CRITICAL** 


packages/common/src/initiative-templates/edit.ts
```diff
@@ -113,6 +114,10 @@ export async function updateInitiativeTemplate(
     updatedModel,
     initiativeTemplate
   );
+  // the showMap prop is recently added (2025-09) to the default initiative
+  // template, so we need to ensure it is set to true for existing ones that
+  // may not have it
+  setProp("view.showMap", true, initiativeTemplate);
```
tommybramble, line 120: I'm assuming there is no way for the `showMap` property to ever be set to false by a user as this would override it?

tommybramble (approved):
LGTM, but was curious about the one line to `setProp`. I am not really familiar with initiatives or templates, but just wanted to make sure that it wouldn't override a valid update to set the property to false. Disregard the comment if that is not possible.

packages/common/src/initiative-templates/edit.ts
vivzhang: Currently, we don't have a field in the initiative template workspace for users to configure show/hide maps, but that's a valid point, it's safer to only assign the value if it doesn't exist. So in the future if we decide to support the configuration, this won't get overwritten.

packages/common/src/core/types/IHubInitiativeTemplate.ts
```diff
@@ -1,13 +1,19 @@
 import { IHubItemEntity } from "./IHubItemEntity";
-import { IWithLayout, IWithSlug, IWithCatalog } from "../traits";
+import {
+  IWithLayout,
+  IWithSlug,
+  IWithCatalog,
+  IWithViewSettings,
+} from "../traits";
```
rweber-esri, line 7: These import from a "barrel" file (i.e. index.ts) which we should avoid doing at all costs because it can:
1. Potentially introduce cyclical dependencies
2. Cause issues trying to stub fns in tests
3. Negatively impact tree-shaking leading to larger asset bundles that import things from Hub.js

I'm working on a PR that fully-qualifies import paths in all hub-common modules and tests as part of my efforts to reduce cyclical dependencies in Hub.js. I plan to experiment with some ESLint rules to prevent us from importing from barrel files moving forward.

rweber-esri (approved):
LGTM. One minor/quick note about not importing from barrel files. Approved once those import paths are fully-qualified to the modules that define and export the imported artifacts.

packages/common/src/core/types/IHubInitiativeTemplate.ts
vivzhang: Fixed the imports

PR #2125, opened 2025-11-14
chore: get build and test commands running under packages/common

tomwayson:
1. Description:

This PR prepares hub-common to be moved to hub-ui by refactoring the build and test tooling to enable independent operation within the packages/common directory. The main changes consolidate build and linting configurations at the package level rather than relying on monorepo-wide tooling.

Key Changes
- Moved linting and build tool dependencies from root to packages/common/package.json
- Inlined TypeScript configuration in packages/common/tsconfig.json instead of extending root config
- Updated ESLint file path patterns to be relative to package root rather than monorepo root

NOTE: These changes finally broke the Acetate based `docs:build` script - so I have disabled that step.

I also removed eslint from the pre-commit hook as it's no longer available at the root.

1. Instructions for testing:

1. [ ] Updated meaningful TSDoc to methods including Parameters and Returns, see [Documentation Guide](https://friendly-adventure-7w1eyl2.pages.github.io/storybook/?path=/story/guides-documentation--page)

1. [x] Either ran `npm run changeset` or this change does not require a release.

1. [ ] These changes have been verified by QA using a `?uiVersion` that includes a PR-Preview of this branch and the `v_req` label has been applied to the issue.

1. [ ] OD-UI E2E tests pass against these changes using a `?uiVersion` that includes a PR-Preview of this branch


package.json
```diff
@@ -73,7 +65,7 @@
 		"build": "npm run build:esm && npm run build:node",
 		"changeset": "npx @changesets/cli",
 		"ci:build": "npm run build",
-		"ci:lint": "npm run lint && npm run madge:circular",
+		"ci:lint": "npm --prefix ./packages/common run ci:lint",
```
rweber-esri: Running `npm run build` and `npm run test` from the monorepo root still works as expected, but running `npm run lint` results in an error:



Would be good/more consistent if we could still trigger the lint from the monorepo root.

rweber-esri (approved):
Overall, this is looking good to me. I was able to run `npm i`, build & test the project. I additionally verified ESLint is working as expected. One minor note relating to the linting.

